PR #3, opened 2020-02-28
Add link to more documentation in README

AliSoftware:
Also fixed some typos in the documentation

\<reverted> ~Also fixed protocol conformance part in README example because that surprised me that we declared conformance on type declaration but without implementing the conformance, and only implementing the conformance in extension later… which at least I thought would not compile… but now @harlanhaskins  made me doubt it was invalid 🤔)~ \</reverted>

### DIFF
--- a/Documentation/01 Getting Started.md
+++ b/Documentation/01 Getting Started.md
@@ -65,9 +65,9 @@ Count.main()
 
 In the code above, the `inputFile` and `outputFile` properties use the `@Argument` property wrapper. `ArgumentParser` uses this wrapper to denote a positional command-line input — because `inputFile` is specified first in the `Count` type, it's the first value read from the command-line, and `outputFile` is read second.
 
-We've implemented the command's logic in its `run()` method. Here, we're printing out a message confirming the names of the files the user gave. (You can find a full implementation of the completed command at the end of this guide.)
+We have implemented the command's logic in its `run()` method. Here, we are printing out a message confirming the names of the files the user gave. (You can find a full implementation of the completed command at the end of this guide.)
 
-Finally, you tell the parser to execute the `Count` command by calling its static `main()` method. This method parses the command-line arguments, verifies that they match up with what we've defined in `Count`, and either calls the `run()` method or exits with a helpful message.
+Finally, you tell the parser to execute the `Count` command by calling its static `main()` method. This method parses the command-line arguments, verifies that they match up with what we have defined in `Count`, and either calls the `run()` method or exits with a helpful message.
 
 
 ## Working with Named Options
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -3,13 +3,12 @@
...
 
 Begin by declaring a type that defines the information you need to collect from the command line.
-Decorate each stored property with one of `ArgumentParser`'s property wrappers,
-and declare conformance to `ParsableCommand`.
+Decorate each stored property with one of `ArgumentParser`'s property wrappers.
 
 ```swift
 import ArgumentParser
 
-struct Repeat: ParsableCommand {
```
zntfdr: I advise against this: it's important to know what the type is conforming to in its declaration. 

By moving it in the extension, people have to scroll to the extension to discover the conformation.

AliSoftware: But is this valid Swift though ? (If yes, then TIL)

I've seen other syntaxes that I know are valid yes, but never the case of declaring conformance without actually confirming on the declaration

```
protocol P {
  func pReq()
}

// Ok for that of course :
struct S1: P {
  func pReq()
}

// ok for that too;
struct S2 {}
extension S2: P {
  func pReq() {}
}

// And even also this:
struct S3 {
  func pReq()
}
extension S3: P {}

// But this is news to me though:
struct S4: P {}
extension S4 {
  func pReq()
}
```
If that's indeed valid Swift I'll be happy to revert and to end up learning something today, but it just felt surprising to me when I read that in the README in the first place that's all

zntfdr: Yes, you it's valid 😃 here's a working example (you were missing the `{}` in the last extension):
```swift
protocol P {
  func pReq()
}

struct S4: P {}

extension S4 {
  func pReq() {}
}
```

AliSoftware: Wow, good to know, TIL 👍
I'm guessing the extension has to be in the same file? Or it also works if declared elsewhere (well as long as declared in the same module)?

Will revert those changes then 👍

zntfdr: Yes, it works as long as it's defined in the same module (which means that having an extension in a separate file is also valid, as long as the extension is at least `internal`) 🚀

natecook1000: I agree that it looks a bit weird with `run()` split out into an extension. What about replacing the closing bracket in the type declaration with a comment like `// continued below...` and then deleting the line that opens the extension? Then it would be just one type declaration that’s broken up by some explanatory text.

AliSoftware:
@natecook1000 Just pushed your suggestion. Didn't remove the closing bracket though as it felt a bit odd to read too, so instead added `// continued below…` and `// continued from above` while keeping the declaration scope visible. wdyt?

I also took the occasion to push additional typo fixes that I saw while reading the rest of the documentation 🙂

zntfdr:
Yes I like it much better now! Thank you @AliSoftware @natecook1000 🤗

Documentation/01 Getting Started.md
```diff
@@ -65,9 +65,9 @@ Count.main()
 
 In the code above, the `inputFile` and `outputFile` properties use the `@Argument` property wrapper. `ArgumentParser` uses this wrapper to denote a positional command-line input — because `inputFile` is specified first in the `Count` type, it's the first value read from the command-line, and `outputFile` is read second.
 
-We've implemented the command's logic in its `run()` method. Here, we're printing out a message confirming the names of the files the user gave. (You can find a full implementation of the completed command at the end of this guide.)
+We have implemented the command's logic in its `run()` method. Here, we are printing out a message confirming the names of the files the user gave. (You can find a full implementation of the completed command at the end of this guide.)
```
natecook1000: The contractions are okay as is. Could you revert those changes?

AliSoftware: Reverted in f086616 🙂

PR #9, opened 2020-02-28
Fix readme requirement name

zntfdr:
I've checked on both SPM master and the swift-5.2 branch: 
it seems that the `Package.Dependency.Requirement` name is always [`.upToNextMinor(..)`](https://github.com/apple/swift-package-manager/blob/f2318da4f96b0607ade27eff51558206f57c2f98/Sources/PackageDescription/PackageRequirement.swift#L95).


natecook1000:
Absolutely right, thanks @zntfdr!

PR #10, opened 2020-02-28
Provide suggestions for unknown options

### LINKED ISSUE #2: Provide near-miss suggestions
If a user mistypes an option, flag, or command, the error message should include a suggestion for the near-miss.

Example:

```
% example --indx 5
Error: Unexpected argument '--indx'
       Did you mean '--index'?
Usage: example [--index <index>] [<values> ...]
```

klaaspieter:
When an unknown option is encountered use levenshtein to find the closest matching option. Simply prints the error if there is no close match ([distance >= 7](https://github.com/git/git/blob/228f53135a4a41a37b6be8e4d6e2b6153db4a8ed/help.c#L592)).

Fixes #2

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -78,6 +78,13 @@ extension CommandParser {
     
     // We should have used up all arguments at this point:
     guard split.isEmpty else {
+      // Check if one of the arguments is an unknown option
+      for (index, element) in split.elements {
+        if case .option(let argument) = element {
+          throw ParserError.unknownOption(InputOrigin.Element.argumentIndex(index), argument.name)
+        }
+      }
+       
       let extra = split.coalescedExtraElements()
       throw ParserError.unexpectedExtraValues(extra)
     }
--- a/Sources/ArgumentParser/Usage/UsageGenerator.swift
+++ b/Sources/ArgumentParser/Usage/UsageGenerator.swift
@@ -248,14 +248,7 @@ extension ErrorMessageGenerator {
   }
   
   func unknownOptionMessage(origin: InputOrigin.Element, name: Name) -> String {
-    switch name {
-    case .long(let n):
-      return "Unknown option '--\(n)'"
-    case .short(let n):
-      return "Unknown option '-\(n)'"
-    case .longWithSingleDash(let n):
-      return "Unknown option '-\(n)'"
-    }
+    return "Unknown option '\(name.synopsisString)'."
   }
   
   func missingValueForOptionMessage(origin: InputOrigin, name: Name) -> String {
--- a/Sources/ArgumentParser/Usage/UsageGenerator.swift
+++ b/Sources/ArgumentParser/Usage/UsageGenerator.swift
@@ -246,9 +246,35 @@ extension ErrorMessageGenerator {
   var invalidState: String {
     return "Internal error. Invalid state while parsing command line arguments."
   }
-  
+
+
   func unknownOptionMessage(origin: InputOrigin.Element, name: Name) -> String {
-    return "Unknown option '\(name.synopsisString)'."
+    if case .short = name {
+      return "Unknown option '\(name.synopsisString)'"
+    }
+    
+    // An empirically derived magic number
+    let SIMILARITY_FLOOR = 4
+
+    let notShort: (Name) -> Bool = { (name: Name) in
+      switch name {
+      case .short: return false
+      case .long: return true
+      case .longWithSingleDash: return true
+      }
+    }
+    let suggestion = arguments
+      .flatMap({ $0.names })
+      .filter({ $0.synopsisString.editDistance(to: name.synopsisString) < SIMILARITY_FLOOR }) // only include close enough suggestion
+      .filter(notShort) // exclude short option suggestions
+      .min(by: { lhs, rhs in // find the suggestion closest to the argument
+        lhs.synopsisString.editDistance(to: name.synopsisString) < rhs.synopsisString.editDistance(to: name.synopsisString)
+      })
+    
+    if let suggestion = suggestion {
+        return "Unknown option '\(name.synopsisString)'. Did you mean '\(suggestion.synopsisString)'?"
+    }
+    return "Unknown option '\(name.synopsisString)'"
   }
   
   func missingValueForOptionMessage(origin: InputOrigin, name: Name) -> String {
--- a/Sources/ArgumentParser/Utilities/StringExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/StringExtensions.swift
@@ -144,4 +144,51 @@ extension String {
       self[range].lowercased()
     }.joined(separator: String(separator))
   }
+  
+  /// Returns the edit distance between this string and the provided target string.
+  ///
+  /// Uses the Levenshtein distance algorithm internally.
+  ///
+  /// See: https://en.wikipedia.org/wiki/Levenshtein_distance
+  ///
+  /// Examples:
+  ///
+  ///     "kitten".editDistance(to: "sitting")
+  ///     // 3
+  ///     "bar".editDistance(to: "baz")
+  ///     // 1
+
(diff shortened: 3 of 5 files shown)


Sources/ArgumentParser/Utilities/StringExtensions.swift
```diff
@@ -144,4 +144,38 @@ extension String {
       self[range].lowercased()
     }.joined(separator: String(separator))
   }
+  
+  func editDistance(to target: String) -> Int {
```
Sajjon: Maybe document that this is using [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance)(as per your description in this PR)? :) 

Also can we maybe come up with a better naming? I think `fuzzyDistance:to`, `measuredSimilarity:with`, `measuredDifference:with` are a bit more clear, at least to me.

```swift
"bar".measuredSimilarity(with: "baz")
```
reads out better than:
```swift
"bar".fuzzyDistance(to: "baz")
```

natecook1000: A comment about the algorithm would be great, but I think the name is fine here — "edit distance" is a common term for this algorithm.

natecook1000:
@klaaspieter Are there breaks in the actual functionality (which command is selected / how values are populated), or does it just change the error messages? I don't think we need to treat the latter as a breaking change at this point in development.

Sources/ArgumentParser/Utilities/StringExtensions.swift
klaaspieter: @natecook1000 my thoughts exactly. I'll add the comment :)

klaaspieter:
> @klaaspieter Are there breaks in the actual functionality (which command is selected / how values are populated), or does it just change the error messages? I don't think we need to treat the latter as a breaking change at this point in development.

It's a little bit of both. I think the error actually get's better because it says `Unknown option -x` now. Instead of `Unexpected argument <value of argument>`. 

That said there are other tests that assert a certain default state that now receive an error. For example this [`NestedCommandEndToEndTest`](https://github.com/apple/swift-argument-parser/blob/master/Tests/EndToEndTests/NestedCommandEndToEndTests.swift#L66-L69) fails because `-f` is not a valid option.

natecook1000:
> It's a little bit of both. I think the error is actually better because it says Unknown option -x now. Instead of Unexpected argument <value of argument>.

Sounds like we need to address this issue first. If we don't want to change the overall parsing behavior, we can make this `throw` distinguish between options and values: https://github.com/apple/swift-argument-parser/blob/master/Sources/ArgumentParser/Parsing/CommandParser.swift#L82

klaaspieter:
> Sounds like we need to address this issue first. If we don't want to change the overall parsing behavior, we can make this throw distinguish between options and values: https://github.com/apple/swift-argument-parser/blob/master/Sources/ArgumentParser/Parsing/CommandParser.swift#L82

It looks like this is actually a separate issue https://github.com/apple/swift-argument-parser/issues/23. This should be possible to get done without breaking changes when testing with options that don't take arguments 😄

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -78,6 +78,13 @@ extension CommandParser {
     
     // We should have used up all arguments at this point:
     guard split.isEmpty else {
+      // Check if one of the arguments is an unknown option
```
klaaspieter: This feels like a weird spot to essentially redo the work done here https://github.com/apple/swift-argument-parser/blob/339baea0a5d8c9b26c6e9be47df8e713f6e26234/Sources/ArgumentParser/Parsing/ArgumentSet.swift#L436. However that error is swallowed here https://github.com/apple/swift-argument-parser/blob/e17dac8424d0275d082dacf00524f2c372d50bf8/Sources/ArgumentParser/Parsing/ArgumentSet.swift#L373-L375 to support nested command structures (https://github.com/apple/swift-argument-parser/issues/23#issuecomment-592983185).

Perhaps there is a better solution, but this is what I came up with 😄

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -246,16 +246,35 @@ extension ErrorMessageGenerator {
...
-    case .short(let n):
-      return "Unknown option '-\(n)'"
-    case .longWithSingleDash(let n):
-      return "Unknown option '-\(n)'"
+    if case .short = name {
+      return "Unknown option '\(name.synopsisString)'"
+    }
+    
+    // An empirically derived magic number
+    let SIMILARITY_FLOOR = 4
```
klaaspieter: This was set at 7 because that is what [`git` does](https://github.com/git/git/blob/228f53135a4a41a37b6be8e4d6e2b6153db4a8ed/help.c#L592). However it leads us to suggest `--name` for `-bar` or `--verbose` here: https://github.com/apple/swift-argument-parser/pull/10/files#diff-f55d39a7e333981fb00eaa62155a2c8cL35-R39. The suggestions seemed to disimilar to allow.

Tests/UnitTests/ErrorMessageTests.swift
```diff
@@ -32,19 +32,19 @@ extension ErrorMessageTests {
...
   }
   
   func testUnknownOption_3() {
-    AssertErrorMessage(Bar.self, ["--name", "a", "--format", "b", "-bar"], "Unexpected argument '-bar'")
+    AssertErrorMessage(Bar.self, ["--name", "a", "--format", "b", "-bar"], "Unknown option '-bar'")
   }
   
   func testUnknownOption_4() {
-    AssertErrorMessage(Bar.self, ["--name", "a", "-foz", "b"], "2 unexpected arguments: '-o', '-z'")
+    AssertErrorMessage(Bar.self, ["--name", "a", "-foz", "b"], "Unknown option '-o'")
```
klaaspieter: This is the only change I'm not sure about. I feel the previous message is more informative. Both are correct though.

Sources/ArgumentParser/Utilities/StringExtensions.swift
klaaspieter: Comment added.

Sources/ArgumentParser/Parsing/CommandParser.swift
natecook1000: `ArgumentSet.first(matching:at:)` could just return an `ArgumentDefinition?` instead of throwing, since we've postponed the actual throwing until this later point.

Sources/ArgumentParser/Usage/UsageGenerator.swift
natecook1000: 👍 7 did seem high!

Tests/UnitTests/ErrorMessageTests.swift
natecook1000: We can come back later and handle the multiple unknown options case.

Tests/UnitTests/ErrorMessageTests.swift
```diff
@@ -109,3 +109,28 @@ extension ErrorMessageTests {
...
+
+extension ErrorMessageTests {
+  func testMispelledArgument_1() {
+    AssertErrorMessage(Qwz.self, ["--nme"], "Unknown option '--nme'. Did you mean '--name'?")
+    AssertErrorMessage(Qwz.self, ["-name"], "Unknown option '-name'. Did you mean '--name'?")
+  }
+  
+  func testMispelledArgument_2() {
+    AssertErrorMessage(Qwz.self, ["-ttle"], "Unknown option '-ttle'. Did you mean '-title'?")
+    AssertErrorMessage(Qwz.self, ["--title"], "Unknown option '--title'. Did you mean '-title'?")
```
natecook1000: This will be so helpful!

Sources/ArgumentParser/Parsing/CommandParser.swift
klaaspieter: I'd have to look into it. I'm actually still looking at this code as part of #23 so perhaps I can tacke lit there 😅

PR #14, opened 2020-02-28
Port to Android

buttaface:
Should also get it working for the BSDs, Haiku, etc., as long as they have a working Swift toolchain. Tested with the Swift 5.1.4 package for Android, termux/termux-packages#4895, all 230 tests pass with the following commands:
```
swift build -j 5 --enable-test-discovery
swift test -j 5 --parallel --num-workers 3 --enable-test-discovery
```
Note the space required after `-j`, hopefully this library doesn't have the same problem.


natecook1000:
Looks like a good change, thanks @buttaface!

PR #15, opened 2020-02-28
Use EX_USAGE instead of EXIT_FAILURE for validation errors

IngmarStein:
sysexits(3) documents the "preferable exit codes for programs". It defines

```
     EX_USAGE (64)         The command was used incorrectly, e.g., with the
                           wrong number of arguments, a bad flag, a bad syntax
                           in a parameter, or whatever.
```

So it seems to be a good choice for this use case.

### DIFF
--- a/Sources/ArgumentParser/Usage/MessageInfo.swift
+++ b/Sources/ArgumentParser/Usage/MessageInfo.swift
@@ -102,4 +102,12 @@ enum MessageInfo {
     case .validation, .other: return false
     }
   }
+
+  var exitCode: Int32 {
+    switch self {
+    case .help: return EXIT_SUCCESS
+    case .validation: return EX_USAGE
+    case .other: return EXIT_FAILURE
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


natecook1000:
Thanks, @IngmarStein! Let's discuss this a bit more — that error code looks appropriate for validation errors, but that exit path is for _any_ error thrown from either `run()` or `validate()`. Do you want to look at whether it would make sense for `MessageInfo` to provide the exit code, since it's already distinguishing between exiting cleanly or not?

IngmarStein:
Good suggestion, I've updated the PR accordingly. It might even make sense to replace `shouldExitCleanly` because it now only determines whether the message is printed to stdout or stderr.

Sources/ArgumentParser/Usage/MessageInfo.swift
```diff
@@ -102,4 +102,17 @@ enum MessageInfo {
...
+
+  var exitCode: Int32 {
+    switch self {
+    case .help: return EXIT_SUCCESS
+    case .validation:
+      #if os(Linux)
+      return EXIT_FAILURE
+      #else
+      return EX_USAGE
+      #endif
```
natecook1000: Do you need this platform conditional? [AFAICT](https://repl.it/repls/FlimsyUnwieldyStructure) Glibc defines `EX_USAGE` in the same way as Darwin.

IngmarStein: Thanks for verifying on that platform.

PR #17, opened 2020-02-28
Fixed Various Grammar & Spelling Mistakes

Wildchild9:


Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -104,10 +104,10 @@ extension Option where Value: ExpressibleByArgument {
...
   /// Parse the input after the option. Expect it to be a value.
   ///
-  /// For input such as `--foo foo` this would parse `foo` as the
-  /// value. However, for the input `--foo --bar foo bar` would
-  /// result in a error. Even though two values are provided, they don’t
-  /// succeed each option. Parsing would result in an error such as
+  /// For inputs such as `--foo foo`, this would parse `foo` as the
+  /// value. However, the input `--foo --bar foo bar` would
+  /// result in an error. Even though two values are provided, they don’t
+  /// succeed each option. Parsing would result in an error such as the following.
```
natecook1000: Let's use a colon when introducing a code sample this way.
```suggestion
  /// succeed each option. Parsing would result in an error such as the following:
```

Documentation/01 Getting Started.md
```diff
@@ -100,9 +100,9 @@ struct Count: ParsableCommand {
 }
 ```
 
-The `@Option` property wrapper denotes a command-line input that looks like `--name value`, deriving its name from the name of your property. 
+The `@Option` property wrapper denotes a command-line input that looks like `--name <value>`, deriving its name from the name of your property. 
 
-This interface has a trade-off for the users of our `count` tool: With `@Argument`, users don't need to type as much, but have to remember whether the input file or the output file needs to be given first. Using `@Option` makes the user type a little more, but the distinction between values is explicit. Options are order-independent, as well, so the user can name the input and output files in either order:
+This interface has a trade-off for the users of our `count` tool: With `@Argument`, users don't need to type as much but have to remember whether the input file or the output file needs to be given first. Using `@Option` makes the user type a little more, but the distinction between values is explicit. Options are order-independent, as well, so the user can name the input and output files in either order:
```
natecook1000: What do you think of this edit instead?
```suggestion
This interface has a trade-off for the users of our `count` tool: With `@Argument`, users don't need to type as much, but they have to remember whether to provide the input file or the output file first. Using `@Option` makes the user type a little more, but the distinction between values is explicit. Options are order-independent, as well, so the user can name the input and output files in either order:
```

Documentation/01 Getting Started.md
```diff
@@ -196,7 +196,7 @@ OPTIONS:
   -h, --help              Show help information.
 ```
 
-This is a great start — you can see that all the custom names are visible, and the help shows that values are expected for the `--input` and `--output` options. However, our custom options and flag don't have any descriptive text. Let's add that now, by passing string literals as the `help` parameter:
+This is a great start — you can see that all the custom names are visible, and the help shows that values are expected for the `--input` and `--output` options. However, our custom options and flag don't have any descriptive text. Let's add that now by passing string literals to the `help` parameter:
```
natecook1000: The documentation style we're using is to use "as" here.
```suggestion
This is a great start — you can see that all the custom names are visible, and the help shows that values are expected for the `--input` and `--output` options. However, our custom options and flag don't have any descriptive text. Let's add that now by passing string literals as the `help` parameter:
```

Wildchild9:
I'll add a commit with these changes. Thank you @natecook1000.

natecook1000:
Thanks, @Wildchild9!

PR #21, opened 2020-02-29
Clean up InputOrigin

sjavora:
This removes unused code from `InputOrigin` and replaces the lone `static func` with an `init` to better match the rest of the file.

### DIFF
--- a/Sources/ArgumentParser/Parsing/InputOrigin.swift
+++ b/Sources/ArgumentParser/Parsing/InputOrigin.swift
@@ -16,19 +16,11 @@
 /// This is usually an index into the `SplitArguments`.
 /// In some cases it can be multiple indices.
 struct InputOrigin: Equatable, ExpressibleByArrayLiteral {
-  enum Element: Comparable, Hashable {
+  enum Element: Hashable {
     case argumentIndex(SplitArguments.Index)
   }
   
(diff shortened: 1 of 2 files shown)


Sources/ArgumentParser/Parsing/InputOrigin.swift
```diff
@@ -16,19 +16,12 @@
 /// This is usually an index into the `SplitArguments`.
 /// In some cases it can be multiple indices.
 struct InputOrigin: Equatable, ExpressibleByArrayLiteral {
-  enum Element: Comparable, Hashable {
+
```
natecook1000: Can you close this line up?
```suggestion
```

sjavora: Done

sjavora:
The typealias required changes in a bunch more places, but it let us get rid of another `init` as well as an overload of `remove(at:)` in `SplitArguments` - thanks for the suggestion 👍 

I assumed the enum was there for planned future expansion - that's why I left it alone originally.

natecook1000:
Well, shoot. I think you're right that it's there to make room for being able to get values from the environment, instead of the command line. Could you revert that last change, and I'll merge the rest?

sjavora:
Done.

PR #22, opened 2020-02-29
Remove code that can be synthesized

sjavora:
This removes two types of redundant code:
- `internal` annotations (because they are the default)
- code that the compiler can synthesize, like memberwise initializers and `Equatable` conformances

Some `public init`s were removed, but those couldn't be accessed anyway, because they were nested in `internal` types - I've [complained](https://forums.swift.org/t/members-in-type-more-accessible-than-the-type-itself/24756) about the possibility of doing this before on the forums.

### DIFF
--- a/Sources/ArgumentParser/Parsing/Name.swift
+++ b/Sources/ArgumentParser/Parsing/Name.swift
@@ -29,19 +29,6 @@ enum Name: Equatable {
       self = .longWithSingleDash(String(baseName.dropFirst()))
     }
   }
-  
-  static func == (lhs: Name, rhs: Name) -> Bool {
-    switch (lhs, rhs) {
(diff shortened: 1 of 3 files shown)


xwu:
The use of explicit access levels is a common style choice that's widely used in first-party Swift code and recommended by at least a few style guides.

natecook1000:
The `internal` annotations should remain, even though they're inconsistently applied. We'd take a patch to apply them where they're missing to fix that inconsistency.

The code that can be synthesized is fair game, though!

sjavora:
Ok, I restored the `internal` annotations.

I would suggest adding some info about this to the README or some other file - in my personal experience, I have not come across a project that uses explicit `internal` and I suspect I'm not the only one.

Issue #4, opened 2020-02-28
Option parsing from environment variables


KyleLeneau:
It would be super nice/helpful to parse an options value from an Environment variable value as well taking what is passed in the command line call as the priority over the ENV variable. For CLI's that run in CI or automated environments this is super helpful as it would allow not having to print the value in the ENV variable and could keep it hidden. Not only that but it would be very convenient for repeated uses on commands!

```swift
import ArgumentParser

struct MyCommand: ParsableCommand {
    @Option(name: .shortAndLong, help: "The API Secret.", env: "SERVICE_API_SECRET")
    var secret: String
}
```

**Usage:**
`mycommand --secret QWERTY12345`
or
`mycommand` (where `SERVICE_API_SECRET='QWERTY12345'`

danieleggert:
Yes. We have been working on this already, and we’ll open a PR within the next week or two. There are still a lot of open questions about how to do this right, and we’d love to get feedback from everyone.

KyleLeneau:
Awesome, I am happy to provide some feedback on a PR!

KyleLeneau:
@danieleggert Any update on the progress of this?

danieleggert:
> @danieleggert Any update on the progress of this?

Please check out #109 — it’s almost done.

griffin-stewie:
any updates?

kojirou1994:
still waiting😭

philprime:
I this still on the roadmap?

fkorotkov:
@rauhul, do you have a preference on @KyleLeneau's proposal in this issue vs @danieleggert implementation in https://github.com/apple/swift-argument-parser/pull/109 vs something new like auto-magically populate values based on command names and argument names. For example, if we have a root command `tart` and sub-command `push` with a `--populate-cache` flag then set it automatically to a value of `TART_PUSH_POPULATE_CACHE` if defined?

rauhul:
I was thinking about a different approach where the parsing pipeline is composed of ArgumentValueProviders, the command line would be one such provider. I haven't fleshed out the idea but the benefit is that users could add their own argument providers depending on their use cases.

e.g. we could provide support out of the box for json config files, environment, `.env` files, NSUserDefaults, etc...

manicmachine:
Any word on this? #109 got closed during the Swift branch transition and doesn't look like a replacement was created.

kirilltitov:
So. Any progress on this one?

kirilltitov:
[...]

PR #29, opened 2020-03-01
Add a catch-all argument parsing strategy

natecook1000:
### Description
We don't currently have a way for `@Argument` arrays to capture command-line inputs that look like options. This capability is important for tools like SwiftPM that need to forward input to another command.

For example, the `swift-run` command needs to be able to parse all three of the arguments here into an array, including `--count`:

```
% swift run repeat --count 5 hello
``` 

### Detailed Design
This introduces an `ArgumentArrayParsingStrategy` enum with a `remaining` case that  matches the current behavior, as well as an `unconditionalRemaining` case that captures all remaining input.

```swift
/// The strategy to use when parsing multiple values from `@Option` arguments
/// into an array.
public enum ArgumentArrayParsingStrategy {
  /// Parse only unprefixed values from the command-line input, ignoring
  /// any inputs that have a dash prefix.
  ///
  /// For example, for a parsable type defined as following:
  ///
  ///     struct Options: ParsableArguments {
  ///         @Flag() var verbose: Bool
  ///         @Argument(parsing: .remaining) var words: [String]
  ///     }
  ///
  /// Parsing the input `--verbose one two` or `one two --verbose` would result
  /// in `Options(verbose: true, words: ["one", "two"])`. Parsing the input
  /// `one two --other` would result in an unknown option error for `--other`.
  ///
  /// This is the default strategy for parsing argument arrays.
  case remaining
  
  /// Parse all remaining inputs after parsing any known options or flags,
  /// including dash-prefixed inputs and the `--` terminator.
  ///
  /// For example, for a parsable type defined as following:
  ///
  ///     struct Options: ParsableArguments {
  ///         @Flag() var verbose: Bool
  ///         @Argument(parsing: .unconditionalRemaining) var words: [String]
  ///     }
  ///
  /// Parsing the input `--verbose one two --other` would include the `--other`
  /// flag in `words`, resulting in
  /// `Options(verbose: true, words: ["one", "two", "--other"])`.
  ///
  /// - Note: This parsing strategy can be surprising for users, particularly
  ///   when combined with options and flags. Prefer `remaining` whenever
  ///   possible, since users can always terminate options and flags with
  ///   the `--` terminator. With the `remaining` parsing strategy, the input
  ///   `--verbose -- one two --other` would have the same result as the above
  ///   example: `Options(verbose: true, words: ["one", "two", "--other"])`.
  case unconditionalRemaining
}
```

The two array-based `@Argument` initializers gain defaulted parameters for the parsing strategy:

```swift
  /// Creates a property that reads an array from zero or more arguments.
  ///
  /// The property has an empty array as its default value.
  ///
  /// - Parameters:
  ///   - parsingStrategy: The behavior to use when parsing multiple values
  ///     from the command-line arguments.
  ///   - help: Information about how to use this argument.
  public init<Element>(
    parsing parsingStrategy: ArgumentArrayParsingStrategy = .remaining,
    help: ArgumentHelp? = nil
  ) where Element: ExpressibleByArgument, Value == Array<Element>

  /// Creates a property that reads an array from zero or more arguments,
  /// parsing each element with the given closure.
  ///
  /// The property has an empty array as its default value.
  ///
  /// - Parameters:
  ///   - parsingStrategy: The behavior to use when parsing multiple values
  ///     from the command-line arguments.
  ///   - help: Information about how to use this argument.
  ///   - transform: A closure that converts a string into this property's
  ///     element type or throws an error.
  public init<Element>(
    parsing parsingStrategy: ArgumentArrayParsingStrategy = .remaining,
    help: ArgumentHelp? = nil,
    transform: @escaping (String) throws -> Element
  ) where Value == Array<Element>
```

### Documentation Plan
This includes documentation for the new enumeration and for the new parameters, as well as a new "Alternative positional argument parsing strategies" section in the _Arguments, Options, and Flags_ guide.

### Test Plan
End to end tests that use the `. unconditionalRemaining` strategy.

### Source Impact
This adds a defaulted parameter to two initializers, so it shouldn't have a significant source impact.

### Checklist

### DIFF
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -317,7 +317,6 @@ The default strategy for parsing options as arrays is to read each value from a 
 ```swift
 struct Example: ParsableCommand {
     @Option() var file: [String]
-    
     @Flag() var verbose: Bool
     
     func run() throws {
@@ -359,4 +358,36 @@ Finally, the `.remaining` parsing strategy uses all the inputs that follow the o
 Verbose: true, files: ["file1.swift", "file2.swift"]
 % example --file file1.swift file2.swift --verbose
 Verbose: false, files: ["file1.swift", "file2.swift", "--verbose"]
+```
+
+## Alternative positional argument parsing strategies
+
+The default strategy for parsing arrays of positional arguments is to ignore  all dash-prefixed command-line inputs. For example, this command accepts a `--verbose` flag and a list of file names as positional arguments:
+
+```swift
+struct Example: ParsableCommand {
+    @Flag() var verbose: Bool
+    @Argument() var files: [String]
+    
+    func run() throws {
+        print("Verbose: \(verbose), files: \(files)")
+    }
+}
+```
+
+The `files` argument array uses the default `.remaining` parsing strategy, so it only picks up values that don't have a prefix:
+
+```
+% example --verbose file1.swift file2.swift
+Verbose: true, files: ["file1.swift", "file2.swift"]
+% example --verbose file1.swift file2.swift --other
+Error: Unexpected argument '--other'
+Usage: example [--verbose] [<files> ...]
+```
+
+The `.unconditionalRemaining` parsing strategy uses whatever input is left after parsing known options and flags, even if that input is dash-prefixed. If `files` were defined as `@Argument(parsing: .unconditionalRemaining) var files: [String]`, then the resulting array would also include strings that look like options:
+
+```
+% example --verbose file1.swift file2.swift --other
+Verbose: true, files: ["file1.swift", "file2.swift", "--other"]
 ```
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -385,9 +385,18 @@ Error: Unexpected argument '--other'
 Usage: example [--verbose] [<files> ...]
 ```
 
-The `.unconditionalRemaining` parsing strategy uses whatever input is left after parsing known options and flags, even if that input is dash-prefixed. If `files` were defined as `@Argument(parsing: .unconditionalRemaining) var files: [String]`, then the resulting array would also include strings that look like options:
+Any input after the `--` terminator is automatically treated as positional input, so users can provide dash-prefixed values that way even with the default configuration:
+
+```
+% example --verbose -- file1.swift file2.swift --other
+Verbose: true, files: ["file1.swift", "file2.swift", "--other"]
+```
+
+The `.unconditionalRemaining` parsing strategy uses whatever input is left after parsing known options and flags, even if that input is dash-prefixed, including the terminator itself. If `files` were defined as `@Argument(parsing: .unconditionalRemaining) var files: [String]`, then the resulting array would also include strings that look like options:
 
 ```
 % example --verbose file1.swift file2.swift --other
 Verbose: true, files: ["file1.swift", "file2.swift", "--other"]
+% example -- --verbose file1.swift file2.swift --other
+Verbose: false, files: ["--", "--verbose", "file1.swift", "file2.swift", "--other"]
 ```
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -104,7 +104,7 @@ public enum ArgumentArrayParsingStrategy {
   case remaining
   
   /// Parse all remaining inputs after parsing any known options or flags,
-  /// including dash-prefixed inputs.
+  /// including dash-prefixed inputs and the `--` terminator.
   ///
   /// For example, for a parsable type defined as following:
   ///
@@ -118,8 +118,11 @@ public enum ArgumentArrayParsingStrategy {
   /// `Options(verbose: true, words: ["one", "two", "--other"])`.
   ///
   /// - Note: This parsing strategy can be surprising for users, particularly
-  ///   when combined with options and flags. Use care when specifying the
-  ///   `unconditionalRemaining` parsing strategy.
+  ///   when combined with options and flags. Prefer `remaining` whenever
+  ///   possible, since users can always terminate options and flags with
+  ///   the `--` terminator. With the `remaining` parsing strategy, the input
+  ///   `--verbose -- one two --other` would have the same result as the above
+  ///   example: `Options(verbose: true, words: ["one", "two", "--other"])`.
   case unconditionalRemaining
 }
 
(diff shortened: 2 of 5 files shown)


Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -83,6 +83,42 @@ extension Argument where Value: ExpressibleByArgument {
...
+  ///
+  ///     struct Options: ParsableArguments {
+  ///         @Flag() var verbose: Bool
+  ///         @Argument(parsing: .remainingValues) var words: [String]
+  ///     }
+  ///
+  /// Parsing the input `--verbose one two --other` would include the `--other`
+  /// flag in `words`, resulting in
+  /// `Options(verbose: true, words: ["one", "two", "--other"])`.
+  case unconditionalRemaining
```
danieleggert: `ArrayParsingStrategy.remaining` has some guidance as to how this usually should be done (i.e. using a `--` terminator on the input). We should probably add a link here to that discussion and make sure they line up. And generally give some guidance as to when `.remaining` vs. `. unconditionalRemaining` is a good choice for developing new tools. I’d argue that `.remaining` is the _correct_ choice and have users of the tool use the `--` escape / termination.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
```diff
@@ -356,27 +356,35 @@ extension ArgumentSet {
...
       case let .option(parsed):
+        // Look for an argument that matches this `--option` or `-o`-style
+        // input. If we can't find one, just move on to the next input. We
+        // defer catching leftover arguments until we've fully extracted all
+        // the information for the selected command.
         guard
           let argument: ArgumentDefinition = try? first(matching: parsed, at: origin)
-          else { continue }
+          else {
+            unusedOptions.append((origin, String(describing: parsed)))
```
danieleggert: I’m not sure if `String(describing: parsed)` is the right approach. For #4 I’ve been adding something like
```swift
extension SplitArguments {
    func original(at origin: InputOrigin.Element) -> String {
        switch origin {
        case .argumentIndex(let index):
            return originalInput[index.inputIndex.rawValue]
        }
    }
}
```
and then use it here like
```swift
unusedOptions.append((origin, all.original(at: origin)))
```
such that this will actually get the string from the original input — regardless of how it was parsed. But I could also try to clean up these uses as part of #4 if that makes more sense.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
```diff
@@ -394,7 +402,7 @@ extension ArgumentSet {
     // We have parsed all non-positional values at this point.
     // Next: parse / consume the positional values.
     do {
-      try parsePositionalValues(from: positionalValues, into: &result)
+      try parsePositionalValues(positionalValues, unusedOptions: unusedOptions, into: &result)
```
danieleggert: I’m not sure, but: Would it make sense to interleave `positionalValues` and `unusedOptions` to get the original ordering? Possibly even use the `SplitArguments` type for this, since it already is aware of “a list of things that we parsed and their index into the original input array”…? I’m not sure.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
natecook1000: Way better not to set up this implicit dependency on `Parsed`'s string representation, thanks.

Sources/ArgumentParser/Parsable Properties/Argument.swift
natecook1000: I added a note here to give a bit more context. When we have documentation of the `--` behavior, let's add a link from here as well.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
natecook1000: I ended up keeping the separate arrays, but changing the way they're consumed in `parsePositionalValues`. We can look at this from a performance standpoint later, if it's a problem.

Documentation/02 Arguments, Options, and Flags.md
```diff
@@ -360,3 +359,35 @@ Verbose: true, files: ["file1.swift", "file2.swift"]
...
+```
+% example --verbose file1.swift file2.swift
+Verbose: true, files: ["file1.swift", "file2.swift"]
+% example --verbose file1.swift file2.swift --other
+Error: Unexpected argument '--other'
+Usage: example [--verbose] [<files> ...]
+```
+
+The `.unconditionalRemaining` parsing strategy uses whatever input is left after parsing known options and flags, even if that input is dash-prefixed. If `files` were defined as `@Argument(parsing: .unconditionalRemaining) var files: [String]`, then the resulting array would also include strings that look like options:
+
```
danieleggert: Minor nitpick, but I’d show the example here where `--verbose` is towards the end, and how it ends up being one of `files`, to make developers aware of that trade-off.

Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -83,6 +83,46 @@ extension Argument where Value: ExpressibleByArgument {
...
+  ///         @Argument(parsing: .remainingValues) var words: [String]
+  ///     }
+  ///
+  /// Parsing the input `--verbose one two --other` would include the `--other`
+  /// flag in `words`, resulting in
+  /// `Options(verbose: true, words: ["one", "two", "--other"])`.
+  ///
+  /// - Note: This parsing strategy can be surprising for users, particularly
+  ///   when combined with options and flags. Use care when specifying the
+  ///   `unconditionalRemaining` parsing strategy.
```
danieleggert: I’d explicitly call out the alternative here:
```suggestion
  /// - Note: This parsing strategy can be surprising for users, particularly
  ///   when combined with options and flags. Use care when specifying the
  ///   `unconditionalRemaining` parsing strategy.
  ///   It’s generally recommended to use `.remaining`, and tell users to
  ///   _terminate_ options with `--` if any arguments may start with `-` or `--`,
  ///   i.e. use `.remaining` and tell users to call the tool with the input
  ///   `--verbose -- one two --other`.
```

Tests/EndToEndTests/RepeatingEndToEndTests.swift
```diff
@@ -300,6 +300,42 @@ extension RepeatingEndToEndTests {
...
+    }
+    
+    AssertParse(Foozle.self, ["one", "two", "three", "--other", "--verbose"]) { foozle in
+      XCTAssertTrue(foozle.verbose)
+      XCTAssertEqual(foozle.names, ["one", "two", "three", "--other"])
+    }
+    
+    AssertParse(Foozle.self, ["--verbose", "--other", "one", "two", "three"]) { foozle in
+      XCTAssertTrue(foozle.verbose)
+      XCTAssertEqual(foozle.names, ["--other", "one", "two", "three"])
```
danieleggert: Would be good to add a test case that include `--`, e.g.
```swift
AssertParse(Foozle.self, ["--verbose", "one", "--", "--other", "two", "three"]) { foozle in
```
in order to make sure we don’t change/regress the `--` terminator behaviour with `.unconditionalRemaining`.

Documentation/02 Arguments, Options, and Flags.md
natecook1000: `--verbose` gets parsed as a flag no matter where it's placed. I'll add a note that it makes the most sense to use `.unconditionalRemaining` when it's on the only property in the command.

PR #33, opened 2020-03-01
Allow run to be mutating

nuclearace:
I noticed that I was allowed to mark the run method as mutating, however, this caused issues where it was always thinking it was expecting arguments... which is strange. 

But is there a reason to not mark this as possibly mutating?


nuclearace:
Hmm, turns out my initial problem is unrelated to the mutableness of this guy

natecook1000:
Thanks for this, @nuclearace! I'd like to experiment with this a bit first and see the effect of making `run()` mutating.

natecook1000:
Superseded by #163.

PR #40, opened 2020-03-02
Fix a code example

glessard:
`ParsableCommand.parseAsRoot()` drops the first item from `CommandLine.arguments` when it receives a `nil` array as a parameter, so it is necessary to do the same before pre-processing arguments to feed to any `ParsableCommand`.


natecook1000:
Thanks, @glessard! 👏

PR #42, opened 2020-03-02
port to Windows

compnerd:
This adjusts the imports to enable building for Windows.

The major change here is that `_terminalHeight` and `_terminalWidth`
have been merged into a `_terminalSize` which returns a tuple of the
width and the height.  This is then implemented for Windows as well.


natecook1000:
Looks great, thanks @compnerd!

PR #44, opened 2020-03-02
build: add a CMake based build for Windows

compnerd:
Since swift-package-manager doesn't work on Windows yet, add a CMake
based build system.


natecook1000:
I had errors when trying to generate an Xcode project and build, so I'll trust you that this works on Windows.

PR #47, opened 2020-03-02
treat empty default strings and nil default strings the same way in help output

glessard:
Previously, if one defined an option thusly:
```
  @Option(name: .long, default: "", help: "help text")
  var option: String
```

the corresponding line of help output would be:
```
  --option <option>       help text (default: )
```

After this patch, the output becomes:
```
  --option <option>       help text
```

This did not affect the tests.


glessard:
I added to `testHelpWithDefaultValueButNoDiscussion`. Let me know if you'd prefer the additions elsewhere.

natecook1000:
Thanks, @glessard! :clap::clap:

PR #48, opened 2020-03-02
Don't show Optional(x) when displaying default values in help.

### LINKED ISSUE #26: Default Option value in help text includes `Optional(…)`
the following option

```swift
@Option(default: true)
var doStuff: Bool?
```

would result in the following default value description:

```
(default: Optional(true))
```

This is counterintuitive and requires user to workaround it.

sjavora:
Fixes #26

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
+++ b/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
@@ -66,3 +66,16 @@ extension Double: ExpressibleByArgument {}
 
 extension Bool: ExpressibleByArgument {}
 
+extension ExpressibleByArgument {
+
+  var defaultValueDescription: String {
+
+    let mirror = Mirror(reflecting: self)
+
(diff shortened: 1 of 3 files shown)


sjavora:
@natecook1000 could you please take a look?

natecook1000:
I've been thinking about the right approach to #26, since it's not actually a configuration that we want people to use. If you provide a non-`nil` default value, the property will never be `nil`, so it really shouldn't be optional in the first place. We may actually need to drop `ExpressibleByArgument` conformance from `Optional` and add an initializer specifically for optional values. 

In any case, it would be great to fix this, but I don't want to add public API to do so. Can you look for a solution that could be handled internally to the library?

sjavora:
I've removed the protocol requirement – the implementation now uses `Mirror` to determine whether the type is optional. I'm not thrilled about it, but didn't come up with anything better.

sjavora:
@natecook1000 fixed conflicts - please let me know if this approach is OK from your POV 🙂

natecook1000:
Let's do this!

Issue #49, opened 2020-03-02
Unhelpful error output when using two mutually exclusive flags


glessard:
I defined a set of mutually exclusive flags thusly:
```
enum OutputBehaviour: String, CaseIterable { case stats, count, list }

struct Options: ParsableCommand {
  @Flag(default: .list, help: "Program output")
  var behaviour: OutputBehaviour
}

let options = Options.parseOrExit()
```

Upon misusing the command by using two of these flags, the error message is unhelpful:
```
$ ./options  --count --list
Error: Value at position InputOrigin(_elements: Set([ArgumentParser.InputOrigin.Element.argumentIndex(1)])) has already been set from value at position InputOrigin(_elements: Set([ArgumentParser.InputOrigin.Element.argumentIndex(0)])).
Usage: options [--stats] [--count] [--list]
```

The error message should print out the conflicting flag names instead.

PR #52, opened 2020-03-03
Pass information to ParserError in order to show user a better error message

glessard:
This is an attempt to fix https://github.com/apple/swift-argument-parser/issues/49

Questions about this proposed fix:
- Would it be better to simply extract the strings at the point where the error is thrown?
  I touched every place where "duplicateExclusiveValues" appears, so it's not clear that error case would have another use than this one.
- Is the proposed replacement message any good?

### DIFF
--- a/Sources/ArgumentParser/Usage/UsageGenerator.swift
+++ b/Sources/ArgumentParser/Usage/UsageGenerator.swift
@@ -188,8 +188,8 @@ extension ErrorMessageGenerator {
       return unexpectedValueForOptionMessage(origin: o, name: n, value: v)
     case .unexpectedExtraValues(let v):
       return unexpectedExtraValuesMessage(values: v)
-    case .duplicateExclusiveValues(previous: let previous, duplicate: let duplicate):
-      return duplicateExclusiveValues(previous: previous, duplicate: duplicate)
+    case .duplicateExclusiveValues(previous: let previous, duplicate: let duplicate, originalInput: let arguments):
+      return duplicateExclusiveValues(previous: previous, duplicate: duplicate, arguments: arguments)
     case .noValue(forKey: let k):
       return noValueMessage(key: k)
     case .unableToParseValue(let o, let n, let v, forKey: let k):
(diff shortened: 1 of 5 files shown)


glessard:
This conflicts with the changes in https://github.com/apple/swift-argument-parser/pull/21, where the internals of `InputOrigin` were closed off.

Tests/UnitTests/ErrorMessageTests.swift
```diff
@@ -134,3 +134,19 @@ extension ErrorMessageTests {
...
+  var behaviour: OutputBehaviour
+
+  @Flag() var bool: Bool
+}
+
+extension ErrorMessageTests {
+  func testDuplicateFlags() {
+    AssertErrorMessage(Options.self, ["--list", "--bool", "-s"], "Value to be set with flag \"-s\" had already been set with flag \"--list\"")
+    AssertErrorMessage(Options.self, ["-cbl"], "Value to be set with flag \"l\" in \"-cbl\" had already been set with flag \"c\" in \"-cbl\"")
+    AssertErrorMessage(Options.self, ["-bc", "--stats", "-l"], "Value to be set with flag \"--stats\" had already been set with flag \"c\" in \"-bc\"")
```
natecook1000: Nice work, these are great messages!

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -301,8 +301,21 @@ extension ErrorMessageGenerator {
     }
   }
   
-  func duplicateExclusiveValues(previous: InputOrigin, duplicate: InputOrigin) -> String? {
-    return "Value at position \(duplicate) has already been set from value at position \(previous)."
+  func duplicateExclusiveValues(previous: InputOrigin, duplicate: InputOrigin, arguments: [String]) -> String? {
+    func elementString(_ origin: InputOrigin, _ arguments: [String]) -> String? {
+      guard case .argumentIndex(let split) = origin.elements.first else { return nil }
+      var argument = "\"\(arguments[split.inputIndex.rawValue])\""
```
natecook1000: We're using single quotes around user input in the other messages, can you do the same here?

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -301,8 +301,21 @@ extension ErrorMessageGenerator {
...
+      if case let .sub(offsetIndex) = split.subIndex {
+        let stringIndex = argument.index(argument.startIndex, offsetBy: offsetIndex+2)
+        argument = "\"\(argument[stringIndex])\" in \(argument)"
+      }
+      return "flag \(argument)"
+    }
+
+    let dupeString = elementString(duplicate, arguments) ?? "position \(duplicate)"
```
natecook1000: Can you add to-do comment here to about updating this when we support environment values? It's worth noting that the RHS of these coalescing operators won't ever get hit.

Sources/ArgumentParser/Usage/UsageGenerator.swift
glessard: I had failed to notice. Done!

Sources/ArgumentParser/Usage/UsageGenerator.swift
glessard: Done

glessard:
I reinstated some of what had been removed in 6b3a0a1, because it was what I'd used to peek in and extract the necessary info. If there's another method, I can use that, but I haven't seen it.

PR #58, opened 2020-03-04
Constrain `ExpressibleByArgument` implementation for `LosslessStringC…

Lantua:
…onvertible`.

This extension provides default implementation for `ExpressibleByArgument` to types conforming to `LosslessStringConvertible`.
It should be restricted only to ones that wants to conform to `ExpressibleByArgument`.


Lantua:
From discussion here:

https://forums.swift.org/t/why-is-losslessstringconvertible-init-argument-non-conditional/34289/

natecook1000:
Thanks for this fix, @Lantua!

Issue #56, opened 2020-03-04
Default value not displayed in help for Flags built from CaseIterable enums


glessard:
When using flags built from a `(String & CaseIterable)` enum, such as this:
```
enum OutputBehaviour: String, CaseIterable { case stats, count, list }

struct Options: ParsableCommand {
  @Flag(name: .short, default: .list, exclusivity: .chooseFirst, help: "Help!")
  var behaviour: OutputBehaviour
}

let options = Options.parseOrExit()
```

The corresponding line in the help message does not display the default value.

The Flag initializer passes the default value along to the ArgumentDefinition constructor, but when the array of ArgumentDefinition is read back in HelpGenerator, the `help.defaultValue` properties for these is `nil`.

PR #63, opened 2020-03-05
output default value in help for Flags built from CaseIterable enums

glessard:
This fixes https://github.com/apple/swift-argument-parser/issues/56


glessard:
That test is now updated in 749cb6f.

PR #65, opened 2020-03-05
Add an EnumerableFlag protocol

natecook1000:
### Description
This addresses the need for providing name specifications for `enum` flags, since property wrappers can't be used for enumeration cases.

### Detailed Design
This includes a new `EnumerableFlag` protocol:

```swift
/// A type that represents the different possible flags to be used by a
/// `@Flag` property.
public protocol EnumerableFlag: CaseIterable {
  /// Returns the name specification to use for the given flag.
  static func name(for value: Self) -> NameSpecification

  /// Returns the help information to show for the given flag.
  static func help(for value: Self) -> ArgumentHelp?
}
```

As well as new `EnumerableFlag`-constrained `@Flag` initializers that replace the current ones that are constrained to `String`/`CaseIterable`.

### Documentation Plan
Wrote type- and symbol-level documentation for `EnumerableFlag`, revised docs for the new initializers, and updated the "Argument, Options, and Flags" guide.

### Test Plan
Modified unit tests to use the new protocol and added tests that use the original, deprecated versions.

### Source Impact
This deprecates the `@Flag` initializers that are constrained to `CaseIterable`/`RawRepresentable` and `RawValue == String`. These initializers will continue to work, and can be removed in a future version. 

### Checklist

### DIFF
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -211,7 +211,7 @@ false false
 Error: Missing one of: '--enable-required-element', '--disable-required-element'
 ```
 
-To create a flag with custom names for a Boolean value, to provide an exclusive choice between more than two names, or for collecting multiple values from a set of defined choices, define an enumeration that conforms to the `EnumerableFlag` documentation.
+To create a flag with custom names for a Boolean value, to provide an exclusive choice between more than two names, or for collecting multiple values from a set of defined choices, define an enumeration that conforms to the `EnumerableFlag` protocol.
 
 ```swift
 enum CacheMethod: EnumerableFlag {
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -115,10 +115,10 @@ struct Example: ParsableCommand {
 
 Arguments and options can be parsed from any type that conforms to the `ExpressibleByArgument` protocol. Standard library integer and floating-point types, strings, and Booleans all conform to `ExpressibleByArgument`.
 
-You can make your own custom types conform to `ExpressibleByArgument` by implementing `init(argument:)`:
+You can make your own custom types conform to `ExpressibleByArgument` by implementing `init?(argument:)`:
 
 ```swift
-struct Path {
+struct Path: ExpressibleByArgument {
     var pathString: String
     
     init?(argument: String) {
@@ -178,7 +178,7 @@ struct Example: ParsableCommand {
 }
 ```
 
-Throw an error from the `transform` function to indicate that the user provided an invalid value for that type.
+Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. See [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) for more about customizing `transform` function errors.
 
 ## Using flag inversions, enumerations, and counts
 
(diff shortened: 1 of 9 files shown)


Sources/ArgumentParser/Parsable Properties/Flag.swift
```diff
@@ -240,15 +242,118 @@ extension Flag where Value: CaseIterable, Value: RawRepresentable, Value.RawValu
...
+
+// - MARK: Deprecated CaseIterable/RawValue == String
+
+extension Flag where Value: CaseIterable, Value: RawRepresentable, Value.RawValue == String {
+  /// Creates a property that gets its value from the presence of a flag,
+  /// where the allowed flags are defined by a case-iterable type.
   ///
   /// - Parameters:
   ///   - name: A specification for what names are allowed for this flag.
+  ///   - initial: A default value to use for this property. If `initial` is
```
thomasvl: Drive by - "default", not "initial", right?  The second sentence also doesn't seem right.

natecook1000: Parameter names are used for documentation, not the argument labels, so `initial` is correct here. How would you word the second sentence? Would something like "If `initial` is `nil`, the flag declared with this `@Flag` attribute is not required by the user." make more sense?

thomasvl: Re param names - odd, for doc generation I thought it was the public name that was always listed.  https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/SingleLineComment.html seems to imply that since it shows the name listed in the generated docs.

Second sentence - yea, the fact that it just said _flag_ was what threw me off, I was thinking it was referring to another parameter or something.

natecook1000: Right, for that declaration the parameter name and argument label are the same. We always use the parameter name since some parameters don’t have argument labels (e.g. `func foo(_ v: Int)`).

I’ll revise that second sentence throughout, I think we’re using it in a pretty standard way. Thanks for the feedback!

natecook1000:
@swift-ci Please test

Documentation/02 Arguments, Options, and Flags.md
```diff
@@ -211,15 +211,15 @@ false false
 Error: Missing one of: '--enable-required-element', '--disable-required-element'
 ```
 
-You can also use flags with types that are `CaseIterable` and `RawRepresentable` with a string raw value. This is useful for providing custom names for a Boolean value, for an exclusive choice between more than two names, or for collecting multiple values from a set of defined choices.
+To create a flag with custom names for a Boolean value, to provide an exclusive choice between more than two names, or for collecting multiple values from a set of defined choices, define an enumeration that conforms to the `EnumerableFlag` documentation.
```
kylemacomber, line 214:
```suggestion
To create a flag with custom names for a Boolean value, to provide an exclusive choice between more than two names, or for collecting multiple values from a set of defined choices, define an enumeration that conforms to the `EnumerableFlag`.
```

natecook1000: 🤦 Thanks!

kylemacomber:
At first I thought the Array case was counter intuitive:

```swift
enum Color: EnumerableFlag {
    case pink, purple, silver
}

   @Flag() var colors: [Color]
   
   func run() throws {
       print(colors)
   }
}
```

```
% example --pink --silver
[.pink, .silver]
```

... and was going to suggest it be broken out into a separate property wrapper, but I see the synergy with the Int use case.

kylemacomber:
Seems like we could accept any `RangeReplaceableCollection`. And also maybe `SetAlgebra`? It's a little weird to conform an `OptionSet` to `CaseIterable`, but that's a more efficient storage representation for your `Color` example.

natecook1000:
> Seems like we could accept any RangeReplaceableCollection. And also maybe SetAlgebra? It's a little weird to conform an OptionSet to CaseIterable, but that's a more efficient storage representation for your Color example.

We could definitely make these generic over `RangeReplaceableCollection`, though we'd want to do that across the board (i.e. for `@Option` and `@Argument` types, too). `SetAlgebra` would require another set of overloads, since we don't have an abstraction that ties the two "create an empty one and then add things to it" protocols.

natecook1000:
@swift-ci Please test

PR #70, opened 2020-03-06
simplify a switch statement

glessard:
This change has no effect on functionality, documentation or testing.

The switch in `ArgumentSet.flagUpdate` (originally in one of the Flag constructors) switches on 3 items, one of which (the return value of `values.element(forKey:)`) is needed in only one of the 12 possibilities. This moves the call to `values.element(forKey:)` into the one `case` where it is needed. The function becomes slightly more efficient, and the `switch` is more readable (down to 6 possibilities), and no longer needs a `default` case. Note that the probably-impossible failure mode where `values.element(forKey: key` returns `nil` and `hasUpdated` is `true` still results in the same behaviour, namely the argument is unused and the function returns `true`; this eventually leads to a call to `ErrorMessageGenerator.unknownOptionMessage`.

### Checklist


glessard:
Superseded by https://github.com/apple/swift-argument-parser/pull/71

Issue #59, opened 2020-03-04
Repeated flag causes error in some cases


glessard:
When repeating a flag defined from a mutually-exclusive set, one gets an error.
```
enum Enum: String, CaseIterable { case s, t, u }

struct Options: Parsable Command {
  @Flag(name: .short, exclusivity: .exclusive, help: "Help!")
  var flag: Enum
}

let options = Options.parseOrExit()
```
```
$ options -s -s
Error: Value to be set with flag '-s' had already been set with flag '-s'
Usage: options -s -t -u
```

Should repetition in such a case simply be ignored?

natecook1000:
Allowing it sounds like the right behavior to me — can you think of circumstances where it would be semantically important for the command to fail instead?

glessard:
No, I can't think of such circumstances. Logic seems to dictate there shouldn't be, either...

PR #71, opened 2020-03-07
Repeated flags fix

glessard:
Stop failing upon seeing a repeated value from a mutually-exclusive set of flags.

This adds an `Equatable` requirement to the `Flag` constructors from `CaseIterable` enums. Given the other requirements on those enums, `Equatable` conformance is automatic and therefore costs nothing.

The switch statement in `ArgumentSet.updateFlag` is also much simplified, by calling `values.element(forKey:)` only when it is needed.

### Checklist

This fixes https://github.com/apple/swift-argument-parser/issues/59
This supersedes https://github.com/apple/swift-argument-parser/pull/70

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Flag.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Flag.swift
@@ -226,7 +226,7 @@ extension Flag where Value == Int {
   }
 }
 
-extension Flag where Value: CaseIterable, Value: RawRepresentable, Value.RawValue == String {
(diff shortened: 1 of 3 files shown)


glessard:
Noting the draft PR https://github.com/apple/swift-argument-parser/pull/65, specifying the new `EnumerableFlag` to require `Equatable` would not be onerous, I think. It would be automatically generated in an overwhelming majority of uses.

natecook1000:
@glessard This looks great — I think adding the `Equatable` constraint is fine here, and I'll add that constraint to `EnumerableFlag` as well.

Can you update with latest and I'll merge?

glessard:
I rebased and squashed. Thanks!

Issue #18, opened 2020-02-29
Target named "TestHelpers" conflict with "swift-se0270-range-set" unit-test target


dehengxu:
As title mentioned

natecook1000:
Hi @dehengxu — I'm having trouble reproducing this error. Could you say more about what you're trying to do when you run into this?

dehengxu:
> Hi @dehengxu — I'm having trouble reproducing this error. Could you say more about what you're trying to do when you run into this?

I use packages  ArgumentParser and swift-se0270-range-set which both have TestHelpers target

PR #74, opened 2020-03-09
Prefix test target names with "ArgumentParser"

elliottwilliams:
Similar to https://github.com/apple/swift-argument-parser/commit/d8fb6bdd63fa541e0ea52a99556093e1c3a60586 and #18.

Prior to Swift 5.2, test target names need to be unique throughout the entire package graph. This change adds ~"SAP"~ "ArgumentParser" to each test target to avoid ambiguities.

Context: I'm trying to use swift-argument-parser in a package repo which also depends on [Carthage](https://github.com/Carthage/Carthage), and they have [a nested dependency with a `UnitTests` target](https://github.com/Carthage/Carthage#carthagekit). Since the argument parser is under active development and I assume I'm not the only one in this situation, I figured I would propose the change here.

### Checklist
- [n/a] I've added at least one test that validates that my change is working, if appropriate
- [n/a] I've followed the code style of the rest of the project
- [n/a] I've updated the documentation if necessary


natecook1000:
Thanks for getting this workaround ready, @elliottwilliams! Looking at all the names, we'd rather go with an `ArgumentParser` prefix for everything. Would you mind adopting that prefix, including renaming `SAPTestHelpers`?

elliottwilliams:
@natecook1000 sure thing, updated!

natecook1000:
Thanks, @elliottwilliams! Merged locally.

Issue #64, opened 2020-03-05
Add support for default Argument value


griffin-stewie:
I hope `@ Argument` has a default value.

This is useful when a command needs a path argument to do some work on the path. The user can pass any path or omit it, in which case the current directory will be the path. It is similar to the `ls` command.

```swift
// pseudo code
@Argument(default: ".")
var path: URL
```

## Related thread

[Default Argument values? \- Related Projects / ArgumentParser \- Swift Forums](https://forums.swift.org/t/default-argument-values/34174)

griffin-stewie:
I implemented this feature in #75

PR #75, opened 2020-03-10
Add support for default Argument value

griffin-stewie:
### Description

I added support default value for `@Argument` property wrapper.

Related issue is #64. and Swift Forums thread is [`Default Argument values?  - Related Projects / ArgumentParser`](https://forums.swift.org/t/default-argument-values/34174)

### Detailed Design

The way to implement is same as `@Option` property wrapper. Just add a optional `default` parameter on its `init` method and default value is nil. `ArgumentSet` and` ArgumentDefinition` do all the work in the `init` method.

### Documentation Plan

> How has the new feature been documented? 

I added documents on methods I changed.

> Have the relevant portions of the guide been updated in addition to symbol-level documentation?

I fixed a bit talking about `default` parameter on `02 Arguments, Options, and Flags.md`.

### Test Plan

> How is the new feature tested?

I added tests similar to the existing test code to `DefaultsEndToEndTests.swift`.

For example, if there is no argument when the `Default` value is given Whether the` Default` value is stored. If there is an argument when giving the `Default` value, whether the value will be overridden properly.

### Source Impact

> What is the impact of this change on existing users?

`default` value is Optional and default parameter is nil, so no impact on existing users.

> Does it deprecate or remove any existing API?

No it doesn't.

### Checklist

### DIFF
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -43,7 +43,7 @@ This example shows how `ArgumentParser` provides defaults that speed up your ini
 - Option and flag names are derived from the names of your command's properties.
 - Whether arguments are required and what kinds of inputs are valid is based on your properties' types.
 
-In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option or flag with a `default` parameter can also be omitted by the user.
+In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option, flag or argument with a `default` parameter can also be omitted by the user.
 
 Users must provide values for all properties with no implicit or specified default. For example, this command would require one integer argument and a string with the key `--user-name`.
 
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -43,7 +43,7 @@ This example shows how `ArgumentParser` provides defaults that speed up your ini
 - Option and flag names are derived from the names of your command's properties.
 - Whether arguments are required and what kinds of inputs are valid is based on your properties' types.
 
-In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option, flag or argument with a `default` parameter can also be omitted by the user.
+In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option, flag, or argument with a `default` parameter can also be omitted by the user.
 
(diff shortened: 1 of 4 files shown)


griffin-stewie:
I'm thinking about how to add test code. Should I add the same content as  `OptionalEndToEndTests.swift` as` ArgumentEndToEndTests.swift`?

griffin-stewie:
Should I add it to [PositionalEndToEndTests.swift](https://github.com/apple/swift-argument-parser/blob/master/Tests/EndToEndTests/PositionalEndToEndTests.swift)?

natecook1000:
Looking good! I would add tests to `DefaultEndToEndTests.swift`.

Also, since this will change the public API, we need to use this pull request template instead: https://raw.githubusercontent.com/apple/swift-argument-parser/master/.github/PULL_REQUEST_TEMPLATE/NEW.md

griffin-stewie:
I see. I'll give it a try.

griffin-stewie:
I added test code and fixing documentation a bit. Could you review this?

Documentation/02 Arguments, Options, and Flags.md
```diff
@@ -43,7 +43,7 @@ This example shows how `ArgumentParser` provides defaults that speed up your ini
 - Option and flag names are derived from the names of your command's properties.
 - Whether arguments are required and what kinds of inputs are valid is based on your properties' types.
 
-In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option or flag with a `default` parameter can also be omitted by the user.
+In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option, flag or argument with a `default` parameter can also be omitted by the user.
```
natecook1000: One more comma here:
```suggestion
In this example, all of the properties have default values — Boolean flags always default to `false`, optional properties default to `nil`, and arrays default to an empty array. An option, flag, or argument with a `default` parameter can also be omitted by the user.
```

Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -73,12 +73,15 @@ extension Argument where Value: ExpressibleByArgument {
   /// If the property has an `Optional` type, the argument is optional and
   /// defaults to `nil`.
   ///
-  /// - Parameter help: Information about how to use this argument.
+  /// - Parameters
```
natecook1000:
```suggestion
  /// - Parameters:
```

griffin-stewie:
@natecook1000 
Thank you for reviewing, I fixed documentation. I'll add more tests on UsageGenerationTests.

griffin-stewie:
@natecook1000
I added test code that verifies the usage string is generated correctly at c54ac05. I fixed all the issues pointed out to you.

natecook1000:
Looks great, @griffin-stewie! 👏👏

PR #79, opened 2020-03-13
Add an API for converting an error to an exit code

natecook1000:
### Description
This provides a method on all `ParsableArguments`/`ParsableCommand` types that lets you retrieve the exit code associated with an error.

### Detailed Design
1) Add an `exitCode(for:)` method that converts an error to an `ExitCode` instance. 

```swift
extension ParsableArguments {
  /// Returns the exit code for the given error.
  ///
  /// - Parameter error: An error to generate a message for.
  /// - Returns: The exit code for the `error`.
  public static func exitCode(for error: Error) -> ExitCode
}
```

2) Upgrade the `ExitCode` type to be a bit more useful, making it `RawRepresentable` and  `Hashable`. The represented code is now available through the `rawValue` property.

### Documentation Plan
I included symbol documentation for the new APIs.

### Test Plan
I added unit tests for the `exitCode(for)` method and converted the integration test helper to use `ExitCode`s instead of plain `Int32`s.

### Source Impact
No source impact or deprecations.

### Checklist

### DIFF
--- a/Sources/ArgumentParserTestHelpers/TestHelpers.swift
+++ b/Sources/ArgumentParserTestHelpers/TestHelpers.swift
@@ -136,7 +136,7 @@ extension XCTest {
   public func AssertExecuteCommand(
     command: String,
     expected: String? = nil,
-    exitCode: Int32 = ExitCode.success.rawValue,
+    exitCode: ExitCode = .success,
     file: StaticString = #file, line: UInt = #line)
   {
     let splitCommand = command.split(separator: " ")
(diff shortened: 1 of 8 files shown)


Sources/ArgumentParserTestHelpers/TestHelpers.swift
```diff
@@ -136,7 +136,7 @@ extension XCTest {
   public func AssertExecuteCommand(
     command: String,
     expected: String? = nil,
-    exitCode: Int32 = 0,
+    exitCode: Int32 = ExitCode.success.rawValue,
```
miguelangel-dev: I think you should use directly `ExitCode` as a type for `AssertExecuteCommand` - to force you to keep tests with the same semantic and if it is the output for the execution, it will be always the same type. 

If you use `Int32` you will be less strict, and some bugs can be introduced. wdut?

```suggestion
    exitCode: ExitCode = .success,
```

natecook1000: 👍 I was resisting this for some reason. Thanks!

RLovelett:
Are there any guides on how one might use this API to give custom exit codes?

PR #91, opened 2020-03-19
Ensure that all parsable arguments have corresponding coding keys

YuAo:
This PR add a validator to ensure each flag, option, argument of a `ParsableCommand` has the corresponding coding key.

For the following command:
```
struct Repeat: ParsableCommand {
    @Option(help: "The number of times to repeat 'phrase'.")
    var count: Int?

    @Flag(help: "Include a counter with each repetition.")
    var includeCounter: Bool

    @Argument(help: "The phrase to repeat.")
    var phrase: String
    
    enum CodingKeys: String, CodingKey {
        case includeCounter
        case phrase
        case count
    }

    func run() throws { /* ... */ }
}
```

### Current behavior

- If the command author forget to add the `includeCounter` to the `CodingKeys`. The command failed with fatal error: 

    `Can't read a value from a parsable argument definition.`

- If the command author forget to add the `count` key and the user supplies a `--count` option. The command failed with message: 

    `Error: Unknown option '--count'. Did you mean '--count'?` 🤔

- If the command author forget to add the `phrase` key. The command failed with message: 
    
    `Error: Unexpected argument 'Hello'`

### Proposed behavior

None of the current behaviors seems appropriate. This PR proposed a validator to ensure all parsable arguments have corresponding coding keys. If the coding key of a parsable argument is missing the command parser reports fatal error: `Can't find the coding key for a parsable argument.`

### Detail

It looks like there's no way to do the validation at compile time based on the current parser design. Create a runtime check is not straight forward too. 

The `ArgumentSet` uses reflection to get the names of the arguments, however we can't get a list of CodingKeys so we can't see if the `CodingKeys` contains all of the arguments.

The solution proposed by this PR uses a special `Decoder` to get the type of `CodingKey` then try to create instances of `CodingKey` using given argument names. If `CodingKey` can be created from all the argument names we know that the command passes the validation.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -49,9 +49,16 @@ struct ParsableArgumentsCodingKeyValidator {
   
   /// This error indicates that an option, a flag, or an argument of
   /// a `ParsableArguments` is defined without a corresponding `CodingKey`.
-  struct Error: Swift.Error {
+  struct Error: Swift.Error, CustomStringConvertible {
     let parsableArgumentsType: ParsableArguments.Type
     let missingCodingKeys: [String]
+    var description: String {
+      if missingCodingKeys.count > 1 {
+        return "Arguments \(missingCodingKeys.map({ "`\($0)`" }).joined(separator: ",")) of `\(parsableArgumentsType)` are defined without corresponding `CodingKey`s."
+      } else {
+        return "Argument `\(missingCodingKeys[0])` of `\(parsableArgumentsType)` is defined without a corresponding `CodingKey`."
+      }
+    }
   }
   
   static func validate(_ type: ParsableArguments.Type) throws {
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -9,14 +9,21 @@
 //
 //===----------------------------------------------------------------------===//
 
+extension ParsableArguments {
+  static func _validate() throws {
+    try ParsableArgumentsCodingKeyValidator.validate(self)
+  }
+}
+
+/// Ensure that all arguments have corresponding coding keys
 struct ParsableArgumentsCodingKeyValidator {
   
   private struct Validator: Decoder {
     let argumentKeys: [String]
     
-    enum ValidationResult: Error {
+    enum ValidationResult: Swift.Error {
       case success
-      case codingKeyNotFound(String)
+      case missingCodingKeys([String])
     }
     
     let codingPath: [CodingKey] = []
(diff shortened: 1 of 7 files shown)


YuAo:
@natecook1000 what do you think?

natecook1000:
@YuAo This looks great! I'd like to pull it out into a separate validation step, since there are a variety of validations of the command declarations that we'll want to perform in debug mode but not in release. Could you move this into a separate method on `ParsableArguments`?

```swift
extension ParsableArguments {
    public static func _validate() throws { ... }
}
```

Also, the validation should throw on an error instead of trapping. Thanks!

YuAo:
@natecook1000 Thanks for the feedback!

I thought trapping would be better because this is an unrecoverable error and will cause unexpected behaviors. We don't want the program to continue when this happens.

>  we'll want to perform in debug mode but not in release.

If the validation is only performed in debug mode, then we are leaving the responsibility to the library users. It is not so important to trap, I agree that we should throw an error.

What do you think is the proper way to perform the validation only in debug mode? `#if DEBUG` or  `assert({ do { try type._validate(); return true } catch { return false }}())` or anything better?

I prefer

```
#if DEBUG
do { 
  try type._validate() 
} catch {
  assertionFailure("\(error)") 
}
#endif
```

---

>  public static func _validate() throws { ... }

Do we need making this public? There seems to be no need of running the check outside the library.

YuAo:
About other validations, did you mean this one: https://github.com/apple/swift-argument-parser/blob/9e77589d8532789635dc06e9ffce0280e2ca83d1/Sources/ArgumentParser/Parsing/ArgumentSet.swift#L44-L48

Then we can do something like this?
```
extension ParsableArguments {
  static func _validate() throws {
    // Use ArgumentSet(self).hasValidArguments or move the whole logic here.
    if !ArgumentSet(self).hasPositionalFollowingRepeated {
      throw Error.hasPositionalFollowingRepeated
    }
    try ParsableArgumentsCodingKeyValidator.validate(self)
  }
}
```

YuAo:
@natecook1000 Any thoughts?

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+  }
+  
+  /// This error indicates that an option, a flag, or an argument of
+  /// a `ParsableArguments` is defined without a corresponding `CodingKey`.
+  struct Error: Swift.Error {
+    let parsableArgumentsType: ParsableArguments.Type
+    let missingCodingKeys: [String]
+  }
```
natecook1000: Could you make this type conform to `CustomStringConvertible` and provide a clear explanation of what’s going wrong? Right now the fatal error doesn’t do much to help a library user fix the problem.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -0,0 +1,86 @@
...
+  
+  static func validate(_ type: ParsableArguments.Type) throws {
+    let argumentKeys: [String] = Mirror(reflecting: type.init())
```
natecook1000: This looks like a generally useful pattern that we’ll want for a variety of validations. What do you think about breaking this out as a `ParsableArguments` extension?

YuAo: I prefer making this a protocol. So we can have multiple validators, separate the logics, and we can write tests for each of them.

```Swift
protocol ParsableArgumentsValidator {
  static func validate(_ type: ParsableArguments.Type) throws
}

// for hasPositionalFollowingRepeated
struct ArgumentPositionValidator: ParsableArgumentsValidator {}

// for coding keys
struct ParsableArgumentsCodingKeyValidator: ParsableArgumentsValidator {}

extension ParsableArguments {
  static func _validate() throws {
    let validators: [ParsableArgumentsValidator.Type] = [
      ArgumentPositionValidator.self,
      ParsableArgumentsCodingKeyValidator.self
    ]
    for validator in validators {
      try validator.validate(self)
    }
  }
}
```

YuAo:
Thanks for the feedback @natecook1000 !

I left a comment about the design of the validation layer. And added my propose 
2316411.

I did not add any existing validations here, I think that would be a little off the topic and should be in a separate PR or something.

Let me know what do you think!

natecook1000:
LGTM, @YuAo! I'll merge this now; would be great to add some more validators, as well. I think we may want to tweak the behavior once we have multiple validations happening, since it would be nice to see _all_ the failed validations, rather than just the first one that doesn't pass.

PR #101, opened 2020-03-28
Add `PositionalArgumentsValidator`

YuAo:
### Abstract

Continue restructuring the `ParsableArguments` validation layer. (Previous discussions: #91)

### Detail

- Move the "positional argument following an array of positional arguments" check to a new struct `PositionalArgumentsValidator`.

- Make `ParsableArguments._validation` throw a combination of validation errors, and add the proper error description.

- Add tests for `PositionalArgumentsValidator`

So a `ParsableArguments` with both coding key error and positional arguments error is now reported in the following form:
```
Fatal error: Validation failed for `Repeat`:
- Can't have a positional argument `phrase` following an array of positional arguments `items`.
- Arguments `includeCounter`,`items` are defined without corresponding `CodingKey`s.
: file /Users/***/Developer/swift-argument-parser/Sources/ArgumentParser/Parsable Types/ParsableArguments.swift, line 196
```

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -13,13 +13,77 @@ fileprivate protocol ParsableArgumentsValidator {
   static func validate(_ type: ParsableArguments.Type) throws
 }
 
+struct ParsableArgumentsValidationError: Error, CustomStringConvertible {
+  let parsableArgumentsType: ParsableArguments.Type
+  let underlayingErrors: [Error]
+  var description: String {
+    """
+    Validation failed for `\(parsableArgumentsType)`:
+    \(underlayingErrors.map({"- \($0)"}).joined(separator: "\n"))
+    
+    """
+  }
+}
+
 extension ParsableArguments {
   static func _validate() throws {
     let validators: [ParsableArgumentsValidator.Type] = [
+      PositionalArgumentsValidator.self,
       ParsableArgumentsCodingKeyValidator.self
     ]
-    for validator in validators {
-      try validator.validate(self)
+    let errors: [Error] = validators.compactMap { validator in
+      do {
+        try validator.validate(self)
+        return nil
+      } catch {
+        return error
+      }
+    }
+    if errors.count > 0 {
+      throw ParsableArgumentsValidationError(parsableArgumentsType: self, underlayingErrors: errors)
+    }
+  }
+}
+
+/// For positional arguments to be valid, there must be at most one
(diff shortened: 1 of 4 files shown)


Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
```diff
@@ -106,11 +104,48 @@ final class CodingKeyValidationTests: XCTestCase {
...
+      if let error = error as? PositionalArgumentsValidator.Error {
+        XCTAssert(error.positionalArgumentFollowingRepeated == "phrase")
+        XCTAssert(error.repeatedPositionalArgument == "items")
+      } else {
+        XCTFail()
+      }
+      XCTAssert(error is PositionalArgumentsValidator.Error)
+    }
+    try PositionalArgumentsValidator.validate(H.self)
+  }
```
natecook1000: Could you add a test where the valid/invalid cases are reached by combining `@OptionGroup`s in different orders?

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -13,13 +13,99 @@ fileprivate protocol ParsableArgumentsValidator {
...
+    guard let repeatedPositional = sets.firstIndex(where: { $0.firstRepeatedPositionalArgument != nil })
+      else { return }
+    let positionalFollowingRepeated = sets[repeatedPositional...]
+      .dropFirst()
+      .first(where: { $0.firstPositionalArgument != nil })
+    
+    if let positionalFollowingRepeated = positionalFollowingRepeated {
+      let firstRepeatedPositionalArgument: ArgumentDefinition = sets[repeatedPositional].firstRepeatedPositionalArgument!
+      let positionalFollowingRepeatedArgument: ArgumentDefinition = positionalFollowingRepeated.firstPositionalArgument!
+      throw Error(repeatedPositionalArgument: firstRepeatedPositionalArgument.valueName,
```
YuAo: What do you think about using `valueName` here.
An argument `numberOfItems` will be reported as `number-of-items` using `valueName`. Should we use `help.keys.first?.rawValue`?

YuAo:
Tests updated! 
I'm a little bit unsure about the display of argument names. I'd like to hear your suggestion.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
natecook1000: Good call — the value name is user-facing; we want to use the actual property names declared in the type here if possible.

PR #102, opened 2020-03-28
Add built-in support for --version flag

### LINKED ISSUE #25: Is there a way to balance the processing of the argument with the "--version" option?
Is there a way to balance the processing of the argument with the "--version" option?

Here is my sample code. This command is simple echo, prints argument and prints version number if `--version` flag is applied.

```swift
import ArgumentParser

struct MyEcho: ParsableCommand {

    @Flag(name: .shortAndLong)
    var version: Bool

    @Argument()
    var text: String

    func run() {
        if version {
            print("v1.0.0")
            return
        }

        print(text)
    }
}

MyEcho.main()
```

However this sample code does not work well when you run `myEcho --version` because default validation by parser detects no argument. Like this

```sh
% myEcho --version
Error: Missing expected argument '<text>'
Usage: my-echo <text> [--version]
```

If I change `var text: String` property as optional, then `myEcho --version` works well but I have to handle a situation that no argument applied.

Here is my current solution bellow. Is there a better way to handle this situation?

```swift
struct MyEcho: ParsableCommand {

    @Flag(name: .shortAndLong)
    var version: Bool

    @Argument()
    var text: String?

    func run() throws {
        if version {
            print("ver 1.0.0")
            return
        }

        guard let text = text else {
            throw ValidationError("need argument")
        }

        print(text)
    }
}
```

natecook1000:
### Description
Adds built-in support for responding to a `--version` flag. This flag is only valid when a command provides a `version:` parameter in its configuration. Fixes #25.

### Detailed Design
If a command in the tree has specified a version, the `--version` flag is checked for only after evaluating any explicitly defined arguments. This allows commands to define their own `--version` flags or options and not have them be overridden by the auto-generated version.

Nested commands inherit their version from their parent commands, and can override their parent by providing their own version string.

`version` is added to the `CommandConfiguration` type and to the initializer:

```swift
  /// Creates the configuration for a command.
  ///
  /// - Parameters:
  ///   - commandName: The name of the command to use on the command line. If
  ///     `commandName` is `nil`, the command name is derived by converting
  ///     the name of the command type to hyphen-separated lowercase words.
  ///   - abstract: A one-line description of the command.
  ///   - discussion: A longer description of the command.
  ///   - version: The version number for this command. When you provide a
  ///     non-empty string, the arguemnt parser prints it if the user provides
  ///     a `--version` flag.
  ///   - shouldDisplay: A Boolean value indicating whether the command
  ///     should be shown in the extended help display.
  ///   - subcommands: An array of the types that define subcommands for the
  ///     command.
  ///   - defaultSubcommand: The default command type to run if no subcommand
  ///     is given.
  ///   - helpNames: The flag names to use for requesting help, simulating
  ///     a Boolean property named `help`.
  public init(
    commandName: String? = nil,
    abstract: String = "",
    discussion: String = "",
    version: String = "",
    shouldDisplay: Bool = true,
    subcommands: [ParsableCommand.Type] = [],
    defaultSubcommand: ParsableCommand.Type? = nil,
    helpNames: NameSpecification = [.short, .long]
  )
```

### Documentation Plan
How has the new feature been documented? Have the relevant portions of the guide been updated in addition to symbol-level documentation?

### Test Plan
- Unit tests for generating the version string.
- End-to-end tests to verify that the version is being rendered as the correct value when defined, and that commands that explicitly define `--version` get that behavior.
- Example tests with versions defined in `Math`.

### Source Impact
Because the `version:` parameter is defaulted, this has no source impact. In addition, since there's no change in behavior unless the command defines a version, there's no impact on existing tools' behavior.

### Checklist

### DIFF
--- a/Examples/math/main.swift
+++ b/Examples/math/main.swift
@@ -18,6 +18,9 @@ struct Math: ParsableCommand {
         // Optional abstracts and discussions are used for help output.
         abstract: "A utility for performing maths.",
 
+        // Commands can define a version for automatic '--version' support.
+        version: "1.0.0",
+
         // Pass an array to `subcommands` to set up a nested tree of subcommands.
         // With language support for type-level introspection, this could be
         // provided by automatically finding nested `ParsableCommand` types.
@@ -89,7 +92,8 @@ extension Math {
 extension Math.Statistics {
     struct Average: ParsableCommand {
         static var configuration = CommandConfiguration(
-            abstract: "Print the average of the values.")
+            abstract: "Print the average of the values.",
+            version: "1.5.0-alpha")
         
         enum Kind: String, ExpressibleByArgument {
             case mean, median, mode
--- a/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
+++ b/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
@@ -24,6 +24,9 @@ public struct CommandConfiguration {
   /// display.
   public var discussion: String
   
+  /// Version information for this command.
+  public var version: String
+
   /// A Boolean value indicating whether this command should be shown in
   /// the extended help display.
   public var shouldDisplay: Bool
@@ -57,6 +60,7 @@ public struct CommandConfiguration {
     commandName: String? = nil,
     abstract: String = "",
     discussion: String = "",
+    version: String = "",
     shouldDisplay: Bool = true,
     subcommands: [ParsableCommand.Type] = [],
     defaultSubcommand: ParsableCommand.Type? = nil,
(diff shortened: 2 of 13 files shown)


natecook1000:
@griffin-stewie What do you think of this design?

griffin-stewie:
@natecook1000  Thank you for adding built-in support for version flag. It looks great.

One thing I noticed is that the information about the version flag is not printed in the `OPTIONS` section of the help. I felt it would have been nice to have a help message output to make it more end-user friendly.

You don't support short names like `-v`, do you? It might be nice to be able to customize it like `var helpNames: NameSpecification` which is already provided.

natecook1000:
Great points, @griffin-stewie! I've added the flag to the generated help, but I'm holding off for now on customizing the flag. There might be a more general pattern for customizing these built-in options, which I'd like to look at after getting code-completion support built.

griffin-stewie:
I agree with you. Thank you for implementing it.

dduan:
This is awesome. Thanks!

Issue #103, opened 2020-03-30
`validate()` is not called on parent commands even though their arguments are required


kennyyork:
When a subcommand is parsed and executed, its parent command arguments are parsed but `validate()` is never called.  It seems odd that the arguments are required for the command to be accepted, but are not validated.  

**ArgumentParser version:** `master`
**Swift version:** Apple Swift version 5.2

### Checklist
- [x] If possible, I've reproduced the issue using the `master` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
```
private struct BaseCommand: ParsableCommand {
  static var configuration = CommandConfiguration(
    commandName: "base",
    subcommands: [SubCommand.self]
  )

  @Option()
  var baseFlag: String

  mutating func validate() throws {
    throw HelpRequested
  }
}

extension BaseCommand {
  struct SubCommand : ParsableCommand {
    static var configuration = CommandConfiguration(commandName: "sub")

    @Flag()
    var subFlag: Bool

    mutating func validate() throws {
      print("Validated SubCommand")
    }
  }
}
```

### Expected behavior
Because `--base-flag` is required for the command to run, I would expect at the least that `BaseCommand.validate()` be called. 

### Actual behavior
`BaseCommand.validate()` is never called when invoking SubCommand

natecook1000:
Fixed by #104! Thanks @kennyyork!

PR #104, opened 2020-03-31
Call validate on the entire command stack

kennyyork:
This change calls validate on every ParsableCommand in the command stack, instead of just the leaf command to fix issue #103 

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -106,7 +106,7 @@ extension CommandParser {
   
   /// Extracts the current command from `split`, throwing if decoding isn't
   /// possible.
-  fileprivate mutating func parseCurrent(_ split: inout SplitArguments) throws {
+  fileprivate mutating func parseCurrent(_ split: inout SplitArguments) throws -> ParsableCommand {
     // Build the argument set (i.e. information on how to parse):
     let commandArguments = ArgumentSet(currentNode.element)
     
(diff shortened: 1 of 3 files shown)


natecook1000:
Thanks, @kennyyork! I agree that this is definitely worth fixing.

I think there’s the potential for a confusing order of validation here, depending on how things are declared. In particular, if a “super” command is declared as an `@OptionGroup` property on a subcommand, it will get validated during the process of that subcommand being decoded, while a command that isn’t later used as a group will only get validated during the final stack validation. 

Perhaps we should validate each command immediately after it’s decoded, en route to parsing the whole command stack? I think that would provide more understandable behavior, since it would always have a “top-down” validation order.

kennyyork:
I did that in my initial implementation, but I wasn't sure which approach was best.  You either:

1) parse all the command line arguments to verify valid inputs, then validate objects
2) parse and then validate each command

For 1, you expose missing/extra arguments first
For 2, you expose invalid arguments or combinations first

I could see the argument (puns!) that input arguments are read from left to right, so it does make sense to validate each command after it is parsed.  This would also line up with how the OptionGroups are validated.

kennyyork:
Moved the validation call to happen after decoding each command

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -145,14 +145,23 @@ extension CommandParser {
...
     while true {
-      try parseCurrent(&split)
-      
+      var parsedCommand = try parseCurrent(&split)
+
+      // after decoding a command, make sure to validate it
+      do {
+        try parsedCommand.validate()
+      } catch {
+        throw CommandError(commandStack: [currentNode.element], parserError: ParserError.userValidationError(error))
```
natecook1000, line 162: This needs to pass the full `commandStack` instead of just the current element, so that the error can have the right context when printing the usage information.

PR #108, opened 2020-04-03
Modified convertedToSnakeCase to match the Xcode behaviour

fblondiau:
There is no unanimous agreement on the _right_ way to treat numbers while converting to snake cases.

`”PDF2Text".convertedToSnakeCase` could have been giving `pdf2text`, `pdf2_text` or `pdf_2_text`.

Today, the second option is used but, arguably, according to the development tools on the Mac platform, the last option seems to be more natural.

Indeed, when increasing, sub-word by sub-word, the selection of this camelCased formatted identifier (1), Xcode will detect 3 parts : `PDF`, `2` and `Text` — hence this modification proposal.

(1) check by placing the caret at the beginning of the identifier and, holding down the alternate and shift keys simultaneously, press the right arrow key.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Utilities/StringExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/StringExtensions.swift
@@ -97,20 +97,25 @@ extension String {
     var result = ""
     // Whether we should append a separator when we see a uppercase character.
     var separateOnUppercase = true
+    var justGotLastDigit = false
     for index in indices {
       let nextIndex = self.index(after: index)
       let character = self[index]
-      if character.isUppercase {
-        if separateOnUppercase && !result.isEmpty {
+      if justGotLastDigit && character != separator {
+        result += "\(separator)"
+      }
+      if character.isUppercase || character.isNumber {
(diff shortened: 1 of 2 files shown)


natecook1000:
I agree that it's hard to say which outcome is correct, primarily because we don't know whether the number is an independent entity, as in PDF2Text, or if it belongs to the preceding or following word. Given that, and based on some [highly scientific polling data](https://twitter.com/nnnnnnnn/status/1247969925634953216), I agree that the default should be to switch to your second option.

That said, this would be a breaking change (not source-breaking, but interface-breaking) for existing users of the library, so I'm going to wait a bit before integrating this.

natecook1000:
@fblondiau On further consideration, I've decided to not take this change at this time. While I can see the argument in your initial post, there are plenty of counter-examples that are made worse by breaking before numbers. In addition, the current behavior matches the behavior of `JSONDecoder`/`Encoder`, so it is the most consistent behavior for users, if that claim can be made of any default behavior.

Thanks again for opening this PR!

PR #109, opened 2020-04-05
Parse from environment

danieleggert:
### Description

This allows reading values for `@Option` and `Flag` from environment variables.

I started work on this before this repo was open sourced, and it was requested as #4

### Detailed Design

Flags can now be specified as
```swift
@Flag(name: [.environment, .long])
```
and similarly options as
```swift
@Option(name: [.environment, .long])
```

There are two new `NameSpecification` to support this:
```swift
    /// Read the value from an environment variable. The name is using the `THE_QUICK_BROWN_FOX` style.
    case environment
    
    /// Read the value from an environment variable with the specified name/
    case customEnvironment(String)
```

As one would expect, `.environment` picks a auto-genreated name based on on the flag / option name, similar to how for command line names are generated. But for environment variables, the generated strings are using the `THE_QUICK_BROWN_FOX` style.

TODO: I’d like to prefix the executable name to the auto-generated environment variable names. Need to funnel that through.

#### Example

If we have
```swift
enum Shape: String, CaseIterable {
  case round
  case square
  case oblong
}

  @Flag(name: [.environment, .long])
  var shape: Shape?
}
```
we can set an environment variable `ROUND` to make `shape = .round` — similar to the command line argument `--round`.

For options if we have
```swift
  @Option(name: [.environment, .long]) var name: String?
}
```
we can set the environment variable `NAME` to `Daniel`, and `name = "Daniel` — similar to the command line arguments `--name Daniel`.

#### Details

There’s a new `EnvironmentName` type that’s a `RawRepresentable` wrapping a `String` to give some type-safety. Environment variables get mapped to a `[EnvironmentName: String]`.

For flags and options that are exclusive, we still allow the command line to override any environment variables.

There’s a new `OriginalInput` type that’s using in a few places where we used to have a `[String]`. This would be the array of arguments passed to the executable. Now it’s type safe and holds both this array and the environment variables that are set. This cleans up the code in a few places.

The `InputOrigin` now can both represent an index into the arguments array or an environment variable name.

### Documentation Plan

TODO:
- [ ] update Documentation
- [ ] update the guide

### Test Plan

Updated the unit tests as applicable.

### Source Impact

Purely additional. Should not require any source changes by users.

### Checklist
- [ ] Added code to prefix the executable name to environment variable names

### DIFF
--- a/Sources/ArgumentParser/Parsing/ArgumentSet.swift
+++ b/Sources/ArgumentParser/Parsing/ArgumentSet.swift
@@ -96,7 +96,7 @@ extension ArgumentSet {
   /// Creates an argument set for a single Boolean flag.
   static func flag(key: InputKey, name: NameSpecification, help: ArgumentHelp?) -> ArgumentSet {
     let help = ArgumentDefinition.Help(options: .isOptional, help: help, key: key)
-    let arg = ArgumentDefinition(kind: .name(key: key, specification: name), help: help, update: .nullary({ (origin, name, values) in
+    let arg = ArgumentDefinition(kind: .name(key: key, specification: name), environmentNames: name.makeEnvironmentNames(key), help: help, update: .nullary({ (origin, name, values) in
       values.set(true, forKey: key, inputOrigin: origin)
     }), initial: { origin, values in
       values.set(false, forKey: key, inputOrigin: origin)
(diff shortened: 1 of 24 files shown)


Sources/ArgumentParser/Parsing/ArgumentSet.swift
```diff
@@ -96,31 +96,20 @@ extension ArgumentSet {
...
-          values.set(value, forKey: key, inputOrigin: origin)
-        }
-        else {
-          throw ParserError.duplicateExclusiveValues(previous: previous.inputOrigin, duplicate: origin, originalInput: values.originalInput)
-        }
-      }
-    case (true, .chooseFirst):
-      values.update(forKey: key, inputOrigin: origin, initial: value, closure: { _ in })
-    case (false, _), (_, .chooseLast):
+  static func updateFlag<Value: Equatable>(key: InputKey, value: Value, origin: InputOrigin, values: inout ParsedValues, didUpdate: Bool, exclusivity: FlagExclusivity) throws -> Bool {
```
glessard, line 107: I wonder whether it would be better to decouple `updateFlag` from the `exclusivity.shouldAllowUpdate` call, and just have it take the `ExclusivityResult` enum as an argument.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
```diff
@@ -162,12 +151,49 @@ extension ArgumentSet {
...
+extension FlagExclusivity {
+  /// Should we allow updating the value?
+  ///
+  /// Will return `.allow` if the value should be updated. Throws an error if updating is invalid. Returns `.ignore` if the value should be ignored.
+  func shouldAllowUpdate<Value: Equatable>(key: InputKey, origin: InputOrigin, value: Value, values: ParsedValues, didUpdate: Bool) throws -> ExclusivityResult {
+    let previous = values.element(forKey: key)
+    switch (didUpdate, previous?.inputOrigin, self) {
+    case (_, let o?, _) where !o.containsAnyArguments:
+      // If the value was set from environment variables only,
+      // we don’t care able the exclusivity.
```
glessard, line 168: typo: `// we don't care *about* the exclusivity.`

natecook1000:
@danieleggert Awesome to see this coming along! 👏👏 A few questions about the behavior:

- Are options or flags allowed to have _only_ `.environment` as their name, such that there's no way to set them on the command line?
- Does it make sense to think about allowing positional arguments to be specified in the environment?
- How do you anticipate making the environment keys visible to users?

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

sergiocampama:
@danieleggert will you take this on to re-add it to the queue? or do you need some help pushing this?

danieleggert:
@natecook1000 What’s your thought on this? I know you wanted to look at this, too? I should be able to get a few spare cycles to update this over the coming weeks… 🤞

PR #110, opened 2020-04-07
Print custom errors thrown from a `transform` closure

rjstelling:
> This PR is probably incomplete, I just wanted some feedback that I was heading in the right direction before I spent any more time on it.

## Rational

If an argument property supplies a `transform` closure, like so,

```swift
@Option(help: "Some string input please.", transform: { try convert($0) })
var input: String
```

...and if the function `convert(...)` throws, this error should be printed to standard error, the same way throwing from `run()` or `validate()` does.

I raised this issue on the [Swift Forums](https://forums.swift.org/t/throwing-from-a-transform-closure/35137) and it was suggested by @natecook1000 this was a bug.

## Changes

1. I added a file called TransformEndToEndTests.swift that test simple cases of using `transform` closures with `@Option` properties. As far as I could tell there were no tests for `transform` closures.
2. I made a change to `parse(arguments: ...)` to catch any `ValidationError` and wrap it in a `ParserError.userValidationError`. The `ValidationError` is then printed as expected.
3. I updated the fail test to check for the `ValidationError` message.

## Before merging

1. Add tests (and implementation?) for `@Argument`
2. Discuss what errors can be thrown from a `transform` closure. Just `ValidationError` or should it support `CleanExit`, `ExitCode`?
3. Update documentation 

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -206,6 +206,8 @@ extension CommandParser {
       return .failure(error)
     } catch let error as ParserError {
       return .failure(CommandError(commandStack: commandStack, parserError: error))
+    } catch let error as ValidationError {
+      return .failure(CommandError(commandStack: commandStack, parserError: .userValidationError(error)))
     } catch is HelpRequested {
       return .success(HelpCommand(commandStack: commandStack))
     } catch {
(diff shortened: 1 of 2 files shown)


natecook1000:
@rjstelling Thanks for this PR! We actually want to catch these errors a bit earlier, at the place where the `transform` closure is called, so that we can properly respond to _any_ kind of error that's thrown during parsing. If errors thrown from that point are caught and converted into `ParserError.unableToParseValue`, I think we'll be on the right path.

Note that we don't have any custom messaging with that error, so you may want to look at adding another associated value.

PR #111, opened 2020-04-08
Changed Readme notes about the ArgumentParser integration via SwiftPM

sgl0v:
The ArgumentParser dependency should be specified using the `target` block, not `product`.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,5 +102,17 @@ use this dependency specification instead:
 Finally, include `"ArgumentParser"` as a dependency for your executable target:
 
 ```swift
-.product(name: "ArgumentParser", package: "swift-argument-parser"),
+let package = Package(
+    // name, platforms, products, etc.
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-argument-parser", from: "0.0.1"),
+        // other dependencies
+    ],
+    targets: [
+        .target(name: "<command-line-tool>", dependencies: [
+            .product(name: "ArgumentParser", package: "swift-argument-parser"),
+        ]),
+        // other targets
+    ]
+)
 ```


natecook1000:
Hi @sgl0v! Could you show an example of a `Package.swift` file with this syntax? I'm not seeing what this fix is meant to address.

sgl0v:
Hi @natecook1000 ! Here is the example of a `Package.swift`:

```swift
// swift-tools-version:5.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "TestProject",
    platforms: [
        .macOS(.v10_11)
    ],
    products: [
        .executable(name: "TestApp", targets: ["TestTarget"]),
    ],
    dependencies: [
        .package(
            url: "https://github.com/apple/swift-argument-parser",
            from: "0.0.1"
        )
    ],
    targets: [
        .target(name: "TestTarget", dependencies: ["ArgumentParser"])
    ]
)
```

I've defined a `TestApp` executable product, that is built with the `TestTarget`.  The one has `ArgumentParser` as a dependency.

It might be that I've missed something, but looks like `.product(name: "ArgumentParser", package: "swift-argument-parser"),` is not the way you can specify a target's dependency. You have to use `.target(name: "TestTarget", dependencies: ["ArgumentParser"])` instead.

natecook1000:
Ah, I see! The package manifest format has a [change in version 5.2](https://github.com/apple/swift-evolution/blob/master/proposals/0226-package-manager-target-based-dep-resolution.md) that only allows specifying the target dependency with a string (e.g. `"ArgumentParser"`) when the product has the same name as the package it's exported from. In our case, that's not true — the package is called `swift-argument-parser`, but the product is `ArgumentParser` — so we need to use the longer `.product(...)` syntax.

Here's a larger example (this works with swift-tools-version 5.1 or 5.2):

```swift
let package = Package(
    // name, platforms, products, etc.
    dependencies: [
        .package(url: "https://github.com/apple/swift-argument-parser", from: "0.0.1"),
        // other dependencies
    ],
    targets: [
        .target(name: "<command-line-tool>", dependencies: [
            .product(name: "ArgumentParser", package: "swift-argument-parser"),
        ]),
        // other targets
    ]
)
```

Would you like to add this to the README section instead of just showing the individual lines? That might make it more clear.

sgl0v:
Thanks for the explanation. Yes, it makes it much more clear. I've added the code snippet to the README section.

natecook1000:
@swift-ci Please test

PR #115, opened 2020-04-14
Improve errors messages thrown from a `transform` closure

rjstelling:
> This PR is an improved version after some feedback on #110. However, I wanted to to make sure this is still the right direction, and I'm happy for more feedback, especially because this PR is more substantial. 

## Rational

If an argument property supplies a `transform` closure, like so,

```swift
@Option(help: "Some string input please.", transform: { try convert($0) })
var input: String
```

...and if the function `convert(...)` throws, this error should be printed to standard error, the same way throwing from `run()` or `validate()` does.

I raised this issue on the [Swift Forums](https://forums.swift.org/t/throwing-from-a-transform-closure/35137) and it was suggested by @natecook1000 this was a bug.

### tl;dr

Improves default error messages when `transform` closures throw errors. Adds the ability to print custom error messages when `transform` closures throw. 

### Detail
1. Added a file called TransformEndToEndTests.swift that test simple cases of using `transform` closures with `@Option`and `@Argument` properties.
2. Added error message tests for custom and default errors.
3. In the `init(...)` calls of `Option` and `Argument` any errors thrown by the `transform` closure care caught and converted into an `ParserError.unableToParseValue`
4. Added a `customMessage` associated value to `ParserError.unableToParseValue`, this is a `String` optional. When this is set the default error message is suppressed and `customMessage` is used.
5. Added a public protocol called `CustomParserErrorConvertible`. Errors can conform to this to provide there own custom messages. Added conformance to `ValidationError`.

## Before merging

1. Discuss the usage of `CustomParserErrorConvertible`. The only other alternative I considered was using `localizedDescription`. But this would require another mechanism to use the default error message.
3. Update documentation 

### Checklist

### DIFF
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -178,7 +178,7 @@ struct Example: ParsableCommand {
 }
 ```
 
-Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. (See: [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) in [Validation and Errors](./05%20Validation%20and%20Errors.md) for details about customising `transform` function errors.)
+Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. See [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) for more about customizing `transform` function errors.
 
 ## Using flag inversions, enumerations, and counts
 
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -178,7 +178,7 @@ struct Example: ParsableCommand {
 }
 ```
 
-Throw an error from the `transform` function to indicate that the user provided an invalid value for that type.
+Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. (See: [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) in [Validation and Errors](./05%20Validation%20and%20Errors.md) for details about customising `transform` function errors.)
 
 ## Using flag inversions, enumerations, and counts
 
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -122,7 +122,7 @@ struct Example: ParsableCommand {
 
   // Reads in the argument string and attempts to transform it to
   // a `ExampleDataModel` object using the `JSONDecoder`. If the
-  // string is not valid JSON `decode` will throw an error and
+  // string is not valid JSON, `decode` will throw an error and
   // parsing will halt.
   @Argument(transform: {
     guard let data = $0.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -130,7 +130,7 @@ struct Example: ParsableCommand {
   var inputJSON: ExampleDataModel
   
   // Specifiying this option will always cause the parser to exit
-  // and print the cistom error.
+  // and print the custom error.
   @Option(transform: { throw ExampleTransformError(description: "Trying to write to failOption always produces an error. Input: \($0)") })
   var failOption: String?
 }
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -138,3 +138,19 @@ struct Example: ParsableCommand {
   var failOption: String?
 }
 ```
+
+Throwing from a transform closure not only befits users by providing more context but can reduce development time by pinpointing issues faster and mor precisely.
+
+```
+% example '{"Bad JSON"}'
+Error: The value '{"Bad JSON"}' is invalid for '<input-json>': dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "No value for key in object around character 11." UserInfo={NSDebugDescription=No value for key in object around character 11.})))
+Usage: example <input-json> --fail-option <fail-option>
+```
+
+While throwing Standard Library or Foundation errors adds context, custom errors provide the best experience for users and developers.
+
+```
+% example '{"tokenCount":0,"tokens":[],"identifier":"F77D661C-C5B7-448E-9344-267B284F66AD"}' --fail-option="Some Text Here!"
+Error: The value 'Some Text Here!' is invalid for '--fail-option <fail-option>': Trying to write to failOption always produces an error. Input: Some Text Here!
+Usage: example <input-json> --fail-option <fail-option>
+```
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
     }
 }
 ```
+
+## Handling Transform Errors
+
+During argument and option parsing you can use a closure to transform the command line stings to custom types. If this transformation fails you can throw a `ValidationError` its `message` property will be displayed to the user. 
+
+In addition, you can throw your own errors. Errors that conform to `CustomStringConvertible` or `LocalizedError` provide the best experience for users.
+
+```swift
+struct ExampleTransformError: Error, CustomStringConvertible {
+    var description: String
+}
+
+struct ExampleDataModel: Codable {
+  let identifier: UUID
+  let tokens: [String]
+  let tokenCount: Int
+}
+
+struct Example: ParsableCommand {
+
+  // Reads in the argument string and attempts to transform it to
+  // a `ExampleDataModel` object using the JSONDecoder. If the
+  // string is not valid JSON `decode` will throw an error and
+  // parsing will halt.
+  @Argument(transform: {
+    guard let data = $0.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
+    return try JSONDecoder().decode(ExampleDataModel.self, from: data) })
+  var inputJSON: ExampleDataModel
+  
+  // Specifiying this option will always cause the parser to exit
+  // and print the cistom error.
+  @Option(transform: { throw ExampleTransformError(description: "Trying to write to failOption always produces an error. Input: \($0)") })
+  var failOption: String?
+}
+```
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -109,24 +109,27 @@ In addition, you can throw your own errors. Errors that conform to `CustomString
 
 ```swift
 struct ExampleTransformError: Error, CustomStringConvertible {
-    var description: String
+  var description: String
 }
 
 struct ExampleDataModel: Codable {
   let identifier: UUID
   let tokens: [String]
   let tokenCount: Int
+  
+  static func dataModel(_ jsonString: String) throws -> ExampleDataModel  {
+    guard let data = jsonString.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
+    return try JSONDecoder().decode(ExampleDataModel.self, from: data)
+  }
 }
 
 struct Example: ParsableCommand {
 
   // Reads in the argument string and attempts to transform it to
-  // a `ExampleDataModel` object using the `JSONDecoder`. If the
+  // an `ExampleDataModel` object using the `JSONDecoder`. If the
   // string is not valid JSON, `decode` will throw an error and
   // parsing will halt.
-  @Argument(transform: {
-    guard let data = $0.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
-    return try JSONDecoder().decode(ExampleDataModel.self, from: data) })
+  @Argument(transform: ExampleDataModel.dataModel )
   var inputJSON: ExampleDataModel
   
   // Specifiying this option will always cause the parser to exit
--- a/Documentation/05 Validation and Errors.md
+++ b/Documentation/05 Validation and Errors.md
@@ -103,7 +103,7 @@ struct Example: ParsableCommand {
 
 ## Handling Transform Errors
 
-During argument and option parsing you can use a closure to transform the command line stings to custom types. If this transformation fails you can throw a `ValidationError` its `message` property will be displayed to the user. 
+During argument and option parsing, you can use a closure to transform the command line strings to custom types. If this transformation fails, you can throw a `ValidationError`; its `message` property will be displayed to the user. 
 
 In addition, you can throw your own errors. Errors that conform to `CustomStringConvertible` or `LocalizedError` provide the best experience for users.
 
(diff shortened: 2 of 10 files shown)


Sources/ArgumentParser/Parsing/ParserError.swift
```diff
@@ -25,7 +25,7 @@ enum ParserError: Error {
   case duplicateExclusiveValues(previous: InputOrigin, duplicate: InputOrigin, originalInput: [String])
   /// We need a value for the given key, but it’s not there. Some non-optional option or argument is missing.
   case noValue(forKey: InputKey)
-  case unableToParseValue(InputOrigin, Name?, String, forKey: InputKey)
+  case unableToParseValue(InputOrigin, Name?, String, forKey: InputKey, customMessage: String?)
```
jayrhynas, line 28: `customMessage` could have a default `nil` value so that you don't need to specify it every time

rjstelling: Adding a default value to the `customMessage` parameter also reduced the file footprint of this PR.

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,16 +339,18 @@ extension ErrorMessageGenerator {
...
-    case let (_, v?):
+    case let (_, v?, _):
       return "The value '\(value)' is invalid for '<\(v)>'"
-    case let (n?, _):
+    case let (n?, _, _):
       return "The value '\(value)' is invalid for '\(n.synopsisString)'"
-    case (nil, nil):
+    case (nil, nil, _):
       return "The value '\(value)' is invalid."
     }
```
jayrhynas, line 355: I suppose it's a matter of preference, but I feel like we should just keep the context, and append the custom message if it is not `nil`. This would also remove the need for `CustomParserErrorConvertible` and instead just use `localizedDescription`.

rjstelling: My preference is not to append custom messages to any default error or context. It's not really a custom message at that point, if I'm customising the error I want full control.

But I agree the context is useful. And using something like `localizedDescription` would make it cleaner. 

Maybe we should consider wrapping the original error (thrown from the transform closure) in the `unableToParseValue` instead of a custom error message. 

But then we'd need some error formatting option, possibly in `CommandConfiguration`?

natecook1000: I agree with @jayrhynas here — the printed error for this should combine the existing error output with a string from the thrown error. The output for a transformation throws on non-alphanumeric input could look like this:

```
$ command --option bad-input-#!@#
Error: The value 'bad-input' is invalid for '--option': Contains invalid characters '#!@#'
Usage: command [--option <option>] ...
``` 

where `Contains invalid characters '#!@#'` is the string representation of the thrown error.

> Maybe we should consider wrapping the original error (thrown from the transform closure) in the unableToParseValue instead of a custom error message.

This sounds good! We don't want to add a protocol just for this functionality, so doing a best effort conversion of the error to a string would be great. In general we try to do that by trying to match on `LocalizedError where error.errorDescription != nil` and then just falling back to `String(describing: error)`. That makes it simple for people to have nice error messages by just conforming their error type to `CustomStringConvertible`.

rjstelling: I have updated `unableToParseValue` to take an optional `Error` associated value. 

If this error is not nil `unableToParseValueMessage(…)` makes best-efforts to create a custom error message.

> The output for a transformation throws on non-alphanumeric input could look like this:
> 
> ```
> $ command --option bad-input-#!@#
> Error: The value 'bad-input' is invalid for '--option': Contains invalid characters '#!@#'
> Usage: command [--option <option>] ...
> ```
> 
> where Contains invalid characters '#!@#' is the string representation of the thrown error.

This is exactly how the errors are now formatted. "Contains invalid characters '#!@#'" must be returned by either `errorDescription` form a `LocalizedError` or `description` from `CustomStringConvertible`. If the error is `nil` the errors are formatted as before.

`ValidationError` works as expected, printing the `message` string.

Simple `enum`s, such as:

```swift
enum FooError: Error {
    case outOfBounds
    case unknown 
}
```

Simply print out the name of the `case`.

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,16 +339,36 @@ extension ErrorMessageGenerator {
...
+            case is LocalizedError where (error as? LocalizedError)?.errorDescription != nil:
+                return (error as? LocalizedError)?.errorDescription
+            case let err?:
+                return String(describing: err)
+            default:
+                return nil
+        }
+    }()
+    
+    switch (name, valueName, customErrorMessage) {
```
natecook1000, line 360: Since we should be doing the same thing in each case, this switch statement would be easier to understand if we just concatenate the `"The value ..."` text with the custom error message (if it exists).

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -239,8 +239,16 @@ extension Option {
...
       var arg = ArgumentDefinition(kind: kind, help: help, parsingStrategy: ArgumentDefinition.ParsingStrategy(parsingStrategy), update: .unary({
-        (origin, _, valueString, parsedValues) in
-        parsedValues.set(try transform(valueString), forKey: key, inputOrigin: origin)
+        (origin, name, valueString, parsedValues) in
+        
+        do {
+            let transformedValue = try transform(valueString)
+            parsedValues.set(transformedValue, forKey: key, inputOrigin: origin)
+        }
+        catch /* All other errors */ {
```
natecook1000, line 248: Nit: don't start a new line before `catch`, here and elsewhere.

```suggestion
        } catch {
```

Sources/ArgumentParser/Parsable Properties/Errors.swift
```diff
@@ -20,7 +20,7 @@ import MSVCRT
 /// An error type that is presented to the user as an error with parsing their
 /// command-line input.
 public struct ValidationError: Error, CustomStringConvertible {
-  var message: String
+  public internal(set) var message: String
```
natecook1000, line 23: This is a good idea, but since it's an API change, let's handle it in a separate PR.

rjstelling: This was a hangover from the previous protocol based design. I will open a separate PR.

Sources/ArgumentParser/Usage/UsageGenerator.swift
rjstelling: Trying to be a bit too clever and adding complexity in the process. 

I think some of these cases will never have custom messages, but it will just work if they ever do start passing in errors.

rjstelling:
I have also updated the "Validation and Errors" file with a section on "Handling Transform Errors”.

Its not my greatest work 😬

Hopefully by the time anyone see it I will have had chance to have another draft or 2.

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -343,33 +343,29 @@ extension ErrorMessageGenerator {
...
-                return (error as? LocalizedError)?.errorDescription
-            case let err?:
-                return String(describing: err)
-            default:
-                return nil
-        }
+    let customErrorMessage: String = {
+      switch error {
+      case is LocalizedError where (error as? LocalizedError)?.errorDescription != nil:
+        return ": " + (error as! LocalizedError).errorDescription! // !!! Checked above that this will not be nil
```
jayrhynas, line 352: This could be written as:
```
case let err as LocalizedError where err.errorDescription != nil:
    return ": " + err.errorDescription
```

rjstelling: I'm not sure why I have such a blind spot for switch/case patterns!

Sources/ArgumentParser/Parsable Properties/Errors.swift
rjstelling: @natecook1000 This change is no in a separate PR #119

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
     }
 }
 ```
+
+## Handling Transform Errors
+
+During argument and option parsing you can use a closure to transform the command line stings to custom types. If this transformation fails you can throw a `ValidationError` its `message` property will be displayed to the user.
```
xwu, line 106:
```suggestion
During argument and option parsing, you can use a closure to transform the command line strings to custom types. If this transformation fails, you can throw a `ValidationError`; its `message` property will be displayed to the user. 
```

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
...
+  // a `ExampleDataModel` object using the JSONDecoder. If the
+  // string is not valid JSON `decode` will throw an error and
+  // parsing will halt.
+  @Argument(transform: {
+    guard let data = $0.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
+    return try JSONDecoder().decode(ExampleDataModel.self, from: data) })
+  var inputJSON: ExampleDataModel
+  
+  // Specifiying this option will always cause the parser to exit
+  // and print the cistom error.
```
xwu, line 133:
```suggestion
  // and print the custom error.
```

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
...
+struct ExampleDataModel: Codable {
+  let identifier: UUID
+  let tokens: [String]
+  let tokenCount: Int
+}
+
+struct Example: ParsableCommand {
+
+  // Reads in the argument string and attempts to transform it to
+  // a `ExampleDataModel` object using the JSONDecoder. If the
```
xwu, line 124:
```suggestion
  // a `ExampleDataModel` object using the `JSONDecoder`. If the
```

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
...
+  // string is not valid JSON `decode` will throw an error and
```
xwu, line 125:
```suggestion
  // string is not valid JSON, `decode` will throw an error and
```

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,17 +339,33 @@ extension ErrorMessageGenerator {
     }
   }
   
-  func unableToParseValueMessage(origin: InputOrigin, name: Name?, value: String, key: InputKey) -> String {
+  func unableToParseValueMessage(origin: InputOrigin, name: Name?, value: String, key: InputKey, error: Error?) -> String {
     let valueName = arguments(for: key).first?.valueName
+    
+    // We want to make the "best effort" in producing a custom error message.
+    // We favour `LocalizedError.errorDescription` and fallback to
```
xwu, line 346:
```suggestion
    // We favour `LocalizedError.errorDescription` and fall back to
```

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,17 +339,33 @@ extension ErrorMessageGenerator {
...
+    // CustomStringConvertible. To opt-in return your custom error mesage
```
xwu, line 347:
```suggestion
    // `CustomStringConvertible`. To opt in, return your custom error message
```

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,17 +339,33 @@ extension ErrorMessageGenerator {
...
+    // in the `description` property of CustomStringConvertible.
```
xwu, line 348:
```suggestion
    // as the `description` property of `CustomStringConvertible`.
```

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -339,17 +339,33 @@ extension ErrorMessageGenerator {
...
+    let customErrorMessage: String = {
+      switch error {
+        case let err as LocalizedError where err.errorDescription != nil:
```
xwu, line 351:
```suggestion
      case let err as LocalizedError where err.errorDescription != nil:
```

rjstelling:
@xwu Thanks! It must of been very late 😂

Documentation/02 Arguments, Options, and Flags.md
```diff
@@ -178,7 +178,7 @@ struct Example: ParsableCommand {
 }
 ```
 
-Throw an error from the `transform` function to indicate that the user provided an invalid value for that type.
+Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. (See: [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) in [Validation and Errors](./05%20Validation%20and%20Errors.md) for details about customising `transform` function errors.)
```
natecook1000, line 181: Nits: This sentence doesn't need to be parenthesized, and let's use the American spelling here.
```suggestion
Throw an error from the `transform` function to indicate that the user provided an invalid value for that type. See [Handling Transform Errors](./05%20Validation%20and%20Errors.md#handling-transform-errors) for more about customizing `transform` function errors.
```

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
...
+
+struct Example: ParsableCommand {
+
+  // Reads in the argument string and attempts to transform it to
+  // a `ExampleDataModel` object using the `JSONDecoder`. If the
+  // string is not valid JSON, `decode` will throw an error and
+  // parsing will halt.
+  @Argument(transform: {
+    guard let data = $0.data(using: .utf8) else { throw ValidationError("Badly encoded string, should be UTF-8") }
+    return try JSONDecoder().decode(ExampleDataModel.self, from: data) })
```
natecook1000, line 129: This is a bit more code than I would like to encourage placing directly in the `@Argument` wrapper. Can you move it into the `ExampleDataModel`, either as an init or a static method that produces the type? Maybe `init(jsonString:) throws`, so that it doesn't match the `ExpressibleByArgument` initializer.

Documentation/05 Validation and Errors.md
```diff
@@ -100,3 +100,38 @@ struct Example: ParsableCommand {
...
+  var inputJSON: ExampleDataModel
+  
+  // Specifiying this option will always cause the parser to exit
+  // and print the custom error.
+  @Option(transform: { throw ExampleTransformError(description: "Trying to write to failOption always produces an error. Input: \($0)") })
+  var failOption: String?
+}
+```
```
natecook1000, line 137: Could you show an example of one or both of these errors in practice?

Tests/ArgumentParserEndToEndTests/TransformEndToEndTests.swift
```diff
@@ -0,0 +1,157 @@
...
+}
+
+fileprivate protocol Convert {
+    static func convert(_ str: String) throws -> Int
+}
+
+extension Convert {
+    
+    
+    static func convert(_ str: String) throws -> Int {
```
natecook1000, line 30: Nit: Close this up, and convert to two space indents in this file.
```suggestion
    static func convert(_ str: String) throws -> Int {
```

natecook1000:
@swift-ci Please test

rjstelling:
@natecook1000 Apologies, I requested a review and a commit had not been pushed. All corrected now.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Issue #112, opened 2020-04-08
Show an error for option and flag name collisions


natecook1000:
It's possible to define `ParsableCommand` and `ParsableArguments` type with colliding option and/or flag names. This should be an error that's surfaced by a `ParsableArgumentsValidator` type. For example, running this program should fail:

```swift
struct TwoOfTheSameName: ParsableCommand {
    @Option()
    var foo: String
    
    @Option(name: .customLong("foo"))
    var notActuallyFoo: String
}

TwoOfTheSameName.main()
```

Note that each option and flag can have more than one name. We'll eventually need to validate this same problem through the command hierarchy, but validating individual commands is a good first step.

iainsmith:
Looks like this can be closed now 😄

PR #118, opened 2020-04-16
Unique argument names validator

toddthomas:
### Abstract
Adds `ParsableArgumentsUniqueNamesValidator`, which conforms to `ParsableArgumentsValidator` and implements the validation requested in https://github.com/apple/swift-argument-parser/issues/112: ensuring that argument names aren't duplicated in a type conforming to `ParsableArguments` or `ParsableCommand`.

### Problem
Custom names can be used to give two or more options or flags the same name. For example:
```swift
import ArgumentParser

struct TwoOfTheSameName: ParsableCommand {
    @Option()
    var foo: String

    @Option(name: .customLong("foo"))
    var notActuallyFoo: String

    func run() throws {
        print("foo: \(foo)")
        print("notActuallyFoo: \(notActuallyFoo)")
    }
}

TwoOfTheSameName.main(["--foo thing"])
```
gives this output in `master`:
```
Error: Missing expected argument '--foo <foo>'
Usage: two-of-the-same-name --foo <foo> --foo <foo>
Program ended with exit code: 64
```

### Proposed solution
With the changes in this PR, the output is now
```
Fatal error: Validation failed for `TwoOfTheSameName`:
- Multiple (2) `Option` or `Flag` arguments are named "foo".
: file /Users/tthomas/Library/Developer/Xcode/DerivedData/argument-uniqueness-test-fuopasnbytbtladimilqpbwgllpu/SourcePackages/checkouts/swift-argument-parser/Sources/ArgumentParser/Parsable Types/ParsableArguments.swift, line 196
```
(The above is actually printed twice, for reasons I haven't yet investigated, but that also happens for the validators currently in `master`.)

### Discussion
I used the existing `ParsableArguments` validation implementation, but that means the reflection code used to obtain a `ParsableArguments`'s `ArgumentSet` is now duplicated four times in the code. I'd like to try a refactor to DRY that up if that's desirable.
 
### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -192,10 +192,6 @@ struct ParsableArgumentsUniqueNamesValidator: ParsableArgumentsValidator {
   struct Error: Swift.Error, CustomStringConvertible {
     var duplicateNames: [String: Int] = [:]
 
-    var occurred: Bool {
-      !duplicateNames.isEmpty
-    }
-
     var description: String {
       duplicateNames.map { entry in
         "Multiple (\(entry.value)) `Option` or `Flag` arguments are named \"\(entry.key)\"."
@@ -230,15 +226,9 @@ struct ParsableArgumentsUniqueNamesValidator: ParsableArgumentsValidator {
       }
     }
 
-    var error = Error()
-
     let duplicateNames = countedNames.filter { $0.value > 1 }
     if !duplicateNames.isEmpty {
-      error.duplicateNames = duplicateNames
-    }
-
-    if error.occurred {
-      throw error
+      throw Error(duplicateNames: duplicateNames)
     }
   }
 }
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -30,7 +30,7 @@ extension ParsableArguments {
     let validators: [ParsableArgumentsValidator.Type] = [
       PositionalArgumentsValidator.self,
       ParsableArgumentsCodingKeyValidator.self,
-      ParsableArgumentsUniqueNamesValidator.self
+      ParsableArgumentsUniqueNamesValidator.self,
     ]
     let errors: [Error] = validators.compactMap { validator in
       do {
--- a/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
@@ -29,7 +29,8 @@ extension ParsableArguments {
   static func _validate() throws {
     let validators: [ParsableArgumentsValidator.Type] = [
       PositionalArgumentsValidator.self,
-      ParsableArgumentsCodingKeyValidator.self
+      ParsableArgumentsCodingKeyValidator.self,
+      ParsableArgumentsUniqueNamesValidator.self
     ]
     let errors: [Error] = validators.compactMap { validator in
       do {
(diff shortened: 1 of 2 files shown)


Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -185,3 +186,63 @@ struct ParsableArgumentsCodingKeyValidator: ParsableArgumentsValidator {
...
+struct ParsableArgumentsUniqueNamesValidator: ParsableArgumentsValidator {
+  struct Error: Swift.Error, CustomStringConvertible {
+    var duplicateNames: [String: Int] = [:]
+
+    var occurred: Bool {
+      !duplicateNames.isEmpty
+    }
+
+    var description: String {
+      var description = duplicateNames.reduce(into: "") { description, entry in
```
natecook1000, line 200: You could write this as `map` and then `joined` to avoid the `removeLast` after building the string.

toddthomas: Thank you for spotting this. The `reduce` was a remnant from when this error type was more complicated, but after I simplified it I didn't rethink this part.

That extra complication is something I wanted to discuss. This `Error` used to be an enum that also represented the case where the user defined an option or flag named `"h"` or `"help"`, in which case the built-in help is masked. For example,
```
import ArgumentParser

struct RedefinedHelp: ParsableCommand {
    @Flag()
    var help: Bool

    func run() throws {
        print("help: \(help)")
    }
}

RedefinedHelp.main(["--help"])
```
gives
```
help: true
```
But I had problems detecting this reliably in the validator, so I removed that code. But now I think I might know what the problem was, so I'll give it another try, if you think detecting this case is a good idea.

Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
```diff
@@ -193,4 +193,127 @@ final class ParsableArgumentsValidationTests: XCTestCase {
...
+    enum Numbers: Int, ExpressibleByArgument {
+      case one = 1
+      case two
+      case three
+    }
+
+    @Option(name: .customLong("foo"))
+    var alsoNotFoo: Numbers
+  }
+
```
rjstelling, line 263: There are no tests for `@Flag`s using `CaseIterable` enums. The code does the right thing but having a test case might be useful in the future.

```swift
enum ExampleEnum: String, ExpressibleByArgument, CaseIterable {
    case first
    case second
    case other
    case forth
    case fith
}

  @Flag(name: .short, default: .first) //this should fail
  var enumFlag: ExampleEnum

  @Flag(name: .long, default: .first) //this should pass
  var enumFlag2: ExampleEnum
}
```

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -185,3 +186,63 @@ struct ParsableArgumentsCodingKeyValidator: ParsableArgumentsValidator {
...
+          var codingKey = child.label,
+          let parsed = child.value as? ArgumentSetProvider
+          else { return nil }
+
+        // Property wrappers have underscore-prefixed names
+        codingKey = String(codingKey.first == "_" ? codingKey.dropFirst(1) : codingKey.dropFirst(0))
+
+        let key = InputKey(rawValue: codingKey)
+        return parsed.argumentSet(for: key)
+    }
```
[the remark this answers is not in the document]
toddthomas: The validation is kicked off in that particular `ArgumentSet` initializer, so using it in a validator causes infinite recursion. I think that's why the reflection code is copied from the initializer into each of the validators. That's the duplication I mentioned I'd like to factor out.

I have a rough idea that could be done by converting the validators to functors with signatures like `(ArgumentSet) throws -> Void` (except that `ParsableArgumentsCodingKeyValidator` needs to be something like `([String]) throws -> Void` because it iterates over the coding keys). Then those functors could be called from the points in the reflection code where the coding keys or argument sets they need are available (for debug only, as currently). But I need to think about that more, because I'm not sure whether there are other validators we'd want that would have different signatures. I wouldn't want to have an explosion of validator types and places they're being called from.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -29,7 +29,8 @@ extension ParsableArguments {
   static func _validate() throws {
     let validators: [ParsableArgumentsValidator.Type] = [
       PositionalArgumentsValidator.self,
-      ParsableArgumentsCodingKeyValidator.self
+      ParsableArgumentsCodingKeyValidator.self,
+      ParsableArgumentsUniqueNamesValidator.self
```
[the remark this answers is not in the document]
toddthomas: Optimism FTW! Done.

Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
toddthomas: Thank you very much--I didn't know about this potential source of conflict. Tests added for the long and short name cases.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
toddthomas: I pushed a commit to simplify using `map`, as suggested. Much nicer! I haven't yet revisited the idea of detecting the case where the user defines their own `help` flag or option. I may have a chance to do that today.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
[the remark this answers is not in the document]
YuAo: Other possible approaches would be:
- Moving the `_validate()` method to another location. 
- Moving the duplicated code to a helper function.

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
```diff
@@ -185,3 +186,59 @@ struct ParsableArgumentsCodingKeyValidator: ParsableArgumentsValidator {
...
+    var error = Error()
+
+    let duplicateNames = countedNames.filter { $0.value > 1 }
+    if !duplicateNames.isEmpty {
+      error.duplicateNames = duplicateNames
+    }
+
+    if error.occurred {
+      throw error
+    }
```
YuAo, line 242: How about creating and throwing the error in `if !duplicateNames.isEmpty { ... }`? Then we can remove `error.occurred` which is actually `!duplicateNames.isEmpty`.

```
if !duplicateNames.isEmpty {
  throw Error(duplicateNames: duplicateNames)
}
```

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
natecook1000: That case is a clean override, and supported, so we wouldn't want to error there. The user-defined flag/option should always win, so please file a bug if you see the other behavior!

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Parsable Types/ParsableArgumentsValidation.swift
toddthomas: Yes, thank you. This is another remnant of `Error` having been a more complicated type. I really needed other sets of eyes on this.

toddthomas: Done.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Issue #134, opened 2020-05-06
Show full help when the user leaves off all arguments


natecook1000:
When running a command-line tool, if the user doesn't specify any inputs at all it's a good bet that they aren't familiar with the command. If there are required arguments/options/flags, we should show the error message with the full help screen instead of just the abbreviated usage.

For example, if you run the example `repeat` command with no inputs, you get this:

```
$ swift run repeat
Error: Missing expected argument '<phrase>'
Usage: repeat [--count <count>] [--include-counter] <phrase>
```

Instead, it would be great to show this:

```
$ swift run repeat
Error: Missing expected argument '<phrase>'

USAGE: repeat [--count <count>] [--include-counter] <phrase>

ARGUMENTS:
  <phrase>                The phrase to repeat. 

OPTIONS:
  --count <count>         The number of times to repeat 'phrase'. 
  --include-counter       Include a counter with each repetition. 
  -h, --help              Show help information.
```

mpapple-swift:
Looks like #135  addresses this issue.

richiwalt:
I like this suggestion @natecook1000

PR #135, opened 2020-05-06
Introduces `defaultToHelp` command configuration

erica:
Adds configuration option that allows a command to present help when
no other input is provided by the user at the command-line call.

### Description
This change introduces a hook so users can configure the command to
respond with a help message on the absence of arguments.

### Detailed Design

```swift
// CommandConfiguration:
// ...
  /// A Boolean value indicating whether this command should display help
  /// when executed without further user input.
  public var defaultToHelp: Bool
// ...
```

### Documentation Plan
Updated code doc in CommandConfiguration.swift

### Test Plan
Test with updated math.swift. Instead of returning 0 without arguments, it should return the help message.

### Source Impact
This is purely additive and optional, defaulting to `false`, the existing behavior.

### Checklist
- [x ] I've added at least one test that validates that my change is working, if appropriate
- [x ] I've followed the code style of the rest of the project
- [x ] I've read the [Contribution Guidelines](CONTRIBUTING.md)

I have not updated the documentation, but will amend this PR if the idea has a chance for moving forward.

### DIFF
--- a/Examples/math/main.swift
+++ b/Examples/math/main.swift
@@ -20,6 +20,9 @@ struct Math: ParsableCommand {
 
         // Commands can define a version for automatic '--version' support.
         version: "1.0.0",
+        
+        // Show help on no input.
+        defaultToHelp: true,
 
         // Pass an array to `subcommands` to set up a nested tree of subcommands.
         // With language support for type-level introspection, this could be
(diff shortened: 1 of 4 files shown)


natecook1000:
Hi @erica — thanks for the PR! We’re trying to constrain the configuration options to only things that aren’t really possible to handle from within client code, so I don’t think this is an option we want to add right now. 

To get this functionality yourself, you can `throw CleanExit.helpRequest()` from your `run()` or `validate` methods.

There’s a very-closely related issue covered in #134 if you’d like to tackle that. We also will need an API to directly generate the help text from a command.

erica:
`throw CleanExit.helpRequest()` worked beautifully for my immediate needs.

natecook1000:
I think this is addressed by #140 and #142 — feel free to re-open if you want to discuss any more!

PR #140, opened 2020-05-13
Display help when no arguments results in error

john-mueller:
This implements the feature mentioned in #134. I think it would also remove the need for #135.

When a `ParserError` is thrown, check to see if there are zero arguments. If so, the error is wrapped in the new `.noArguments` case, which allows the original error to be reported, while also triggering the full help message instead of usage.

If there's a Swiftier/more concise way of writing the addition to MessageInfo.swift, I'd love to know.

I wasn't sure whether this needs explicit mention in the documentation, and if so, which section to add it to.

### Checklist


natecook1000:
Thanks, @john-mueller! This looks great to me. I don't think it requires any change in documentation.

natecook1000:
@swift-ci Please test

PR #141, opened 2020-05-14
Fix short names with equals sign

### LINKED ISSUE #132: Short name arguments don't accept an equal sign between key/value pairs
**ArgumentParser version:** `0.0.5`, `master`
**Swift version:** 

```
Apple Swift version 5.2.2 (swiftlang-1103.0.32.6 clang-1103.0.32.51)
Target: x86_64-apple-darwin19.4.0
```

### Checklist

### Steps to Reproduce

Defining an option with a `.customShort` name doesn't support an equal sign (`=`) between key-value pairs.

### Expected behavior
Specifying an option with an equal sign in between key-value pairs should succeed, as noted in the [documentation](https://github.com/apple/swift-argument-parser/blob/master/Documentation/02%20Arguments%2C%20Options%2C%20and%20Flags.md#declaring-arguments-options-and-flags):

>Options are named key-value pairs. Keys start with one or two dashes (- or --), and a user can separate the key and value with an equal sign (=) or a space.

### Actual behavior

The command line execution fails when `cool -r=4` is attempted:

```
Error: Unknown option '-r'
Usage: cool [--repeat-count <repeat-count>]
```

Here is a sample `main.swift` you can use to reproduce:

```swift
import Foundation
import ArgumentParser

struct BeCool: ParsableCommand {
    static let configuration = CommandConfiguration(
        commandName: "cool",
        abstract: "Be super cool",
        discussion: "Demonstrates how to be cool",
        version: "0.0.1",
        shouldDisplay: true)
    
    
    /** OPTIONS */
    // [FAIL]  -r=4
    // [OK]    -r 4
    // [OK]    --repeat-count=4
    @Option(
        name: [.customShort("r"), .long],
        help: "Number of times the word \"Cool\" is repeated. Defaults to 1."
    )
    var repeatCount: Int?
    
    func run() throws {
        
        let outputString = String(repeating: "Cool ", count: repeatCount ?? 1)
        
        print(outputString)
    }
}

BeCool.main()
```

john-mueller:
This fixes #132, by letting you separate a short option and its value with an equals sign.

The new tests are in EqualsEndToEndTests.swift for now, but arguably they belong in ShortNameEndToEndTests.swift. I can move them if necessary. I also omitted new tests for long argument names with equals sign, as that seems to be tested elsewhere.

I don't think there needs to be any further documentation, as this was already the documented behavior.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/SplitArguments.swift
+++ b/Sources/ArgumentParser/Parsing/SplitArguments.swift
@@ -493,7 +493,14 @@ private extension ParsedArgument {
   }
   
   init(longArgWithSingleDashRemainder remainder: Substring) throws {
-    try self.init(longArgRemainder: remainder, makeName: { Name.longWithSingleDash(String($0)) })
+    try self.init(longArgRemainder: remainder, makeName: {
+      /// If an argument has a single dash and single character,
+      /// followed by a value, treat it as a short name.
+      ///     `-c=1`      ->  `Name.short("c")`
+      /// Otherwise, treat it as a long name with single dash.
+      ///     `-count=1`  ->  `Name.longWithSingleDash("count")`
+      $0.count == 1 ? Name.short($0.first!) : Name.longWithSingleDash(String($0))
+    })
   }
   
   init(longArgRemainder remainder: Substring, makeName: (Substring) -> Name) throws {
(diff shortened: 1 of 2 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
Thanks for this fix, @john-mueller!

PR #147, opened 2020-05-14
Set isOptional flag for options of Optional type

aleksey-mashanov:
Documentation for `Option.init(... transform:)` says: "If the property has an `Optional` type, or you provide a non-`nil` value for the `initial` parameter, specifying this option is not required."
Initializer works exactly this way but provides wrong synopsis for `Optional` type properties.

### Checklist


natecook1000:
Thanks for the catch and the fix, @aleksey-mashanov! 👏

natecook1000:
@swift-ci Please test

PR #162, opened 2020-05-21
Remove `Never` return from ParsableCommand.main()

natecook1000:
### Description
This brings the `main()` method in line with using the new `@main` attribute on the root
command type.

### Detailed Design
The changes here are to make `ParsableCommand.main()` a `Void`-returning function, to break out a separate no-argument version, and to remove the default parameter. The default-argument version doesn't satisfy the `() -> Void` requirement for `@main`.

The `main()` method now returns without exiting on successful execution of a 

### Documentation Plan
No real impact. Documentation for using `@main` with the library to follow.

### Test Plan
All existing tests continue to pass. SwiftPM will need an update to support `@main` in executable targets.

### Source Impact
Typical code should continue to behave as expected, but this is source-breaking if anyone has a strongly-typed usage of `ParsableCommand.main()`, e.g. `let mainFunction: ([String]?) -> Never = MyCommand.main`. Keeping the old version with a deprecation/unavailability annotation isn't possible, since there would be ambiguity between that and the new ones.

In addition, code that followed `MyCommand.main()` in `main.swift` is no longer dead code, and will be executed after successful execution of the command.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -84,18 +84,24 @@ extension ParsableCommand {
   }
 
   /// Parses an instance of this type, or one of its subcommands, from
-  /// command-line arguments and calls its `run()` method, exiting cleanly
-  /// or with a relevant error message.
+  /// the given arguments and calls its `run()` method, exiting with a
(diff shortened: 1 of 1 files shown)


natecook1000:
@swift-ci Please test

PR #163, opened 2020-05-21
Make ParsableCommand.run() a mutating method

natecook1000:
### Description
This changes `ParsableCommand`'s `run` requirement to be mutating. This allows for things like in-place sorting or filtering of arguments without making a mutable copy.

This change supersedes #33.

### Detailed Design
The only change here is the switch to mutating for the `run()` method.

```swift
protocol ParsableCommand: ParsableArguments {
    // other requirements

    /// ...
    mutating func run() throws
}
```

### Documentation Plan
We didn't previously mention the non-mutatingness of `run()`, so that doesn't require a change. All code samples have been updated to implement the new requirement.

### Test Plan
All existing tests pass with the change.

### Source Impact
~This **will require code changes** for every command that implements a `run()` method, since a non-mutating `run()` won't properly satisfy the new requirement. Since `ParsableCommand` provides a default implementation for `run()`, there unfortunately won't be any indication of this change at compile time.~

Apparently this _is_ in fact source-compatible with existing code. Clients can add `mutating` as they want, but existing non-mutating `run()` methods will be called correctly. 🎉

### Checklist

### DIFF
--- a/Documentation/01 Getting Started.md
+++ b/Documentation/01 Getting Started.md
@@ -52,7 +52,7 @@ struct Count: ParsableCommand {
     @Argument()
     var outputFile: String
     
-    func run() throws {
+    mutating func run() throws {
         print("""
             Counting words in '\(inputFile)' \
             and writing the result into '\(outputFile)'.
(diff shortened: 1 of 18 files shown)


natecook1000:
@swift-ci Please test

PR #170, opened 2020-05-30
Allow normal Swift default property initialization syntax

MPLew-is:
This change allows the normal `var foo = "blah"` default initialization syntax for `Option`s, as a parallel initialization method as using the `default` parameter.

This syntax may feel more natural to some developers (myself included), but does not break any existing code using the original initialization method.

### Checklist


MPLew-is:
I've tested this with this simple change, and it seems to work as expected (and all the test cases still pass locally); however, I want to make sure this is something that would be desired before continuing to extend this to the other parameter types.

MPLew-is:
A simple implementation to test:

```swift
import ArgumentParser

struct SomeCli: ParsableCommand {
	@Option()
	var data: String = "test"

	func run() throws {
		print(data)
	}
}

SomeCli.main()
```

```sh
$ swift run cli
test
$ swift run cli --data=test2
test2
```

natecook1000:
@swift-ci Please test

MPLew-is:
Looks like everything passed in CI, so is this something that would be desired for the remaining parameter types?

natecook1000:
Thanks for starting this implementation! It’s been on my to-do list to investigate, but I apparently hadn’t opened an issue. I have a couple in-flight changes that you might want to wait for, most notably [removing `ExpressibleByArgument` conformance for `Optional`](https://github.com/apple/swift-argument-parser/tree/nate/optional_non_expressible), which will change the list of inits for `@Argument` and `@Option`. I’ll try to get that in so you can keep moving forward with this.

The biggest downside I see to this change is that it can effectively double the number of inits in some cases, where the `default: Value? = nil` parameter needs to be replaced by two separate inits (one with a `wrappedValue` parameter and one without). That said, I think the improved syntax is worth it as long as type checking/etc still works out.

MPLew-is:
Given those concerns, would it maybe make sense to drop the current `default:` syntax entirely, assuming that this change works as I intend it to? I realize it would break a lot of existing usage, but this is also in very early development so it may be the best time to do it if you feel this syntax is an improvement over the current one.

And honestly I’ll probably push ahead with the development on this more just as an exercise for myself, so don’t rush your other changes on my account; once they’re done, I can always integrate them back into this.

kylemacomber:
I think this is a great direction

natecook1000:
> Given those concerns, would it maybe make sense to drop the current default: syntax entirely, assuming that this change works as I intend it to? I realize it would break a lot of existing usage, but this is also in very early development so it may be the best time to do it if you feel this syntax is an improvement over the current one.

I definitely agree that we'd just want the one syntax long term. To help existing users transition, we'll want to go from the status quo to three different inits:

```swift
// current
init(default: Value? = nil, ...)

// new
init(wrappedValue: Value, ...)
init(...)
@available(*, deprecated, message: "Use regular property initialization for default values")
init(default: Value?, ...)
```

There's ample room for code deduplication in the initializers, but that can definitely come after this change.

> And honestly I’ll probably push ahead with the development on this more just as an exercise for myself, so don’t rush your other changes on my account; once they’re done, I can always integrate them back into this.

That change landed in #173, so you've got a clear field for this. Thanks again!

MPLew-is:
I'll continue work on this tomorrow/over the next week, but I believe `Option`s are now fully ready for review so please let me know if you have any feedback.

john-mueller:
I just submitted #186, which adds the `default` parameter to the last two initializers for `Option` and `Argument` (when value is an array), so those would have to be reworked as well. 😬

I like this direction for initialization over the `default` parameters.

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -70,6 +70,26 @@ extension Option: DecodableParsedWrapper where Value: Decodable {}
...
+    self.init(_parsedValue: .init { key in
+      ArgumentSet(
+        key: key,
+        kind: .name(key: key, specification: name),
+        parsingStrategy: ArgumentDefinition.ParsingStrategy(parsingStrategy),
+        parseType: Value.self,
+        name: name,
+        default: initial, help: help)
+      })
+  }
```
natecook1000, line 91: Nice job pulling this out 👍

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -79,21 +99,67 @@ extension Option where Value: ExpressibleByArgument {
   ///   - parsingStrategy: The behavior to use when looking for this option's
   ///     value.
   ///   - help: Information about how to use this option.
+  @available(*, deprecated, message: "Use regular property initialization for default values (`var foo: String = \"bar\"`)")
   public init(
     name: NameSpecification = .long,
     default initial: Value? = nil,
```
natecook1000, line 105: We can drop the default value for this parameter since you're adding `init(name:parsing:help:)`. I think overload resolution chooses that one over this because of the number of parameters, but it isn't strictly necessary anymore.

Sources/ArgumentParser/Parsable Properties/Option.swift
MPLew-is: Thanks! And not sure what happened in the `transform` equivalent of this; I had that mirroring this one but I guess didn't actually stage/commit all of it over the weekend so it was in a broken state here 😅 

That's been fixed and pushed now though.

Sources/ArgumentParser/Parsable Properties/Option.swift
MPLew-is: Done in 28d7179

Sources/ArgumentParser/Parsable Properties/Flag.swift
```diff
@@ -196,16 +223,84 @@ extension Flag where Value == Bool {
...
+  ///   - help: Information about how to use this flag.
+  public init(
+    name: NameSpecification = .long,
+    inversion: FlagInversion,
+    exclusivity: FlagExclusivity = .chooseLast,
+    help: ArgumentHelp? = nil
+  ) {
+    self.init(
+      name: name,
+      initial: false,
```
MPLew-is, line 299: @natecook1000 I just want to point out one thing here that we should figure out how to resolve before this gets merged. I've preserved as much of the existing intent as possible, but the old `@Flag(default: nil)` is no longer possible now on this commit. I personally think it'd be better to remove the implicit default-false here:

```diff
-      initial: false,
+      initial: nil,
```

but that would necessarily mean that there would have to be source changes like this:

```diff
 @Flag(inversion: .prefixedNo)
+var someFlag: Bool
-var someFlag: Bool = false
```

or else anything that assumed that this would have some value when no flag was given would break at parse-time.

Do you have any thoughts here?

MPLew-is:
Other than the issue I've pointed out above, this should be functionally complete. I'll work on the high-level documentation next, but please let me know if you see any other changes I should make!

Sources/ArgumentParser/Parsable Properties/Flag.swift
MPLew-is: Let me rephrase some of that for clarity:

The actual source breakage is relatively unimportant since we're already deprecating the old method, but the intent of what to do with a `Flag` if completely uninitialized (by either default value or parsing) is what I want to make sure I understand. I personally feel that an implicit default on a non-`Optional` property is less Swifty than making someone specify an explicit default as a fallback just as you would see in a function call, for instance. Forcing users to specify a default would also resolve another inconsistency I see, that `@Flag(default: nil) var someFlag: Bool` looks on a high level like it would be attempting to assign a `nil` to a non-`Optional` property; this is also compounded by this behavior not being valid for the other argument types. In effect, `nil` is being overloaded here in the public-facing API to be a selector for different parsing behavior, which feels relatively un-Swifty to me.

All of that above leads me to feel that we should change the behavior of `Flag` to have no default value in the public API without the user explicitly setting one (which would be setting the default to `nil` internally). I do realize, though, that changing this behavior will both affect existing users and create a potential minor usability hit for new users, since the most common case definitely is for a flag to be default-false, and would definitely appreciate hearing your opinion as to if that’s a good trade-off.

natecook1000: Great point about the flag initializers. The intent of the existing design is to handle the two different kinds of Boolean flags, both without an inversion and with an inversion. The tricky part is these act a bit differently — here's what the semantics are, and why the current design is what it is:

_No inversion:_ This kind of flag is always optional (for the user) and defaults to `false`; if the user provides the flag, it toggles to `true`. If you were to specify a default of `true`, the flag would never have any effect, so we removed that parameter from the no-inversion `init`. Likewise, if the flag were required, it would also have no effect, since the resulting property would always be `true`.

_With inversion:_ The user can explicitly set this flag to either `true` or `false`, so the default can be either of these, or left out entirely (i.e. `nil`). If the default is `nil`, then one of the two flags is required; if there's a default value, then the flags are optional for the user.

With all that said, what should flags look like with this new way of initializing properties? I agree with you that once we're doing regular Swift default initialization, having uninitialized properties automatically get a value is weird. I think the kinds of declarations we would want to support would be:

```swift
// No inversion
@Flag var regular: Bool = false

// Inversion
@Flag(inversion: .prefixedNo) var required: Bool
@Flag(inversion: .prefixedNo) var defaultTrue: Bool = true
@Flag(inversion: .prefixedNo) var defaultFalse: Bool = false

// Unfortunately, you can also write:
@Flag var alwaysTrue: Bool = true
```

I think we could print a warning on that last case, recommending that they change the default to `false` or provide an inversion. For initializers, we'll want these two, plus any deprecated ones to maintain source-compatibility:

```swift
// cases 1, 3, 4, 5
init(wrappedValue: Bool, inversion: FlagInversion? = nil, ...)
// case 2
init(inversion: FlagInversion, ...)
```

MPLew-is: I think the warning for the `alwaysTrue` case is probably the best way to go; it could also be a parsing error instead, but I can see a use for allowing flags to be pinned to a value for quick testing purposes. In the absolutely pie-in-the-sky ideal scenario it seems best (to me) to have the no-inversion case be represented by something other than a boolean, something wrapped in an `Optional` that is only capable of having a single value. I'm thinking of something akin to:
```swift
enum SingleValue {
    case some
}
@Flag someFlag: SingleValue?
```
That preserves the ability to leave off a default for that case (since it'll default to `nil`), and then allow the default for `Bool` flags to be `inversion: .prefixedNo` and also just represents the high-level usage more (flag present is some value, flag not present is no value). But I realize that will probably not be incredibly intuitive for users to deal with, and opens up further questions of how to easily communicate how to migrate existing code.

So, I think the best next steps for me are pretty much exactly what you have described above:
1. Extend support for default values to the no-inversion case (which is not currently in this PR), with a warning if that value is set to `false`
    - This should be accomplished by combining those initializers, as you indicated
2. Deprecate not having a default value on no-inversion flags in favor of making the user provide an explicit `= false`, but keep it around for now to not completely break existing source
3. Change the intent of flags with inversions such that they no longer default to `false` but to being required if no default value is provided, as in your example above
    - We unfortunately won't be able to provide any sort of deprecation message or other warnings for this, but I also don't know how large a concern that is at this point.

Do you have any other thoughts? I can probably have all that implemented in the next day or so otherwise.

natecook1000: > But I realize that will probably not be incredibly intuitive for users to deal with, and opens up further questions of how to easily communicate how to migrate existing code.

Exactly, this would more accurately model the correct behavior, but at the cost of making the library harder to use.

> 1. Extend support for default values to the no-inversion case (which is not currently in this PR), with a warning if that value is set to `false`
> - This should be accomplished by combining those initializers, as you indicated

I think you mean warning on `true`, but yes on this point. 👍 

> 2. Deprecate not having a default value on no-inversion flags in favor of making the user provide an explicit = false, but keep it around for now to not completely break existing source
> 3. Change the intent of flags with inversions such that they no longer default to false but to being required if no default value is provided, as in your example above
> - We unfortunately won't be able to provide any sort of deprecation message or other warnings for this, but I also don't know how large a concern that is at this point.

This is unfortunate — I hadn't remembered that leaving the default off of a flag with an inversion keeps the default as `false`. We want to change the semantics of that same line of code to be a required flag, but I don't think there's a way to provide a deprecation warning to guide users to the right syntax, which means this will necessarily be a breaking change. I think your plan forward makes sense — I'll play around with a couple different options to communicate the migration path (which may just be a statement in the changelog).

MPLew-is:
This should be ready for review, let me know if you find any issues!

Sources/ArgumentParser/Parsable Properties/Flag.swift
MPLew-is: > I think you mean warning on true, but yes on this point. 👍

Whoops my bad there, I definitely did mean `true`.

All of this is implemented in 83a70e4 and e7397e9/

natecook1000:
@swift-ci Please test

natecook1000:
This is looking great, @MPLew-is! I've reconciled changes with the array defaults addition in #186 — going to merge and then continue that integration. 🎉🎉

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test Linux platform

PR #173, opened 2020-06-01
Remove `ExpressibleByArgument` conformance from `Optional`

natecook1000:
### Description
This drops the `ExpressibleByArgument` conformance for `Optional`. This conformance doesn't quite make sense, and allowed property declarations like this:

```swift
@Option(default: 0) var count: Int?
```

The default value meant that `count` would never be `nil`, and therefore shouldn't be optional.

### Detailed Design
This change adds initializers specifically for when an `Option`'s or `Argument`'s value is optional, without a `default` parameter as well as deprecated versions that include a `default` parameter to guide clients to the correct declaration syntax.

### Documentation Plan
The new initializers are documented; related initializers have been updated. There is no change required in the guides.

### Test Plan
Existing tests exercise the functionality and verify the largely source-compatible nature of the change.

### Source Impact
While this is technically source-breaking, it should have minimal impact in regular usage. All the declaration sites for arguments and options with optional values are still supported, with deprecation warnings for declarations that combine a default value and an optional type. If a property is declared as optional but does not have a default value, there won't even be a deprecation notice.

Code that uses `Optional`'s conformance to `ExpressibleByArgument`, however, will break without warning. Users can work around this by adding an `Optional`-specific code path, or by other means.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -141,14 +141,14 @@ extension Argument {
     help: ArgumentHelp? = nil
   ) where Value == T? {
     self.init(_parsedValue: .init { key in
-      ArgumentSet(
+      var arg = ArgumentDefinition(
         key: key,
         kind: .positional,
         parsingStrategy: .nextAsValue,
-        parseType: T.self,
-        name: .long,
-        default: nil,
-        help: help)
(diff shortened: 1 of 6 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Issue #167, opened 2020-05-23
Prefer the first `long` specifier in help


erica:
My help message (in part):

> When it's that time there: "now --remote 5PM Bath UK"
> USAGE: now [--local <local>] [--when <when>] [<location-info> ...]
> ...
>   -r, --remote, --remote-time, --there, --when <when>

My declaration:

```
name: [.short, .customLong("remote"),  .customLong("when"), .customLong("there")],
```

I'd like the SAP to prefer the first long specification, placing it at the end of the alphabetized items. That would allow my help message to look like this instead:

> When it's that time there: "now --remote 5PM Bath UK"
> USAGE: now [--local <local>] [--remote <remote>] [<location-info> ...]
> ...
>   -r, --there, --when, --remote <remote>

ibrahimoktay:
Hi @natecook1000 ,

I can create a PR for this bug. Just want to make sure we are on the same page. In NameSpecification, "Set" has been used as data structure. Therefore, we cannot preserve the first long specification. So, I can replace it with Array<Element>. After that,

In UsageGenerator,
1. I can only change "preferredNameForSynopsis" and return the first long specification
2. Or I can change "sortedNames" and keep sorting as it is right now, except the first long specification and append it later.

I think, second option is better and I can confirm that these changes does not break any test. If there is a third option, I can help with that too.

natecook1000:
Thanks, @ibrahimoktay! Could you use option 1 instead? I have a related change in the works in #123, so we need to head that direction eventually.

john-mueller:
It seems like there are three possibilities for the ordering in help:
1. Alphabetical (current, if I understand correctly)
2. In source-code order.
3. First `.long` at end (@erica's proposal), presumably with the others in source-code order.


To me, the third option seems counter-intuitive. I think the second option would better match expectations, and still gives @erica the option to definitively order the specifiers.

natecook1000:
I had missed the ordering in the OP, and agree with @john-mueller here — usage should choose the first long name, if available, or otherwise the first short name, and the help screen should show short names, then long names, in the order of their declaration.

ibrahimoktay:
I agree with @john-mueller, too. 
name: [.short, .customLong("remote"),  .customLong("when"), .customLong("there")]
What about value string? @natecook1000 

-t, --remote, --when, --there <remote>
or
-t, --remote, --when, --there <there>

natecook1000:
[...]

PR #179, opened 2020-06-04
Use the first long name in usage

ibrahimoktay:
SortedNames for ArgumentDefinition has been changed.
NameSpecification now uses Array type for elements.

* Usage will choose the first long name, if available, or otherwise the first short name
* Help screen will show short names, then long names, in the order of their declaration.

Solves (#167)

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
+++ b/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
@@ -41,13 +41,7 @@ public struct NameSpecification: ExpressibleByArrayLiteral {
   var elements: Array<Element>
   
   public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
-    let array = Array(sequence)
-    self.elements = Array()
-    for element in array {
-      if !self.elements.contains(element) {
-        self.elements.append(element)
-      }
-    }
+    self.elements = sequence.uniquified()
   }
   
   public init(arrayLiteral elements: Element...) {
--- a/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
+++ b/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
@@ -41,7 +41,13 @@ public struct NameSpecification: ExpressibleByArrayLiteral {
   var elements: Array<Element>
   
   public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
-    self.elements = Array(sequence)
+    let array = Array(sequence)
+    self.elements = Array()
+    for element in array {
+      if !self.elements.contains(element) {
+        self.elements.append(element)
+      }
+    }
   }
   
   public init(arrayLiteral elements: Element...) {
--- a/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
+++ b/Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
@@ -38,10 +38,10 @@ public struct NameSpecification: ExpressibleByArrayLiteral {
     /// Short labels can be combined into groups.
     case customShort(Character)
   }
-  var elements: Set<Element>
+  var elements: Array<Element>
   
   public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
-    self.elements = Set(sequence)
+    self.elements = Array(sequence)
   }
   
   public init(arrayLiteral elements: Element...) {
--- a/Sources/ArgumentParser/Parsing/Name.swift
+++ b/Sources/ArgumentParser/Parsing/Name.swift
@@ -53,6 +53,10 @@ extension Name {
       return n
     }
   }
+  
+  var isShort: Bool {
+    return self == .short(self.valueString.first!)
+  }
 }
 
 // short argument names based on the synopsisString
--- a/Sources/ArgumentParser/Parsing/Name.swift
+++ b/Sources/ArgumentParser/Parsing/Name.swift
@@ -55,7 +55,12 @@ extension Name {
   }
   
   var isShort: Bool {
-    return self == .short(self.valueString.first!)
+    switch self {
+    case .short:
+      return true
+    default:
+      return false
+    }
   }
 }
 
--- a/Sources/ArgumentParser/Usage/UsageGenerator.swift
+++ b/Sources/ArgumentParser/Usage/UsageGenerator.swift
@@ -63,7 +63,7 @@ extension ArgumentDefinition {
     
     switch kind {
     case .named:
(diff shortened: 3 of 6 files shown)


Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
```diff
@@ -38,10 +38,10 @@ public struct NameSpecification: ExpressibleByArrayLiteral {
     /// Short labels can be combined into groups.
     case customShort(Character)
   }
-  var elements: Set<Element>
+  var elements: Array<Element>
   
   public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
-    self.elements = Set(sequence)
+    self.elements = Array(sequence)
```
natecook1000, line 44: It's great that these will be in a consistent order, although we want to preserve the uniqueness we were getting from `Set`. This fails the command validation, but could just be collapsed into a single `.short` name:

```swift
  @Option(name: [.short, .short])
  var example: Int
}
```

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -114,29 +114,11 @@ extension ArgumentDefinition {
   }
   
   var sortedNames: [Name] {
-    return names
```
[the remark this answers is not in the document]
ibrahimoktay: I missed that one. I assume, It will be the same for usage. Right?
If a single-dash long name comes first,
```
  @Option(name: [.short, .customLong("other", withSingleDash: true), .customLong("reminder"), .long])
  var example: Int
}
```

Then, usage will be like this
```
USAGE: example -other <other>
```

Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
ibrahimoktay: I will add this case to tests.

Sources/ArgumentParser/Usage/UsageGenerator.swift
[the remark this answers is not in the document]
natecook1000: Right! We should prefer the first long name declared, whether it has one or two dashes.

Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
```diff
@@ -38,10 +38,16 @@ public struct NameSpecification: ExpressibleByArrayLiteral {
...
   
   public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
-    self.elements = Set(sequence)
+    let array = Array(sequence)
+    self.elements = Array()
+    for element in array {
+      if !self.elements.contains(element) {
+        self.elements.append(element)
+      }
+    }
```
BenLeggiero, line 50: I feel like this would be easier to understand and also less complex if it were written like this:

```swift
  public init<S>(_ sequence: S) where S : Sequence, Element == S.Element {
    self.elements = []
    for element in Array(sequence) where !self.elements.contains(element) {
      self.elements.append(element)
    }
  }
```

ibrahimoktay: I think, This usage without where is the most commonly used style in the project. But I have also seen this kind of for-loop in one place. So, I am not sure If we should change it. Because this code block is also very easy to understand.

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -114,29 +114,11 @@ extension ArgumentDefinition {
...
-        case let (.longWithSingleDash(l), .longWithSingleDash(r)):
-          return l < r
-        }
-    }
+    return names.filter{ $0 == .short($0.valueString.first!) } + names.filter{ $0 != .short($0.valueString.first!) }
   }
   
   var preferredNameForSynopsis: Name? {
-    sortedNames.last
+    names.first{ $0 != .short($0.valueString.first!) } ?? names.first
```
natecook1000, line 121: See note above.

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -114,29 +114,11 @@ extension ArgumentDefinition {
...
-          return l < r
-        case (_, .short):
-          return true
-        case (.short, _):
-          return false
-        case let (.longWithSingleDash(l), .longWithSingleDash(r)):
-          return l < r
-        }
-    }
+    return names.filter{ $0 == .short($0.valueString.first!) } + names.filter{ $0 != .short($0.valueString.first!) }
```
natecook1000, line 117: Two notes here:

- Since we're breaking this array of names into two groups, first all the short names and then all the non-short names, let's pull the test for "this name is short" out into a property on `Name`.
- The returned array is no longer sorted, so we should change the name of the property as well. Alternatively, we could just put them in this order at initialization and get rid of the property altogether.

Sources/ArgumentParser/Parsable Properties/NameSpecification.swift
natecook1000: Uniquing these elements isn't specific to a sequence of `Name`s, so it would be best to extract this to a general purpose `Sequence` extension. That would live in `Sources/ArgumentParser/SequenceExtensions.swift`.

Sources/ArgumentParser/Usage/UsageGenerator.swift
ibrahimoktay: * I am not sure about what you mean by this "let's pull the test for "this name is short" out into a property on Name.". Sorry about my English. Do you mean something like  "Name.isShort" ? (If yes, Open for naming suggestions)

* I can rename it. "partitionedNames" maybe?. Alternative sounds good, too. But I am not sure about changing the original order.

natecook1000: `Name.isShort` would be great! And I think `partitionedNames` is better than `sortedNames` — hard to find a good name for this concept, exactly.

Sources/ArgumentParser/Parsing/Name.swift
```diff
@@ -53,6 +53,10 @@ extension Name {
       return n
     }
   }
+  
+  var isShort: Bool {
+    return self == .short(self.valueString.first!)
```
natecook1000, line 58: One last nit — this should use pattern-matching on `self` instead of an equality comparison. We don't need to be accessing `valueString` for this.

ibrahimoktay: Yes, you are right. Just changed that.

natecook1000:
@swift-ci Please test

PR #186, opened 2020-06-15
Enable default for option and argument arrays

### LINKED ISSUE #107: Add a default value for @Option when the type is an Array
There's currently no way that I've found to have a default option when an `@Option` is an array

```swift
@Option(parsing: .upToNextOption)
var foo: [String]
```

I'd like to do something like:

```swift
@Option(default: ["someValue"], parsing: .upToNextOption)
var foo: [String]
```

So even if this option is not set, we still get this information.

john-mueller:
The adds the `default` parameter to `@Option` and `@Argument` initializers when the value type is an array. Default values are overridden when passing explicit values on the command line. Default is still an empty array if `default` parameter is omitted and no explicit values are passed.

Closes #107

### Checklist


Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -204,44 +204,53 @@ extension Argument {
...
+  /// parameter is not specified.
   ///
   /// - Parameters:
+  ///   - initial: A default value to use for this property. If `initial` is
+  ///     `nil`, this option defaults to an empty array.
   ///   - parsingStrategy: The behavior to use when parsing multiple values
   ///     from the command-line arguments.
   ///   - help: Information about how to use this argument.
   public init<Element>(
+    default initial: Value? = nil,
```
natecook1000, line 217: For all these we should be able to just use `[]` as the default instead of using the optional. Since that’ll be visible in the declaration, we don’t need to cover it in the documentation, either.

Sources/ArgumentParser/Parsing/ArgumentSet.swift
```diff
@@ -315,6 +315,8 @@ extension ArgumentSet {
     
     try setInitialValues(into: &result)
     
+    let initialCounts = getInitialArrayCounts(in: result)
```
natecook1000, line 318: I'm worried about this post-processing to remove the default values. Could you take care of emptying the array in the closure generated by `ArgumentDefinition.Update.appendToArray` instead?

john-mueller: I moved the default clearing code to a custom `update` method on `ParsedValues` that is unique to arrays. Now it removes the defaults up front, when the first array value from the command line is parsed.

Sources/ArgumentParser/Parsable Properties/Argument.swift
john-mueller: Of course. I was trying to match the other initializers, and thought about this right after the PR.

natecook1000:
@swift-ci Please test

PR #187, opened 2020-06-15
Show option-based errors before unexpected positional errors

natecook1000:
This pushes any errors indicated by unexpected arguments after parsing out to the same late position. We were previously stopping immediately when the command is a leaf node; that isn't necessary and created an awkward second error path.

### Checklist


natecook1000:
@swift-ci Please test

PR #188, opened 2020-06-15
Allow parsing negative numbers

### LINKED ISSUE #31: How to pass a negative number as an argument?
Imagine a command like this

```swift
import ArgumentParser

struct Absolute: ParsableCommand {
    
    @Argument() var number: Int
    
    func run() throws {
        print(abs(number))
    }
}

Absolute.main()
```

you can run

```
command 5
```

and it works. but running

```
commant -5
```

fails with `Error: Missing expected argument '<number>'` which makes sense. 

I would have expected being able to pass the argument as `"-5"` but that doesn't work either. 

Any advice on how to handle this?

john-mueller:
Allows negative numbers to be passed as options and arguments.

Fixes #31 

### Checklist
- [X] I've updated the documentation if necessary (no need)

### DIFF
--- a/Sources/ArgumentParser/Parsing/SplitArguments.swift
+++ b/Sources/ArgumentParser/Parsing/SplitArguments.swift
@@ -76,12 +76,15 @@ enum ParsedArgument: Equatable, CustomStringConvertible {
 /// A parsed version of command-line arguments.
 ///
 /// This is a flat list of *values* and *options*. E.g. the
-/// arguments `["--foo", "bar"]` would be parsed into
-/// `[.option(.name(.long("foo"))), .value("bar")]`.
+/// arguments `["--foo", "bar", "-4"]` would be parsed into
+/// `[.option(.name(.long("foo"))), .value("bar"),
+/// .possibleNegative(value: "-4", option: .name(.short("4")))]`.
 struct SplitArguments {
   enum Element: Equatable {
     case option(ParsedArgument)
     case value(String)
+    /// An element that could represent a negative number or an option (or option group).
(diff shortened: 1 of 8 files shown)


john-mueller:
Hmm, good point. I wasn't thinking about `ip -6`, for example. Let me take another crack at it.

john-mueller:
Okay, this should be a fully robust solution. I added a `.possibleNegative` case to `SplitArguments.Element`. This preserves the potential value and option until enough context is known to determine which it is.

I also found and fixed a bug when using `ArrayParsingStrategy.upToNextOption` and `Name.longWithSingleDash` together. Those tests are added to LongNameWithShortDashEndToEndTests.swift.

Sources/ArgumentParser/Parsing/SplitArguments.swift
```diff
@@ -432,6 +444,16 @@ extension SplitArguments {
       elements.append((index, element))
     }
     
+    /// Append as `.possibleNegative` if it could be a negative value;
+    /// otherwise, append as `.option`.
+    func appendAsPossibleNegative(if isDashPrefixedNumber: Bool, value: String, option: ParsedArgument) {
```
natecook1000, line 449: Could we name this `appendOption(_:value:isPossibleNegative:)`?

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

More detail about the branch update - https://forums.swift.org/t/updating-branch-names/40412

PR #193, opened 2020-06-23
Nate/continue default initialization

natecook1000:
### Description
This integrates the changes in #170 and #186, moving array default values to the normal syntax.

### Detailed Design
This kind of declaration is now permitted:

```swift
struct Example: ParsableCommand {
    @Argument(help: "Your lottery numbers")
    var lotto: [Int] = [4, 8, 15, 16, 23, 42]
}
```

### Documentation Plan
Updated symbol documentation and guides.

### Test Plan
Unit tests

### Source Impact
The new defaults in #170 haven't been included in a release yet, so there's no source breakage. Declarations of array options or arguments will have a deprecation warning indicating that users should specify an empty array literal as the property default.

### Checklist
- [x] I've added at least one test that validates that my change is working, if appropriate
- [x] I've followed the code style of the rest of the project
- [x] I've read the [Contribution Guidelines](CONTRIBUTING.md)
- [x] I've updated the documentation if necessary

PR #196, opened 2020-06-25
Allow required array arguments, options, and flags

MPLew-is:
Un-deprecates (but changes the semantics of) the initializers for an array value type without a default, forcing the user to specify at least one value from the command line.

The motivation behind this is to allow users of this framework to force a user to provide at least one argument, without having to provide custom error handling. An example of a command that uses this argument setup is `cp`, which has a variant for one or more `source_file` arguments:
`cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file ... target_directory`

This unfortunately does un-do the deprecations done In #193 and also is a source-breaking change in a similar vein as the one in #170 (changing the semantics of the case where no default is provided) and so will need to wait for the next minor release as per the documentation. I'm happy to sit on this and occasionally re-integrate `master` into it until you're ready for a new release, but figured I'd at least get it in for review.

### Checklist

### DIFF
--- a/Documentation/02 Arguments, Options, and Flags.md
+++ b/Documentation/02 Arguments, Options, and Flags.md
@@ -285,6 +285,46 @@ Verbosity level: 1
 Verbosity level: 4
 ```
 
+
+## Specifying default values
+
+You can specify default values for almost all supported argument, option, and flag types using normal property initialization syntax:
+
+```swift
+enum CustomFlag: String, EnumerableFlag {
+    case foo, bar, baz
+}
+
+struct Example: ParsableCommand {
+    @Flag
+    var booleanFlag = false
+
+    @Flag
+    var arrayFlag: [CustomFlag] = [.foo, .baz]
+
+    @Option
+    var singleOption = 0
+
+    @Option
+    var arrayOption = ["bar", "qux"]
+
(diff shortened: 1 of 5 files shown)


natecook1000:
@MPLew-is Thanks for this PR! This is a great addition, as zero-or-more and one-or-more are the primary expectations for array properties. Agreed that this will need to wait until 0.3.0. 

For documentation, I think we'll want to add a section to guide 2 about how defaults work in general; this can be an example there.

natecook1000:
@MPLew-is Ready to take this change — want to deal with the merge conflicts?

MPLew-is:
Sure! I’ll push a resolution for them tonight, so you should be able to merge tomorrow morning.

MPLew-is:
I've resolved the merge conflicts and extended this to `Flag` arrays (which I apparently missed the first time around), so this should be good to merge pending any review you have. I also took a first pass at updating the documentation, but it was a quick stab so may need some updating.

natecook1000:
@swift-ci Please test

natecook1000:
Tests and the doc additions look great, thank you!

natecook1000:
@swift-ci Please test

Issue #192, opened 2020-06-23
Using a command as its own subcommand crashes


MaxDesiatov:
Specifying a command type as its own subcommand in `CommandConfiguration` causes segmentation fault instead of a proper error.

**ArgumentParser version:** `0.1.0`
**Swift version:** 
```
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.5.0
```

### Checklist
- [x] If possible, I've reproduced the issue using the `master` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

```swift
import ArgumentParser

struct Install: ParsableCommand {
  static var configuration = CommandConfiguration(
    abstract: "Install new thingy.",
    subcommands: [Install.self]
  )
}
```

### Expected behavior
This maybe could fail to compile (which is probably hard to express on the type system level), or at least the recursion should be detected and reported as an error at run time.

### Actual behavior

Running an application with this code crashes with no actionable errors, only with a "segmentation fault" message.

PR #197, opened 2020-06-26
Add a precondition to prevent the parsing of a command that has itself as its subcommand

NicFontana:
This avoids the infinite recursion that causes a crash and shows the user a meaningful error message.

Fixes: https://github.com/apple/swift-argument-parser/issues/192

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -44,6 +44,13 @@ extension ParsableCommand {
   public mutating func run() throws {
     throw CleanExit.helpRequest(self)
   }
+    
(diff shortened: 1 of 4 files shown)


NicFontana:
Should I add an example that shows that the precondition catches the wrong ```CommandConfiguration``` and show the error message?

natecook1000:
Re: testing — we don't have a great way of testing crashes, unfortunately. You could add a new parser error case (capturing the duplicated command?), and throw that instead of using a precondition failure.

NicFontana:
Good point @natecook1000, I've changed the PR to fix the issue as you suggested and added a new test to show it's working.
Let me know if you do think that this still needs changes :)

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #201, opened 2020-06-29
Optimize Sequence.uniquified()

schlagelk:
Proposes a slight optimization of `uniquified ` added in #179.  Using `contains` on the temporary `sequence` variable leads to quadratic performance because `.contains` runs in O(n) time.  

By instead creating a temporary variable that is a dictionary, we can filter on `self` and use the dictionary's [`.updateValue`](https://developer.apple.com/documentation/swift/dictionary/3127179-updatevalue) member, passing in each element. This keeps performance near O(n) - the tradeoff being that this requires elements to conform to `Hashable` (vs just `Equatable`).

Not sure if another test is needed as all tests continue to pass, but happy to add one if this is smiled upon.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
@@ -11,7 +11,7 @@
 
 extension Sequence where Element: Hashable {
   func uniquified() -> [Iterator.Element] {
-    var seen: [Iterator.Element: Bool] = [:]
-    return self.filter { seen.updateValue(true, forKey: $0) == nil }
+    var seen = Set<Element>()
+    return self.filter { seen.insert($0).0 }
   }
 }
--- a/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-extension Sequence where Iterator.Element: Hashable {
+extension Sequence where Element: Hashable {
   func uniquified() -> [Iterator.Element] {
     var seen: [Iterator.Element: Bool] = [:]
     return self.filter { seen.updateValue(true, forKey: $0) == nil }
--- a/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/SequenceExtensions.swift
@@ -10,7 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 extension Sequence where Element: Hashable {
-  func uniquified() -> [Iterator.Element] {
+  func uniquified() -> [Element] {
     var seen = Set<Element>()
     return self.filter { seen.insert($0).0 }
   }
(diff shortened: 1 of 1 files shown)


Sources/ArgumentParser/Utilities/SequenceExtensions.swift
```diff
@@ -9,14 +9,9 @@
...
-    var sequence = Array<Element>()
-    for element in self {
-      if !sequence.contains(element) {
-        sequence.append(element)
-      }
-    }
-    return sequence
+extension Sequence where Iterator.Element: Hashable {
+  func uniquified() -> [Iterator.Element] {
+    var seen: [Iterator.Element: Bool] = [:]
```
natecook1000, line 14: Would you mind using a `Set<Element>` instead of a dictionary here?

schlagelk: @natecook1000 thanks!  The `updateValue` member used to make this possible is on the `Dictionary` type: https://developer.apple.com/documentation/swift/dictionary/3127179-updatevalue.  Is there an alternative provided by `Set`?

Sources/ArgumentParser/Utilities/SequenceExtensions.swift
```diff
@@ -9,14 +9,9 @@
...
-extension Sequence where Element: Equatable {
-  func uniquified() -> [Element] {
-    var sequence = Array<Element>()
-    for element in self {
-      if !sequence.contains(element) {
-        sequence.append(element)
-      }
-    }
-    return sequence
+extension Sequence where Iterator.Element: Hashable {
```
airspeedswift, line 12: Is there a reason to switch from `Element` to `Iterator.Element` in this patch?

schlagelk: @airspeedswift nothing specific and probably just an oversight on my part - happy to make the change back

Sources/ArgumentParser/Utilities/SequenceExtensions.swift
natecook1000: You can use [`insert(_:)`](https://developer.apple.com/documentation/swift/set/3128848-insert), which returns a tuple — the first element of the tuple is `true` if the element is inserted and `false` if it's already in the set.

schlagelk: @natecook1000 thats awesome! I made this change

Sources/ArgumentParser/Utilities/SequenceExtensions.swift
```diff
@@ -9,14 +9,9 @@
...
-  func uniquified() -> [Element] {
-    var sequence = Array<Element>()
-    for element in self {
-      if !sequence.contains(element) {
-        sequence.append(element)
-      }
-    }
-    return sequence
+extension Sequence where Element: Hashable {
+  func uniquified() -> [Iterator.Element] {
```
natecook1000, line 13: This one too, please!
```suggestion
  func uniquified() -> [Element] {
```

natecook1000:
@swift-ci Please test

PR #207, opened 2020-07-25
Allow property wrappers to be used without parentheses

### LINKED ISSUE #206: ParsableArguments type unexpectedly doesn't conform due to missing initialiser
This example from the source documentation of `Argument` doesn't compile:

```swift
  @Argument var name: String
  @Argument var greeting: String?
}
```

The error is `Type 'ParsableArgumentsValidationTests.Options' does not conform to protocol 'ParsableArguments'`, and the missing requirement is the default initialiser, `init()`.

Screen capture from Xcode showing the previously-described struct Options overlaid by a compiler error that reads 'Type ParsableArgumentsValidationTests.Options' does not conform to protocol 'ParsableArguments' and a Fix-it asking 'Do you want to add protocol stubs?'

The way to make this example compile, as found throughout the documentation, is to declare the `name` variable as follows:

```swift
@Argument() var name: String
```

This appears to convince Swift to synthesise both `init()` and `init(name: String, greeting: String?)` initialisers. I've read [SE-0258 Property Wrappers](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#detailed-design) and it's still unclear to me whether the proposal text offers an explanation for this difference in behaviour.

**ArgumentParser version:** `0.2.0`, `master`
**Swift version:**
```
Apple Swift version 5.3 (swiftlang-1200.0.22.4 clang-1200.0.25.1)
Target: x86_64-apple-darwin20.0.0
```

### Checklist

### Steps to Reproduce
To reproduce the behaviour, declare a `ParsableArguments` type using `@Argument` on one of its properties (as opposed to `@Argument()`).

### Expected behavior
The Property Wrappers proposal doesn't have a single example of the "default initialiser" syntax — i.e., with empty parenthesis after the property wrapper annotation — and it's unclear to me whether this is an intended distinction or not.

Given that the `Argument` type is initialisable without any parameters, I'd expect, or at least quite like, for Swift to be able to generate a no-argument initialiser in this case. I don't understand why adding empty parenthesis causes that to happen.

It's syntactically "cleaner" to be able to write `@Argument` instead of `@Argument()`. It feels natural to write it that way, especially when prototyping a new command and help text is not important.

### Actual behavior
See description of the issue at the top of this post.

sharplet:
Fixes #206, where unless the property wrapper types were used with empty parentheses, `ParsableArguments`-conforming types would not have a no-argument initializer synthesized and thus would unexpectedbly fail to
conform to `ParsableArguments`.

This is caused by [a limitation of the compiler's property wrapper support](https://bugs.swift.org/browse/SR-13295) where it fails to see initializers in extensions as candidates for property wrapper requirements.

By declaring a no-argument intializer in each type, but then marking it as unavailable, the no-parenthesis syntax now works, but selects the existing initializers in extensions anyway.

### Checklist

### DIFF
--- a/Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
+++ b/Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
@@ -167,6 +167,17 @@ final class ParsableArgumentsValidationTests: XCTestCase {
     var options: Options
   }
 
+  private struct L: ParsableArguments {
+    struct Options: ParsableArguments {
+      @Argument var items: [Int] = []
+    }
+
+    @Argument var foo: String
+    @Option var bar: String
+    @OptionGroup var options: Options
+    @Flag var flag: Bool
+  }
+
   func testPositionalArgumentsValidation() throws {
     XCTAssertNil(PositionalArgumentsValidator.validate(A.self))
     XCTAssertNil(PositionalArgumentsValidator.validate(F.self))
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -35,6 +35,20 @@ public struct Argument<Value>:
   public init(from decoder: Decoder) throws {
     try self.init(_decoder: decoder)
   }
+
+  /// This initializer works around a quirk of property wrappers, where the
+  /// compiler will not see no-argument initializers in extensions. Explicitly
+  /// marking this initializer unavailable means that when `Value` conforms to
+  /// `ExpressibleByArgument`, that overload will be selected instead.
+  ///
+  /// ```swift
+  /// @Argument() var foo: String // Syntax without this initializer
+  /// @Argument var foo: String   // Syntax with this initializer
+  /// ```
+  @available(*, unavailable, message: "A default value must be provided unless the value type conforms to ExpressibleByArgument.")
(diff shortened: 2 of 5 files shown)


Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
```diff
@@ -167,6 +167,17 @@ final class ParsableArgumentsValidationTests: XCTestCase {
     var options: Options
   }
 
+  private struct L: ParsableArguments {
```
natecook1000, line 170: Could you add a comment here that this is a compilation test validating that parentheses aren't required on the property wrappers? It looks unused otherwise.

natecook1000:
Great find, thanks @sharplet! 👏👏

natecook1000:
@swift-ci Please test

Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
natecook1000: I'll add the comment later, want to get this merged before to deal w/ conflicts. 👍

sharplet: Thanks! I might also follow up with a commit to update the documentation to use this style.

PR #209, opened 2020-07-26
Make arrays with default values use their elements' defaultValueDescription

### LINKED ISSUE #208: defaultValueDescription isn't used in arrays
`defaultValueDescription` isn't used if the argument/option is declared as an array of an `ExpressibleByArgument` type, which means the values presented in help aren't valid values when provided by the user.

**ArgumentParser version:** `0.2.0`.
**Swift version:** `Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)`

### Checklist

### Steps to Reproduce
Example:

```swift
public enum Resolution: Int, ExpressibleByArgument {
    case normal = 1
    case retina = 2
    case retinaHD = 3

    public init?(argument: String) {
        guard let int = Int(argument) else { return nil }

        self.init(rawValue: int)
    }

    public var defaultValueDescription: String {
        return String(rawValue)
    }
}
```

And then in the `ParsableCommand`:

```swift
@Option(
    name: [.customLong("resolution"), .customShort("r")],
    help: ArgumentHelp("Resolutions to generate.", valueName: "scale")
)
var resolutions: [File.Resolution] = [.normal, .retina]
```

### Expected behavior
I'd expect `command --help` to show something like

```
Resolutions to generate. (default: [1, 2])
```

### Actual behavior
`command --help` instead shows

```
Resolutions to generate. (default: [command.Resolution.normal, command.Resolution.retina])
```

Wevah:
Makes default values for `@Argument`s and `@Option`s whose values are arrays of `ExpressibleByArgument` use their elements' `defaultValueDescription` in the help text.

Fixes #208.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -366,7 +366,7 @@ extension Argument {
         initial: { origin, values in
           values.set(wrappedValue, forKey: key, inputOrigin: origin)
         })
-      arg.help.defaultValue = !wrappedValue.isEmpty ? "\(wrappedValue)" : nil
+      arg.help.defaultValue = !wrappedValue.isEmpty ? wrappedValue.defaultValueDescription : nil
       return ArgumentSet(alternatives: [arg])
     })
   }
--- a/Sources/ArgumentParser/Parsable Properties/Option.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Option.swift
@@ -498,7 +498,7 @@ extension Option {
       var arg = ArgumentDefinition(kind: kind, help: help, parsingStrategy: ArgumentDefinition.ParsingStrategy(parsingStrategy), update: .appendToArray(forType: Element.self, key: key), initial: { origin, values in
         values.set(wrappedValue, forKey: key, inputOrigin: origin)
       })
-      arg.help.defaultValue = !wrappedValue.isEmpty ? "\(wrappedValue)" : nil
+      arg.help.defaultValue = !wrappedValue.isEmpty ? wrappedValue.defaultValueDescription : nil
       return ArgumentSet(alternatives: [arg])
       })
   }
--- a/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
+++ b/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
@@ -66,3 +66,9 @@ extension Float: ExpressibleByArgument {}
 extension Double: ExpressibleByArgument {}
 
 extension Bool: ExpressibleByArgument {}
+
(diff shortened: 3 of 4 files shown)


Wevah:
One downside is that it still kind of makes it look like an option *takes* an array (that is, a string formatted as `"[a, b, ...]"`)…but I think it's still better than it is currently.

natecook1000:
This is great, @Wevah! I don't think the square brackets are necessary — it looks fine to me this way:

```
USAGE: p [-o <o> ...] [<remainder> ...]

ARGUMENTS:
  <remainder>             Help Message (default: large)

OPTIONS:
  -o <o>                  Help Message (default: small, medium)
```

What do you think?

Wevah:
Either way is fine to me. I went with brackets just to stay close to the old output, but looking at it again I think bracketless is probably better!

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
Looks like this fixture needs to be updated too: https://github.com/apple/swift-argument-parser/blob/35ceb59427d0cc836bb57dee0a3d4ea7ae9b6c05/Tests/ArgumentParserUnitTests/HelpGenerationTests.swift#L177

Wevah:
Will do.

natecook1000:
No need to duplicate, just take the brackets out of that test, too:

```
XCTAssertEqual failed: ("  --lucky <numbers>       Your lucky numbers. (default: 7, 14)") is not equal to ("  --lucky <numbers>       Your lucky numbers. (default: [7, 14])")
```

Wevah:
Oh, gotcha. I'd been filtering to just the test I was working on so didn't see that crop up. 🤦🏻‍♂️

natecook1000:
@swift-ci Please test

natecook1000:
#
@swift-ci Please test

Wevah:
I just noticed I didn't quite match style with the earlier `ExpressibleByArgument` extension: I left a `return` keyword in.

Wevah:
I *also* found whit might have been a better place to do this (inside `HelpGenerator`).

natecook1000:
> I also found whit might have been a better place to do this (inside HelpGenerator).

I'm not sure that's possible, since you don't have access to the `ExpressibleByArgument` conformance at that point. Feel free to open a PR if you find a way for that approach to work!

> I just noticed I didn't quite match style with the earlier ExpressibleByArgument extension: I left a return keyword in.

No worries! PR welcome if you'd like.

Wevah:
> > I also found whit might have been a better place to do this (inside HelpGenerator).
> 
> I'm not sure that's possible, since you don't have access to the `ExpressibleByArgument` conformance at that point. Feel free to open a PR if you find a way for that approach to work!

Ahh, good point.

PR #217, opened 2020-07-31
Update doc with Zsh completion filename format

miguelangel-dev:
Users do not familiarize with the shell-completions scripts in Zsh could be frustrated following [this guide](https://github.com/apple/swift-argument-parser/blob/280700d361c1b3af6e2345f5e24f67fa9450bec6/Documentation/07%20Completion%20Scripts.md#installing-zsh-completions). `oh-my-zsh` uses a specific filename format `_example` - I think it is useful to add this information to documentation together with an example.

### Checklist


natecook1000:
@swift-ci Please test

PR #223, opened 2020-08-05
Add an experimental customization point for the error label

natecook1000:
This supports SwiftPM, which uses "error:" instead of "Error:" to make its error messages match the compiler.


natecook1000:
@swift-ci Please test

PR #235, opened 2020-09-01
Flatten the ArgumentSet storage into a single array

natecook1000:
This removes the nesting inside the ArgumentSet data structure, which had semantic meaning in an earlier version. This flattening, plus a switch to using dictionary lookup instead of linear scanning, provides another performance boost.


natecook1000:
@swift-ci Please test

PR #238, opened 2020-09-14
Fix CMake build on Linux

gmittert:
- TestHelpers has a implicit dependency on Foundation through
  ArgumentParser, so it needs to link against Foundation
- The find package calls should be REQUIRED and CONFIG so they can take
  advantage of CMake configs

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,9 +22,9 @@ endif()
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
-find_package(dispatch QUIET)
-find_package(Foundation QUIET)
-find_package(XCTest QUIET)
+find_package(dispatch CONFIG)
+find_package(Foundation CONFIG)
+find_package(XCTest CONFIG)
 
 add_subdirectory(Sources)
 if(BUILD_EXAMPLES)
--- a/Sources/ArgumentParserTestHelpers/CMakeLists.txt
+++ b/Sources/ArgumentParserTestHelpers/CMakeLists.txt
@@ -5,4 +5,5 @@ set_target_properties(ArgumentParserTestHelpers PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
 target_link_libraries(ArgumentParserTestHelpers PUBLIC
   ArgumentParser
-  XCTest)
+  XCTest
+  Foundation)


gmittert:
@swift-ci please test platform

gmittert:
ping @natecook1000

natecook1000:
@swift-ci Please test

natecook1000:
@gmittert Thanks for the contribution! I don't think our CI testing exercises the CMake build path yet — can you tell me the configuration where you're seeing this go wrong, so that I can verify locally?

natecook1000:
@compnerd Would you mind giving this a review?

Sources/ArgumentParserTestHelpers/CMakeLists.txt
```diff
@@ -5,4 +5,4 @@ set_target_properties(ArgumentParserTestHelpers PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
 target_link_libraries(ArgumentParserTestHelpers PUBLIC
   ArgumentParser
-  XCTest)
+  XCTest Foundation)
```
compnerd, line 8: Nit: would be nice to do one entry per line

CMakeLists.txt
```diff
@@ -22,9 +22,9 @@ endif()
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
-find_package(dispatch QUIET)
-find_package(Foundation QUIET)
-find_package(XCTest QUIET)
+find_package(dispatch QUIET CONFIG REQUIRED)
+find_package(Foundation QUIET CONFIG REQUIRED)
+find_package(XCTest QUIET CONFIG REQUIRED)
```
compnerd, line 27: I'm worried about this part of the change.  With this it is impossible to actually build with a prebuilt toolchain image.  The `CONFIG` is fine, the `REQUIRED` is the piece that worries me.

gmittert: I'll remove both the QUIET and the REQUIRED then. My main concern was that it was silently failing and was a pain to debug why.

compnerd: The thing is that it isn't an absolutely required thing - so the non `QUIET` form could be confusing too.  But I would be fine with it being non-quiet as long as it is not required.

gmittert:
@natecook1000 The precise invocation I use is a little complex since we have a lot of infrastructure to support cross compiling so I'm unfortunately not able to give you a command to run, but it's something like:
```
cmake -G Ninja
-DCMAKE_MAKE_PROGRAM="/usr/bin/ninja"
-DCMAKE_BUILD_TYPE="Release"
-DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_INSTALL_PREFIX=
-DTOOLCHAIN_SOURCE_DIR=/data/users/gwenm/toolchain
-DTOOLCHAIN_BUILD_DIR=/data/users/gwenm/toolchain/build/Release
-DCMAKE_TRY_COMPILE_PLATFORM_VARIABLES=TOOLCHAIN_SOURCE_DIR
-DCMAKE_TOOLCHAIN_FILE="/path/to/platform/specific/toolchain/file"
-DUSE_CMAKE_INSTALL=YES
-Ddispatch_DIR=/path/to/swift-corelibs-libdispatch/cmake/modules
-DFoundation_DIR=/path/to/swift-corelibs-foundation/cmake/modules
-DXCTest_DIR=/path/to/swift-corelibs-xctest/cmake/modules
-C "/path/to/platform/specific/cache/vars/linux-x86_64.cmake"
/data/users/gwenm/toolchain_dev/external/swift-argument-parser
```
But really the important part is configuring via setting Xyz_DIR to locally build projects instead of building against a prebuilt toolchain.

compnerd:
@gmittert - most of that is specific to the setup that you are building in.  (e.g. the caches).  I think that it is probably easier to explain with the CI I have setup on Azure (https://github.com/compnerd/swift-build/blob/master/swift-argument-parser.yml)

```
cmake ^
      -B $(Build.BinariesDirectory)/swift-argument-parser ^
      -D BUILD_SHARED_LIBS=YES ^
      -D BUILD_TESTING=NO ^
      -D CMAKE_BUILD_TYPE=Release ^
      -D CMAKE_Swift_COMPILER=S:\b\1\bin\swiftc.exe ^
      -G Ninja ^
      -S $(Build.SourcesDirectory)/swift-argument-parser
```

This is Windows specific, but very easy to generalize to other platforms.  The important thing is to use a swift compiler from a swift build tree rather than than from an installed image.

gmittert:
Alright, updated the line spacing and removed REQUIRED. @compnerd exactly, my particular set up isn't particularly optimized for external reproducibility, but the important part is building against locally built dependencies a la the Windows build.

gmittert:
@swift-ci please test platform

compnerd (approved):
LGTM

natecook1000:
Thanks to you both!

@swift-ci Please test

PR #239, opened 2020-09-16
Store mutations made by validate()

elliottwilliams:
Since #104, mutations made to a command struct by its `validate()` function are lost when using `parseAsRoot(_:)`. This is because validations are run on a mutable copy of the parsed command, and that copy is never stored.

My teammates and I use `validate()` as a place to enforce invariants. For example:

```swift
struct Foo: ParsableCommand {
    @Flag()
    var verbose = false

    @Flag()
    var dryRun = false

    mutating func validate() {
        if dryRun {
            verbose = true    // --dry-run implies --verbose
        }
    }

    func run() {
        if verbose { ... }
    }
}
```

While this isn't _technically_ a validation step, it's been useful to separate post-processing of the arguments like this from the business logic of `run()`.

I couldn't find anything indicating that #104 intended to change this behavior, so I'm assuming it was a regression. Here's a fix!

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -157,6 +157,9 @@ extension CommandParser {
       // after decoding a command, make sure to validate it
       do {
         try parsedCommand.validate()
+        var lastArgument = decodedArguments.removeLast()
+        lastArgument.value = parsedCommand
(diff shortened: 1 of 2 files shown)


natecook1000:
@elliottwilliams Thanks for identifying and fixing this! 👏

natecook1000:
@swift-ci Please test

Xl5843003:
****

PR #245, opened 2020-10-03
add support to call custom subcommand provided by user

masasam-shi:
### Description
Some major commands, like swift and cargo, have a feature which calls `swift-sub` when entering `swift sub`. This PR adds support to this.

### Detailed Design
This PR doesn't add new APIs. 
A simple case is here (main is built with swift-argument-parser):

```
$ ls
main main-sub
$ ./main
main is called
$ ./main-sub
sub is called
$ ./main sub
sub is called
```

If `main` needs arguments, custom subcommands behave like swift-argument-parser's subcommand. I'm questioning this behavior, but I've decided to adapt it to the current behavior. 

```
$ ls
main main-custom
$ ./main -h
USAGE: main <text> <subcommand>

ARGUMENTS:
  <text>

OPTIONS:

SUBCOMMANDS:
  sub

  See 'main help <subcommand>' for detailed help.
$
$ ./main sub
main command is called; text: sub
$ ./main file sub
sub command is called
$ ./main sub file 
Error: Unexpected argument 'file'
Usage: main <text> <subcommand>
  See 'main --help' for more information.
$
$ ./main-custom
custom command is called
$ ./main custom
main command is called; text: custom
$ ./main file custom
custom command is called
$ ./main custom file
Error: Unexpected argument 'file'
Usage: main <text> <subcommand>
  See 'main --help' for more information.
```

### Documentation Plan
I don't know whether I should add documentation or not.

### Test Plan
`swift test` tests in `/private/temp` so I have no idea to test this feature. 

### Source Impact
Nothing

### Alternative Design
- Add a flag to `CommandConfiguration` whether allows sub command provided by users.

### Note
`execv` is always failed when I execute text executable file. It is no problem with a binary file.

### Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,11 +19,19 @@ var package = Package(
             name: "ArgumentParser",
             targets: ["ArgumentParser"]),
     ],
-    dependencies: [],
+    dependencies: [
+      .package(
+        url: "https://github.com/apple/swift-tools-support-core",
+        from: "0.1.10")
+    ],
     targets: [
         .target(
             name: "ArgumentParser",
-            dependencies: []),
+            dependencies: [
+                .product(
+                    name: "SwiftToolsSupport",
+                    package: "swift-tools-support-core")
+            ]),
         .target(
             name: "ArgumentParserTestHelpers",
             dependencies: ["ArgumentParser"]),
--- a/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
+++ b/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
@@ -47,6 +47,13 @@ public struct CommandConfiguration {
   /// Flag names to be used for help.
   public var helpNames: NameSpecification
   
+  public enum AllowingExternalCommands {
+    case none, all
+    case list([String])
+  }
+  /// Allow that call `main-sub` when user enters `main sub`
+  public var allowingExternalCommands: AllowingExternalCommands
+
   /// Creates the configuration for a command.
   ///
   /// - Parameters:
@@ -66,6 +73,7 @@ public struct CommandConfiguration {
   ///     is given.
   ///   - helpNames: The flag names to use for requesting help, simulating
   ///     a Boolean property named `help`.
+  ///   - allowingExternalCommands: Allow that call `main-sub` when user enters `main sub`
   public init(
     commandName: String? = nil,
     abstract: String = "",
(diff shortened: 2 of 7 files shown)


natecook1000:
Hi @masasam-shi — thanks for this PR! While it would be great to be able to support this kind of aggregate tool, I'd like to spend some time thinking about the design. Re this particular approach, I see a few issues:

1. This should be an opt-in feature, not one that every tool has by default. I could see being able to choose whether to specify an approved list of subcommands or to let it use the filesystem.
2. The problem you identify where arguments and subcommands can conflict is a big one. There should be clear, predictable behavior — perhaps prohibiting positional arguments on commands that have this kind of subcommand, or at least matching the behavior of subcommands explicitly defined in the same tool.
3. Let's make this work without the added dependency.

It would help to look at how the `swift` and `git` tools operate — with `swift`, any arguments between `swift` and the subcommand name confuses it, while `git` can handle some flags and options, but no positional arguments. Are there other tools that can help us establish the "best practice" for the behavior here?

masasam-shi:
I call subcommand implemented using swift-argument-parser the standard subcommand. 
And I’ll rename CustomCommand to ExternalSubcommand.

1. I’ll add `AllowingExternalSubcommand` and property which set it to CommandConfiguration. It is enum and has `all`, `none`, `list(String)`.

2. Current implementation behave like standard subcommands. Standard subcommands can get arguments of the main command by adding `@OptionGroup var main: Main`, but current external command can’t. IMO, current standard subcommand’s behavior is subtleness. 

```
$ ./main -h
USAGE: main <text> <subcommand>

ARGUMENTS:
  <text>

OPTIONS:

SUBCOMMANDS:
  sub

  See 'main help <subcommand>' for detailed help.
$ ./main help sub
sub command is called, text: help
$ ./main arg help sub
USAGE: main sub <text>

ARGUMENTS:
  <text>

OPTIONS:
$ ./main sub arg
Error: Unexpected argument 'arg'
Usage: main <text> <subcommand>
  See 'main --help' for more information.
$ ./main arg sub
sub command is called, text: arg
```

If I entered `main help sub` or `main sub arg` as help advised it does unexpected behavior or it failed. I don’t know a command which needs argument and has this system so I can’t assertion, but `main arg sub` syntax is strange basically.

3. I though too. I will.

`swift` prefers custom commands. If there is a swift file called `custom` and a command `swift-custom`, then `swift-custom` will be executed. One Japanese article introduced an implementation that looks for external subcommands first, even if they have arguments. I think it is best.

PR #247, opened 2020-10-06
Add localisation support for strings

bitigchi:
### Checklist


mpapple-swift:
Providing the ability to localize command line tools/apps is a great idea and I don't wish to dampen your enthusiasm for this effort.  I do, however, wish to pose an open question for discussion: .lproj resources are unique to Apple/Darwin platforms - is this an appropriate design for a library that is cross-platform in nature?  Of the leading (non-Swift) libraries I surveyed, none of them offers localization capabilities, so leveraging the Apple/Darwin approach is as good as any.  But it is a fair question to ascertain if there are any reservations from developers on Linux or Windows?  Otherwise, I'm +1 on adding this capability absent reservations about the design/implementation.

bitigchi:
I personally think `.lproj` files are fine. There are already different approaches to this in other platforms, like how `gettext` handles stuff. However, it would be neat to have the framework automatically handle the localisations in bundle, like it does with regular AppKit/UIKit/SwiftUI apps, without having to manually locate `.lproj` files.

PR #250, opened 2020-10-21
[Windows] use _dupenv_s instead of getenv

damuellen:
getenv generates a warning on windows.

### DIFF
--- a/Sources/ArgumentParser/Completions/CompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/CompletionsGenerator.swift
@@ -43,8 +43,16 @@ public struct CompletionShell: RawRepresentable, Hashable, CaseIterable {
   /// Returns an instance representing the current shell, if recognized.
   public static func autodetected() -> CompletionShell? {
     // FIXME: This retrieves the user's preferred shell, not necessarily the one currently in use.
+#if os(Windows)
+    var buffer: UnsafeMutablePointer<CChar>?
+    var length: Int = 0
+    guard _dupenv_s(&buffer, &length, "SHELL") != 0, let shellVar = buffer else { return nil }
+    let shellParts = String(cString: shellVar).split(separator: "/")
+    free(buffer)
+#else
     guard let shellVar = getenv("SHELL") else { return nil }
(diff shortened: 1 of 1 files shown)


natecook1000:
Thanks, @damuellen! Are you able to test this with one of the supported shells under Windows? I know we don't have automated tests for this, but it would be good to know the status.

natecook1000:
cc @compnerd

compnerd:
Honestly, I'm not sure if it's too important to test (beyond compiling).  The windows environment doesn't have a SHELL environment variable, your shell is either cmd or powershell.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
```diff
@@ -43,8 +43,16 @@ public struct CompletionShell: RawRepresentable, Hashable, CaseIterable {
   /// Returns an instance representing the current shell, if recognized.
   public static func autodetected() -> CompletionShell? {
     // FIXME: This retrieves the user's preferred shell, not necessarily the one currently in use.
+#if os(Windows)
+    var buffer: UnsafeMutablePointer<CChar>?
+    var length: Int = 0
+    guard _dupenv_s(&buffer, &length, "SHELL") != 0, let shellVar = buffer else { return nil }
+    let shellParts = String(cString: shellVar).split(separator: "/")
+    free(buffer)
```
compnerd, line 51: I would swap the two things, doing a `defer { free(buffer) }` and then just returning the processed value.  However, the separator is unlikely to be `/` and more likely to be `\` if there was such a variable, though it could be either.

natecook1000:
> The windows environment doesn't have a SHELL environment variable, your shell is either cmd or powershell.

🤔 In that case, given that `bash`, `fish`, and `zsh` are the only three choices here, should `CompletionShell.autodetected()` just return `nil` here?

compnerd:
Yeah, returning nil for the time being sounds reasonable.  For powershell, it may be possible to auto complete, but I don't know about cmd.

damuellen:
I will close this pr, and create a new own which simply return nil on windows.

PR #258, opened 2020-12-16
Render version in help output

jpsim:
I believe this is generally useful information to have, if it's been specified.

This was requested by a SwiftLint user who's been generous enough to try out a version of the tool that's integrated this library: https://github.com/realm/SwiftLint/issues/1088#issuecomment-745738042

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ package updates, you can specify your package dependency using
 
 ## [Unreleased]
 
-*No changes yet.*
+- The version string is now included in the help output if specified.
 
 ---
 
--- a/Documentation/04 Customizing Help.md
+++ b/Documentation/04 Customizing Help.md
@@ -74,7 +74,7 @@ OPTIONS:
 
 ## Customizing Help for Commands
 
-In addition to configuring the command name and subcommands, as described in [Command and Subcommands](03%20Commands%20and%20Subcommands.md), you can also configure a command's help text by providing an abstract and discussion.
+In addition to configuring the command name and subcommands, as described in [Command and Subcommands](03%20Commands%20and%20Subcommands.md), you can also configure a command's help text by providing an abstract, discussion or version number.
 
 ```swift
 struct Repeat: ParsableCommand {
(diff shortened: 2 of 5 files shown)


jpsim:
I'm open to suggestions for where I could reference this in the documentation, if helpful. Given that this doesn't change any APIs or integrations, only the contents of the output, I'm not sure it's important to call out in the documentation.

mpapple-swift:
Not speaking for Nate, but formal adoption of a version number for display would be nice.  Perhaps adding a blurb to [Customizing Help](https://github.com/apple/swift-argument-parser/blob/main/Documentation/04%20Customizing%20Help.md) would be good.

jpsim:
I think I found a natural home for it in that Customizing Help doc as you suggested, although I do think there's an opportunity for calling out other benefits of defining a version in a command's configuration as a standalone section in the docs, like how a `--version` argument is automatically generated.

mpapple-swift:
> I think I found a natural home for it in that Customizing Help doc as you suggested, although I do think there's an opportunity for calling out other benefits of defining a version in a command's configuration as a standalone section in the docs, like how a `--version` argument is automatically generated.

Agreed, but I'll defer to @natecook1000 to identify where it would be best to break out detail in the docs.

natecook1000:
Thanks for opening this PR, @jpsim! While I like the idea of more visible version information, we're trying to hew to CLI conventions as much as possible, such as they are. I haven't been able to find any examples of existing tools that include the version along with the help — if anything, help screens are crowded enough with just help text. Do you have any such examples?

Great to see the adoption in SwiftLint! I've opened #259 to track the usability problem flagged in that other issue.

jpsim:
> Do you have any such examples?

I don't have any examples of existing tools that do this. This idea came from @mr-fixit who seemed to have some expectation to see the version number in the general help output, so perhaps they know of existing tools that do this?

mr-fixit:
Well, I found a couple, but it's not common.  

```
$ man --help
man, version 1.6g...

$ bash -help
GNU bash, version 3.2.57(1)-release-(x86_64-apple-darwin20)
...

natecook1000:
Alright — given that, I think we already have enough support for adding the version in the abstract and overview. Thanks again for opening this!

Issue #269, opened 2021-02-09
MessageInfo uses CustomNSError's errorCode as the process exit status


lorentey:
This package uses the `errorCode` property of `CustomNSError` errors as the process exit status. 

```swift
      // MessageInfo.swift:109
      case let error as CustomNSError:
         self = .other(message: error.localizedDescription, exitCode: Int32(error.errorCode))
```

The purpose of `errorCode` is merely to differentiate error cases within the same error domain; it is not appropriate to use it as an exit status. The process exit status has a limited range (matching the range of `Int8` or `Int32` depending on the syscall used to retrieve it); additional bits get silently truncated. As a result, when the classic `wait`/`waitpid` syscalls are used, `ArgumentParser` processes may appear to exit with a zero status (EXIT_SUCCESS) even if they failed.

Additionally, an `errorCode` that doesn't fit in an `Int32` leads to a fatal runtime error.

The broken behavior was introduced in #244. One easy way to fix this is to revert that change.

One way to correctly implement #243 is to define a custom mix-in error protocol using an error property that is kept distinct from `errorCode`.

**ArgumentParser version:** `0.3.2`
**Swift version:** Latest stable

### Checklist
- [X] If possible, I've reproduced the issue using the `main` branch of this package
- [X] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

```swift
import Foundation
import ArgumentParser

enum MyError: Int, CustomNSError {
  case foo = 256
  case bar = 10_000_000_000

  static var errorDomain: String { "MyError" }
  var errorCode: Int { self.rawValue }
  var errorUserInfo: [String : Any] { [:] }
}

struct Command: ParsableCommand {
  @Flag
  var trap: Bool = false

  func run() throws {
    throw trap ? MyError.bar : MyError.foo
  }
}

Command.main()
```

### Expected behavior

I expect `ArgumentParser.main` to exit normally with a non-zero status whenever `run` throws an error it doesn't otherwise3 handle.

On the bash/fish/zsh prompt, I expect to see:

```
$ swift run test
$ echo $?
1   # Or some other non-zero value

$ swift run test --trap
$ echo $?
1   # Or some other non-zero value
```

### Actual behavior

Depending on the `errorCode` value, the subprocess appears to exit with a zero status, or exits with a signal instead.

With the bash/zsh installations on default macOS, and fish installed from homebrew, I get:

```
$ swift run test
$ echo $?
0
[...]

PR #272, opened 2021-02-11
Beautify `NSError` cases when thrown 💄 

### LINKED ISSUE #270: NSErrors aren't printed nicely when caught
NSError instances thrown from Objective-C aren't handled nicely by `MessageInfo`.

Because protocol conformances don't apply over bridging, `let e as CustomNSError` in an exception handler does not match NSError instances, even if the NSError happens to be bridgeable to `CocoaError` (or another domain-specific error type) that does conform to CustomNSError.

Note: In Foundation as of 2020, `NSError` and `CFError` conform to none of `CustomNSError`, `LocalizedError`, or any other error protocol other than `Error` itself. To handle NSError instances, `MessageInfo.init` needs to explicitly add a `case let e as NSError` clause.

**ArgumentParser version:** `0.3.2`
**Swift version:** 5.4

### Checklist
- [X] If possible, I've reproduced the issue using the `main` branch of this package
- [X] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

```swift
import Foundation
import ArgumentParser

struct Command: ParsableCommand {
  func run() throws {
    let url = URL(fileURLWithPath: "/This file does not exist")
    try print(NSData(contentsOf: url, options: .mappedIfSafe))
  }
}

Command.main()
```

### Expected behavior

```
$ swift run test
Error: The file “This file does not exist” couldn’t be opened because there is no such file.
```

### Actual behavior

```
$ swift run test 
Error: Error Domain=NSCocoaErrorDomain Code=260 "The file “This file does not exist” couldn’t be opened \
because there is no such file." UserInfo={NSFilePath=/This file does not exist, \
NSUnderlyingError=0x7fd669706a40 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

schlagelk:
Ensures the `NSError` case is handled in `MessageInfo` and attempts to resolve #270 

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Usage/MessageInfo.swift
+++ b/Sources/ArgumentParser/Usage/MessageInfo.swift
@@ -73,17 +73,6 @@ enum MessageInfo {
       )
       return
 
-    case let e as NSError:
-      // Send NSError back through the CommandError path
-      self.init(
-        error: CommandError(
-          commandStack: [type.asCommand],
-          parserError: .userValidationError(e)
-        ),
-        type: type
-      )
-      return
-
     default:
       commandStack = [type.asCommand]
       // if the error wasn't one of our two Error types, wrap it as a userValidationError
@@ -121,10 +110,12 @@ enum MessageInfo {
         self = .other(message: error.localizedDescription, exitCode: Int32(error.errorCode))
       case let error as LocalizedError where error.errorDescription != nil:
         self = .other(message: error.errorDescription!, exitCode: EXIT_FAILURE)
-      case let error as NSError:
-        self = .other(message: error.localizedDescription, exitCode: EXIT_FAILURE)
       default:
-        self = .other(message: String(describing: error), exitCode: EXIT_FAILURE)
+        if Swift.type(of: error) is NSError.Type {
+          self = .other(message: error.localizedDescription, exitCode: EXIT_FAILURE)
+        } else {
+          self = .other(message: String(describing: error), exitCode: EXIT_FAILURE)
+        }
       }
     } else if let parserError = parserError {
       let usage: String = {
--- a/Sources/ArgumentParser/Usage/MessageInfo.swift
+++ b/Sources/ArgumentParser/Usage/MessageInfo.swift
@@ -73,6 +73,17 @@ enum MessageInfo {
       )
       return
 
+    case let e as NSError:
+      // Send NSError back through the CommandError path
+      self.init(
+        error: CommandError(
+          commandStack: [type.asCommand],
+          parserError: .userValidationError(e)
+        ),
+        type: type
+      )
+      return
+
     default:
       commandStack = [type.asCommand]
       // if the error wasn't one of our two Error types, wrap it as a userValidationError
@@ -110,6 +121,8 @@ enum MessageInfo {
         self = .other(message: error.localizedDescription, exitCode: Int32(error.errorCode))
       case let error as LocalizedError where error.errorDescription != nil:
         self = .other(message: error.errorDescription!, exitCode: EXIT_FAILURE)
+      case let error as NSError:
+        self = .other(message: error.localizedDescription, exitCode: Int32(error.code))
       default:
         self = .other(message: String(describing: error), exitCode: EXIT_FAILURE)
       }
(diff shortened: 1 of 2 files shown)


Sources/ArgumentParser/Usage/MessageInfo.swift
```diff
@@ -110,6 +121,8 @@ enum MessageInfo {
         self = .other(message: error.localizedDescription, exitCode: Int32(error.errorCode))
       case let error as LocalizedError where error.errorDescription != nil:
         self = .other(message: error.errorDescription!, exitCode: EXIT_FAILURE)
+      case let error as NSError:
+        self = .other(message: error.localizedDescription, exitCode: Int32(error.code))
```
lorentey, line 125: Please do not use the error code as the exit code; see #269. For now, `EXIT_FAILURE` is the best option.

schlagelk: @lorentey roger that - done on 37fb5d99ecc3e7d0fa12937514616bd3389caa6e

lorentey (approved):
LGTM! 👍

natecook1000 (approved):
Looks great, @schlagelk — thanks for your contribution! 👏🏻👏🏻

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Usage/MessageInfo.swift
```diff
@@ -73,6 +73,17 @@ enum MessageInfo {
...
+      // Send NSError back through the CommandError path
+      self.init(
+        error: CommandError(
+          commandStack: [type.asCommand],
+          parserError: .userValidationError(e)
+        ),
+        type: type
+      )
+      return
+
```
natecook1000, line 86: It turns out this bit isn't necessary! The default case handles general errors just fine.

Sources/ArgumentParser/Usage/MessageInfo.swift
```diff
@@ -110,6 +121,8 @@ enum MessageInfo {
         self = .other(message: error.localizedDescription, exitCode: Int32(error.errorCode))
       case let error as LocalizedError where error.errorDescription != nil:
         self = .other(message: error.errorDescription!, exitCode: EXIT_FAILURE)
+      case let error as NSError:
+        self = .other(message: error.localizedDescription, exitCode: EXIT_FAILURE)
       default:
         self = .other(message: String(describing: error), exitCode: EXIT_FAILURE)
```
natecook1000, line 127: Here we have a problem — on macOS, the `error as NSError` cast always succeeds, which means we lose the customization for non-`NSError` errors that conform to `CustomStringConvertible`. Since `NSError`, in turn, conforms to `CustomStringConvertible`, there's no way to distinguish them by casting alone.

This is an unfortunate but successful way of distinguishing actual `NSError` errors from plain Swift errors:
```suggestion
      default:
        if Swift.type(of: error) is NSError.Type, let error = error as? NSError {
          self = .other(message: error.localizedDescription, exitCode: EXIT_FAILURE)
        } else {
          self = .other(message: String(describing: error), exitCode: EXIT_FAILURE)
        }
```

natecook1000 (commented):
Thanks so much for tackling this, @schlagelk! We need a bit of a different approach for identifying `NSError`s — notes below.

Sources/ArgumentParser/Usage/MessageInfo.swift
schlagelk: @natecook1000 interesting, and as always thanks for the amazing insight!  I'm wondering if the conditional cast to `NSError` needs to happen here for any reason then, or if a13105c would suffice?

natecook1000: Does that compile? I don’t think `localizedDescription` will be available on `error` until the cast succeeds. Also note that this bridging behavior isn’t consistent between macOS and Linux!

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Usage/MessageInfo.swift
schlagelk: It does compile!  In fact Xcode gave me a warning about the cast always succeeding. I'm just making sure I didn't miss anything, but I'm happy to add it in if it's necessary.

natecook1000: TIL! Let’s run with it :shipit:

PR #284, opened 2021-03-05
Fix zsh dashes in command names

### LINKED ISSUE #277: Dashes in command names break completion scripts
Dashes aren't valid variable identifier characters for Zsh (~and maybe Bash~ just Zsh). When building a completion script for a command that includes one or more dashes in the command name, the completion script prints errors instead of providing completions.

**Swift version:** Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)

### Checklist

### Steps to Reproduce
Create a CLI tool with a dash in the command name, then generate and install a Zsh completion script for the command.

Command:
```swift
struct ExampleCommand: ParsableCommand {
  @Option var name: String
  @Flag var verbose = false
}
```

Script:
```
#compdef example-command
local context state state_descr line
_example-command_commandname=$words[1]
typeset -A opt_args

_example-command() {
    integer ret=1
    local -a args
    args+=(
        '--name:name:'
        '--verbose'
        '(-h --help)'{-h,--help}'[Print help information.]'
    )
    _arguments -w -s -S $args[@] && ret=0

    return ret
}


_custom_completion() {
    local completions=("${(@f)$($*)}")
    _describe '' completions
}

_example-command
```

### Expected behavior
Typing `example-command` and then `TAB` should offer completions.

### Actual behavior
The shell prints `_example-command:3: command not found: _example-command_commandname=example-command`.

adellibovi:
Hi, this PR fixes #277

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
@@ -17,7 +17,7 @@ struct ZshCompletionsGenerator {
     return """
     #compdef \(type._commandName)
     local context state state_descr line
-    _\(type._commandName)_commandname=$words[1]
+    _\(type._commandName.zshEscapingCommandName())_commandname=$words[1]
     typeset -A opt_args
 
     \(generateCompletionFunction([type]))
@@ -121,6 +121,10 @@ extension String {
 
   fileprivate func zshEscaped() -> String {
     self.zshEscapingSingleQuotes().zshEscapingMetacharacters()
+  }
+  
+  fileprivate func zshEscapingCommandName() -> String {
+    self.replacingOccurrences(of: "-", with: "_")
   }
 }
 
--- a/Tests/ArgumentParserUnitTests/CompletionScriptTests.swift
+++ b/Tests/ArgumentParserUnitTests/CompletionScriptTests.swift
@@ -125,13 +125,13 @@ extension CompletionScriptTests {
 }
 
 extension CompletionScriptTests {
-  struct Escaped: ParsableCommand {
+  struct EscapedCommand: ParsableCommand {
     @Option(help: #"Escaped chars: '[]\."#)
     var one: String
   }
 
   func testEscaped_Zsh() throws {
-    XCTAssertEqual(zshEscapedCompletion, Escaped.completionScript(for: .zsh))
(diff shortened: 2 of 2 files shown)


Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
```diff
@@ -12,12 +12,13 @@
 struct ZshCompletionsGenerator {
   /// Generates a Zsh completion script for the given command.
   static func generateCompletionScript(_ type: ParsableCommand.Type) -> String {
-    let initialFunctionName = [type].completionFunctionName()
+    let initialFunctionName = [type].completionFunctionName().zshEscapingCommandName()
+    let commandName = type._commandName.zshEscapingCommandName()
 
     return """
-    #compdef \(type._commandName)
+    #compdef \(commandName)
```
natecook1000, line 19: I don't think this part can change — IIUC the `#compdef` tells Zsh what command the completions are for, so they need to match the executable.

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
```diff
@@ -12,12 +12,13 @@
...
     local context state state_descr line
-    _\(type._commandName)_commandname=$words[1]
+    _\(commandName)_commandname=$words[1]
```
natecook1000, line 21: As far as I understand it, this is the only part that needs to change, since dashes _are_ allowed in the function names, just not as a variable name. Can you still reproduce the issue if you only change this part here? As an extra benefit, this smaller change wouldn't cause problems for people who declare both `sub-command` and `sub_command` subcommands (which is a bad idea, but it will surely happen eventually).

natecook1000 (commented):
Thanks for tackling this, @adellibovi! 👏 A couple notes below.

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
adellibovi: You are right,  this is the only required changes. I made the change and tried locally :) Thanks for catching it @natecook1000. PR updated!

natecook1000:
@swift-ci Please test

natecook1000 (approved):
LGTM!

PR #286, opened 2021-03-06
Don't error when decoding optional unparsed values

### LINKED ISSUE #285: 0.4.0 seems to think there's an argument var that doesn't actually exist
Upon updating ArgumentParser to 0.4.0 (from 0.3.2), my project fails with the error "Error: No value set for non-argument var InputKey(rawValue: "products"). Replace with a static variable, or let constant."

`products` is a codable key on the `Config` struct; verified by changing the name. Removing the loading of that key also causes it to report the same error but with the next key down.

I'm using a custom struct called Config, and ArgumentParser seems to be trying to validate that object even though it doesn't conform to any of the ArgumentParser protocols, only `Decodable` (it loads a configuration from JSON in `run()`).

Seems to be caused the changes e99a8ef488533e3b331535902843230d2566d4ed.

Is there a way revert to the old behavior making it `static`?

**ArgumentParser version:** 0.4.0
**Swift version:** Apple Swift version 5.3.2

### Checklist

### Steps to Reproduce
Build the project in question with ArgumentParser 0.4.0.

### Expected behavior
Behavior should be the same as older versions.

### Actual behavior
The above error is reported.

natecook1000:
We're newly able to decode actual optional types due to allowing unparsed variable properties. "Normal" optional values are still wrapped in non-optional property wrappers, so `ArgumentDecoder` didn't need to handle optional values until now. Fixes #285.

### Checklist


natecook1000:
@swift-ci Please test

PR #288, opened 2021-03-09
Add edge case to remove duplicate help arguments

### LINKED ISSUE #259: Ignore `--help` flag when specified on a `cmd help subcommand` command
As [noted here](https://github.com/realm/SwiftLint/issues/1088#issuecomment-745738042), adding the `--help` flag subverts the user's intent if they're trying to get help about a subcommand.

**Swift version:** any

### Checklist

### Steps to Reproduce
1. Build the project with `swift build`
2. Call the `math` command both the `help` subcommand and a `--help` flag, e.g. `.build/debug/math help stats --help`

### Expected behavior
It should print the help text for the `stats` subcommand.

### Actual behavior
It prints the help text for the top-level `math` command.

Mark-McCracken:
Resolves #259 

Literally ignores any --help flag, if the first argument is help.

I did spend a while digging into the behaviour using the debugger, but this code is fairly complex, and adding an edge case at this low level would make it harder to understand I think. This ultimately is a little hacky, but I think better than making that low level complexity more complex.

FYI, it was happening somewhere in this function: https://github.com/apple/swift-argument-parser/blob/831ed5e860a70e745bc1337830af4786b2576881/Sources/ArgumentParser/Parsing/CommandParser.swift#L155
If you add a breakpoint on the next line, run the program, continue program execution 2 times, then the `firstUnused` property jumps from 1 to 3, skipping the unused commands, causing our bug.

### Checklist
- [questionable X] I've followed the code style of the rest of the project
- [not necessary] I've updated the documentation if necessary

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -60,9 +60,19 @@ extension ParsableCommand {
     _ arguments: [String]? = nil
   ) throws -> ParsableCommand {
     var parser = CommandParser(self)
-    let arguments = arguments ?? Array(CommandLine.arguments.dropFirst())
+    var arguments = arguments ?? Array(CommandLine.arguments.dropFirst())
+    arguments = checkForDuplicateHelpCommands(arguments)
     return try parser.parse(arguments: arguments).get()
   }
+    
+    public static func checkForDuplicateHelpCommands(_ arguments: [String]) -> [String] {
+        guard arguments.count >= 2 else { return arguments }
+        let nonOptionalArguments = arguments.compactMap { $0 }
+        if nonOptionalArguments.first == "help" && nonOptionalArguments.contains("--help") {
+            return nonOptionalArguments.filter { $0 != "--help" }
+        }
+        return arguments
+    }
   
   /// Returns the text of the help screen for the given subcommand of this
   /// command.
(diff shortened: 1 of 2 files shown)


natecook1000:
Thanks for taking this on, @Mark-McCracken! We need a bit of a different approach than what you've taken here, since commands are free to define different names for the `--help` flag (using the `helpNames` parameter on `CommandConfiguration`) and not every command has a `help` subcommand.

Instead, I think we could look at making changes to the [`HelpCommand` type](https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Usage/HelpCommand.swift), which is an internal command type that generates help for the selected subcommand. When someone executes a command like `math help stats average`, `HelpCommand` is the one that gets matched and created, with `"stats"` and `"average"` as its arguments, and it figures out which command it should display help for from there.

Mark-McCracken:
Thanks so much for the feedback!
I'll take another look and try your way instead, and will be back in a few days no doubt

natecook1000:
@Mark-McCracken I resolved this issue in #309 — it ended up being a little more involved than I expected. Thanks again for looking at tackling this, and please let me know if I can help with any other issues you'd like to take on!

Mark-McCracken:
Hi Nate, thanks so much, really sorry I didn't get any further with it, I did dig into it a bit, but found that the HelpCommand was actually getting a different input in the first place, and had a bit of a hard time with debugging it.
I did struggle a little and then I had some work commitments I had to spend my free time on. Thanks for sorting it!

PR #289, opened 2021-03-12
Hide flag inversion when help is hidden

werm098:
Fixes an issue where the inversion of a flag would not be hidden when the `ArgumentHelp.shouldDisplay` value is `false`.

Added a unit test to check for this behavior.

Example:
```
@Flag(inversion: .prefixedNo, help: .hidden)
var prompt: Bool = true
```
Would produce the following in the help text:
```
OPTIONS:
  --no-prompt
```
But the desired outcome is that there is no listing for this flag in the help text.

### Checklist


natecook1000 (approved):
Great find, thanks @werm098! 👏

natecook1000:
@swift-ci Please test

PR #290, opened 2021-03-15
Fix invalid state error when decoding an unparsed optional value

werm098:
Fixes “Internal error. Invalid state while parsing command-line arguments.” that is encountered when an unparsed value is optional.

Root causes:
- `ParsedArgumentsContainer.decodeNil` returns false for optional values because it only does a `!contains(key)` check. This should instead return nil if the value of the element is nil.
- The decoder did not know about unparsed input origins that and would result in unexpected behavior when decoding nil default values.
- The `value` of `Mirror.Child` is defined as `Any` but this is confusing because the value could be `Optional<Any>` which is not equal to `nil` even when the `Optional` case is `.none`.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/InputOrigin.swift
+++ b/Sources/ArgumentParser/Parsing/InputOrigin.swift
@@ -103,6 +103,12 @@ struct InputOrigin: Equatable, ExpressibleByArrayLiteral {
   }
 }
 
+extension InputOrigin {
+  var isDefaultValue: Bool {
+    return _elements.count == 1 && _elements.first == .defaultValue
+  }
+}
+
 extension InputOrigin.Element {
   static func < (lhs: Self, rhs: Self) -> Bool {
     switch (lhs, rhs) {
--- a/Sources/ArgumentParser/Utilities/MirrorExtensions.swift
+++ b/Sources/ArgumentParser/Utilities/MirrorExtensions.swift
@@ -1,0 +1,26 @@
+//===----------------------------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Argument Parser open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension Mirror {
+  /// Returns the "real" value of a `Mirror.Child` as `Any?`.
+  ///
+  /// The `value` of `Mirror.Child` is defined as `Any` but this is confusing because the value
+  /// could be `Optional<Any>` which is not equal to `nil` even when the `Optional` case is `.none`.
+  ///
+  /// The purpose of this function is to disambiguate when the `value` of a `Mirror.Child` is actaully nil.
+  static func realValue(for child: Mirror.Child) -> Any? {
+    if case Optional<Any>.none = child.value {
+      return nil
+    } else {
+      return child.value
+    }
+  }
+}
--- a/Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
+++ b/Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
@@ -57,6 +57,73 @@ extension UnparsedValuesEndToEndTests {
   }
 }
 
+// MARK: Two value + unparsed optional variable
+
+fileprivate struct Hogeraa: ParsableArguments {
+  var fullName: String? = "Full Name"
+}
+
+fileprivate struct Hogera: ParsableArguments {
+  @Option() var firstName: String
+  @Flag() var hasLastName = false
+  var fullName: String?
+  mutating func validate() throws {
+    if hasLastName { fullName = "\(firstName) LastName" }
+  }
+}
+
+fileprivate struct Piyo: ParsableArguments {
+  @Option() var firstName: String
+  @Flag() var hasLastName = false
+  var fullName: String!
+  mutating func validate() throws {
+    fullName = firstName + (hasLastName ? " LastName" : "")
+  }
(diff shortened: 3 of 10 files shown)


Sources/ArgumentParser/Utilities/MirrorExtensions.swift
```diff
@@ -0,0 +1,26 @@
...
+//===----------------------------------------------------------------------===//
+
+extension Mirror {
+  /// Returns the "real" value of a `Mirror.Child` as `Any?`.
+  ///
+  /// The `value` of `Mirror.Child` is defined as `Any` but this is confusing because the value
+  /// could be `Optional<Any>` which is not equal to `nil` even when the `Optional` case is `.none`.
+  ///
+  /// The purpose of this function is to disambiguate when the `value` of a `Mirror.Child` is actaully nil.
+  static func realValue(for child: Mirror.Child) -> Any? {
```
werm098, line 19: This name could probably be better?

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
```diff
@@ -106,3 +173,86 @@ extension UnparsedValuesEndToEndTests {
...
+      XCTAssertEqual(bar.baz?.name, "Some")
+      XCTAssertEqual(bar.baz?.age, 100)
+      XCTAssertNotNil(bar.bazz)
+      XCTAssertEqual(bar.bazz?.name, "Other")
+      XCTAssertEqual(bar.bazz?.age, 101)
+    }
+  }
+  
+  func testUnparsedNestedOptionalValue_Fails() {
+    XCTAssertThrowsError(try Bar.parse(["--baz", "xyz"]))
```
werm098, line 235: I can tone this down if it is too much.

Sources/ArgumentParser/Parsing/InputOrigin.swift
```diff
@@ -103,6 +103,12 @@ struct InputOrigin: Equatable, ExpressibleByArrayLiteral {
   }
 }
 
+extension InputOrigin {
+  var isDefaultValue: Bool {
+    return _elements.count == 1 && _elements.first == .defaultValue
```
werm098, line 108: Maybe there is a better way to do this? I also tried a `contains` and `Set` comparison but this seemed the simplest, easiest to read, and (assuming) most performant choice.

Sources/ArgumentParser/Utilities/MirrorExtensions.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+extension Mirror {
+  /// Returns the "real" value of a `Mirror.Child` as `Any?`.
+  ///
+  /// The `value` of `Mirror.Child` is defined as `Any` but this is confusing because the value
+  /// could be `Optional<Any>` which is not equal to `nil` even when the `Optional` case is `.none`.
+  ///
+  /// The purpose of this function is to disambiguate when the `value` of a `Mirror.Child` is actaully nil.
```
werm098, line 18: Maybe there is a Foundation change we could make too? I have run into this problem previously as well, not sure if this is a known behavior of optionals in swift.

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
```diff
@@ -57,6 +57,73 @@ extension UnparsedValuesEndToEndTests {
   }
 }
 
+// MARK: Two value + unparsed optional variable
+
+fileprivate struct Hogeraa: ParsableArguments {
+  var fullName: String? = "Full Name"
+}
+
+fileprivate struct Hogera: ParsableArguments {
```
werm098, line 66: If we move structs into the extension code block then we could reuse names for different test collections. This could cut down on the amount of metasyntactic type names (and modifications of those names) used.

natecook1000:
Thanks for this, @werm098! Do you have a reduced example that shows the incorrect behavior?

werm098:
> Thanks for this, @werm098! Do you have a reduced example that shows the incorrect behavior?

Sure thing, I can create a reduced example project (will post later). I also added to the unit tests for this incorrect behavior too: https://github.com/apple/swift-argument-parser/pull/290/files#diff-aae587b62a491288c0ad8d5dde7f8afd599171fdea4556602063ebe1ccc21259R66

werm098:
@natecook1000 Minimum reproducing project: [MVR.zip](https://github.com/apple/swift-argument-parser/files/6161244/MVR.zip)

You can change in the dependencies between 0.4.1 and 0.3.2 for an error vs no error respectively.

Sources/ArgumentParser/Utilities/MirrorExtensions.swift
natecook1000: I think this would work better as an `(Any) -> Any?` function, since it doesn't really have anything to do with `Mirror`, just the fact that `Mirror.Child.value` is an `Any`. How about calling it `nilOrValue`?

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
natecook1000: For sure! I don't think tests are very consistent around this, so if you'd like to structure yours that way, that'd be great.

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
natecook1000: This is okay by me

natecook1000 (commented):
Sorry for the delay here — this fix looks good! Just a couple nits below. Thanks again for taking this on.

natecook1000:
@swift-ci Please test

PR #299, opened 2021-04-15
Convert all public enums to structs

natecook1000:
Using structs for these public types (instead of enums) will let us add members in the future without breaking source compatibility. This is a source-breaking change, but not alarmingly so — these enum cases are typically used as static members, and that usage is maintained. Only switch exhaustivity is altered here.

### Checklist


MPLew-is:
Out of curiosity, is there a reason this approach is preferred over usage of an `@unknown default` case in user code that needs to deal with these objects? It feels to me like we already have a language-level mechanism to handle the scenario of depending on an enum that might have additional cases added to it down the line, and this change introduces more indirection to accomplish a similar goal.

natecook1000:
@MPLew-is The `@unknown default` bit is only available when a module is compiled with library evolution turned on, which isn’t enabled for this package. I also really don’t want to make _any_ promise about exhaustiveness for these types, so an `enum` is just not a great fit for the public API.

natecook1000:
@swift-ci Please test

PR #300, opened 2021-04-15
Exclude supercommands from help

miggs597:
Currently when asking for help with a subcommand we will receive help for it, and all supercommands that precede it. I propose the ability to only print help for the subcommand. The best example of this is in SwiftPM.

Current Behavior:

```
$ swift package init --help
OVERVIEW: Initialize a new package

USAGE: swift package init <options>

OPTIONS:
  -Xcc <Xcc>              Pass flag through to all C compiler invocations 
  -Xswiftc <Xswiftc>      Pass flag through to all Swift compiler invocations 
  -Xlinker <Xlinker>      Pass flag through to all linker invocations 
  -Xcxx <Xcxx>            Pass flag through to all C++ compiler invocations 
  -c, --configuration <configuration>
                          Build with configuration (default: debug)
  --build-path <build-path>
                          Specify build/cache directory 
  --cache-path <cache-path>
                          Specify the shared cache directory 
  --enable-repository-cache/--disable-repository-cache
                          Use a shared cache when fetching repositories (default: true)
  -C, --chdir <chdir>
  --package-path <package-path>
                          Change working directory before any other operation 
  --multiroot-data-file <multiroot-data-file>
  --enable-prefetching/--disable-prefetching
                          (default: true)
  -v, --verbose           Increase verbosity of informational output 
  --disable-sandbox       Disable using the sandbox when executing subprocesses 
  --manifest-cache <manifest-cache>
                          Caching mode of Package.swift manifests (shared: shared cache, local: package's build directory, none:
                          disabled (default: shared)
  --destination <destination>
  --triple <triple>
  --sdk <sdk>
  --toolchain <toolchain>
  --static-swift-stdlib/--no-static-swift-stdlib
                          Link Swift stdlib statically (default: false)
  --skip-update           Skip updating dependencies from their remote during a resolution 
  --sanitize <sanitize>   Turn on runtime checks for erroneous behavior, possible values: address, thread, undefined, scudo 
  --enable-code-coverage/--disable-code-coverage
                          Enable code coverage (default: false)
  --force-resolved-versions, --disable-automatic-resolution
                          Disable automatic resolution if Package.resolved file is out-of-date 
  --enable-index-store/--disable-index-store
                          Enable or disable  indexing-while-building feature 
  --enable-parseable-module-interfaces
  --trace-resolver
  -j, --jobs <jobs>       The number of jobs to spawn in parallel during the build process 
  --enable-build-manifest-caching/--disable-build-manifest-caching
                          (default: true)
[...]

PR #301, opened 2021-04-19
Hide option group with new OptionGroup constructor

miggs597:
Second PR that is needed to fix --help for SwiftPM

Related PRs:

[SwiftPM](https://github.com/apple/swift-package-manager/pull/3407)
[ArgumentParser](https://github.com/apple/swift-argument-parser/pull/300)

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
+++ b/Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
@@ -31,6 +31,7 @@
 @propertyWrapper
 public struct OptionGroup<Value: ParsableArguments>: Decodable, ParsedWrapper {
   internal var _parsedValue: Parsed<Value>
+  internal var _hidden: Bool?
   
   internal init(_parsedValue: Parsed<Value>) {
     self._parsedValue = _parsedValue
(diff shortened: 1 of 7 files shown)


Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
```diff
@@ -31,6 +31,7 @@
 @propertyWrapper
 public struct OptionGroup<Value: ParsableArguments>: Decodable, ParsedWrapper {
   internal var _parsedValue: Parsed<Value>
+  internal var _hidden: Bool?
```
natecook1000, line 34: Let's have this be non-optional, and just default to `false`. Can you call it something like `_hiddenFromHelp`?

Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
```diff
@@ -84,7 +85,17 @@ extension OptionGroup: CustomStringConvertible {
     case .value(let v):
       return String(describing: v)
     case .definition:
+      if let hidden = _hidden {
+        return "OptionGroup(*definition*) _hidden: \(hidden)"
+      }
```
natecook1000, line 90: This works, but we don't really want to pass configuration values through this type's string representation. Instead, let's add `var _hiddenFromHelp: Bool { get }` as a protocol requirement in `ArgumentSetProvider`, here: https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Parsable%20Types/ParsableArguments.swift#L230

You can add a default implementation that always returns `false`, and then this type will have the customized version, defined above.

natecook1000 (commented):
This is a great start @miggs597! 👏🏻 Added some notes about a different way to do this that will let us avoid using the `description` property.

natecook1000:
@swift-ci Please test

miggs597:
Are we ready to merge @natecook1000?

natecook1000:
Looks good to me! 🎉

PR #304, opened 2021-04-28
Capture values across multiple .upToNextOption option uses

natecook1000:
This fixes a bug where an `@Option` array defined with the `.upToNextOption` parsing strategy would only capture the last "group" of elements. e.g. in:

    example --test one two --test three four

the `--test` property would only have the value `["three", "four"]`.

Fixes rdar://73908471

### Checklist


natecook1000:
Note, this isn't source breaking, but changes the behavior of existing CLI tools. For options defined like `@Option(parsing: .upToNextOption) var names: [String] = []`, the following is no longer a valid use of the command:

```
$ example --names
```

natecook1000:
@swift-ci Please test

PR #315, opened 2021-05-19
Clean up internal property nesting

rauhul:
- Removes one layer of help properties by directly including the members
  of ArgumentHelp in ArgumentDefinition.Help. This also results in the
  discussion field which previously existed in both structures, now
  having a single source of truth. Adds helper method for setting each
  of these members using an instance of ArgumentHelp and additionally
  annotates each of the String? members with a new `@NonEmpty` property
  wrapper that ensures the wrapped string is either nil or nonempty so
  clients no do not need to check for the empty string case.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
+++ b/Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
@@ -39,8 +39,13 @@ struct ArgumentDefinition {
   
   struct Help {
     var options: Options
-    var help: ArgumentHelp?
-    var discussion: String?
+
+    // `ArgumentHelp` members
+    var abstract: String = ""
+    var discussion: String = ""
+    var valueName: String = ""
+    var shouldDisplay: Bool = true
+
     var defaultValue: String?
     var keys: [InputKey]
     var allValues: [String] = []
@@ -55,18 +60,26 @@ struct ArgumentDefinition {
     
     init(options: Options = [], help: ArgumentHelp? = nil, defaultValue: String? = nil, key: InputKey, isComposite: Bool = false) {
       self.options = options
-      self.help = help
       self.defaultValue = defaultValue
       self.keys = [key]
       self.isComposite = isComposite
+      updateArgumentHelp(help: help)
     }
     
     init<T: ExpressibleByArgument>(type: T.Type, options: Options = [], help: ArgumentHelp? = nil, defaultValue: String? = nil, key: InputKey) {
       self.options = options
-      self.help = help
+
       self.defaultValue = defaultValue
       self.keys = [key]
       self.allValues = type.allValueStrings
+      updateArgumentHelp(help: help)
+    }
+
+    mutating func updateArgumentHelp(help: ArgumentHelp?) {
+      self.abstract = help?.abstract ?? ""
+      self.discussion = help?.discussion ?? ""
+      self.valueName = help?.valueName ?? ""
+      self.shouldDisplay = help?.shouldDisplay ?? true
     }
   }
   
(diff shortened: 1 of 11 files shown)


rauhul:
@swift-ci Please test

rauhul:
I'm sure about the @NonEmpty property wrapper. I would personally prefer using something like `swift-nonempty` by point free to embed this information in the typesystem. @natecook1000 thoughts?

Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
```diff
@@ -39,8 +39,13 @@ struct ArgumentDefinition {
   
   struct Help {
     var options: Options
-    var help: ArgumentHelp?
-    var discussion: String?
+
+    // `ArgumentHelp` members
+    @NonEmpty var abstract: String?
+    @NonEmpty var discussion: String?
+    @NonEmpty var valueName: String?
```
natecook1000, line 46: Is there a semantic difference for these between `""` and `nil`? If not, I'd rather make these non-optional and just use empty strings.

natecook1000 (commented):
Looks good, other than `@NonEmpty`, as you mentioned. Thanks, @rauhul!

Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
rauhul: I don't believe so, just that the property is derived from `help: ArgumentHelp?` so the optionality is inherited from there. I am happy to make these non-optional and make the points of use test for empty string. I think the only rough point is losing the ergonomics of optional chaining and defaulting. ex: `valueName ?? otherValueName ?? "default"`

rauhul: Example: Screen Shot 2021-05-20 at 9 03 32 AM

rauhul: Example:
```diff
-    return help.valueName
-      ?? preferredNameForSynopsis?.valueString
+    return help.valueName.isEmpty
+      ? help.valueName
+      : preferredNameForSynopsis?.valueString
       ?? help.keys.first?.rawValue.convertedToSnakeCase(separator: "-")
       ?? "value"
```

rauhul: I think this is less desirable as it makes it easier to accidentally introduce bugs, like I just did in the above code. The condition should be `*!*help.valueName.isEmpty`

natecook1000: Yeah, the ergonomics are annoying, but I'd rather not keep them optional just for that purpose. That's kind of a funny usage point, since it's then lifting `help.valueName` into an optional to use with the following optional chaining… You could add a `String` extension like `func mapEmpty(_ f: () -> String) -> String` that would clear that up:

```swift
help.valueName.mapEmpty {
    preferredNameForSynopsis?.valueString
       ?? help.keys.first?.rawValue.convertedToSnakeCase(separator: "-")
       ?? "value"
}
```

rauhul:
@swift-ci please test

Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
rauhul: I went with you suggestion, but as an extension on `Collection` instead, since this replacement behavior seems generally useful.

---

It almost seems like there should be a `NullValue` protocol, such that this extension could operate on anything with a defined `null` and then the `??` operator could be extended to as well.

```swift
protocol NullValue {
  static var null: Self
}

extension NullValue where Self: Equatable {
  func mapNull(_ replacement: () -> Self) -> Self {
    self == null ? replacement() : self
  }
}

func ??<T: NullValue>(lhs: T, rhs: @autoclosure T) -> T {
  lhs.mapNull(rhs)
}
```

natecook1000: Yeah, it’s been proposed, I think!

natecook1000 (approved):

compnerd:
Hmm, this seems to have caused a regression on the Windows (bootstrap) build: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=51233&view=logs&j=73ce2acc-4df7-56ec-1894-6d2e45aae46f&t=9f34e3dd-3373-5395-8997-8bd4a6b1fb09

PR #319, opened 2021-05-24
Implement Subcommand Autodiscovery

Azoy:
This is an initial draft of subcommand auto discovery using reflection to find a list of protocol conformances to `ParsableCommand` and looking for nested types who conform to said protocol.

I haven't added any new tests, but I've changed some of the existing tests which essentially tests this functionality. There are still tests who have subcommands who aren't nested, thus they still use the subcommand array in `CommandConfiguration` which allows us to test that user defined subcommands also still work. If specific tests are needed, I'll be happy to add them! This is marked WIP because there are still some more documentation needed to fully explain this feature.

### Checklist


Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -44,6 +44,14 @@ extension ParsableCommand {
...
     throw CleanExit.helpRequest(self)
   }
+  
+  static var subcommands: [ParsableCommand.Type] {
+    guard configuration.subcommands != nil else {
+      return discoverSubcommands(for: Self.self)
+    }
+    
+    return configuration.subcommands.unsafelyUnwrapped
+  }
```
rauhul, line 54: nit: I think its simpler to read this as:
```swift
configuration.subcommands
  ?? discoverSubcommands(for: Self.self)
```

Azoy:
@swift-ci Please test

Azoy:
@swift-ci Please test

kylemacomber:
This is so cool!

Azoy:
@swift-ci Please test

Azoy:
> error: unexpected platform condition (expected 'os', 'arch', or 'swift')
#if _ptrauth(_arm64e)

Um, excuse me? (This has compiled in the past, and the sodlib makes use of this conditional as well??)

Sources/ArgumentParser/Utilities/SubcommandDiscovery.swift
```diff
@@ -0,0 +1,106 @@
...
+  }
+  
+  subcommandLock.lock()
+  
+  defer {
+    subcommandLock.unlock()
+  }
+  
+  guard !discoveredSubcommands.keys.contains(selfMetadata.pointer) else {
+    return discoveredSubcommands[selfMetadata.pointer].unsafelyUnwrapped
```
kylemacomber, line 45: I think we should avoid the use of `unsafelyUnwrapped` unless absolutely necessary for performance. My intuition is that in this kind of code we can get away with using the force unwrap operator instead.

Sources/_Runtime/Utilities/ImageInspection.swift
```diff
@@ -0,0 +1,126 @@
...
+}
+
+let conformanceLock = NSLock()
+var conformances:
+  [ContextDescriptor: [ContextDescriptor: [ConformanceDescriptor]]] = [:]
+
+@_cdecl("registerProtocolConformances")
+func registerProtocolConformances(_ section: UnsafeRawPointer, size: Int) {
+  // This section is a list of relative pointers.
+  for i in 0 ..< size / MemoryLayout<Int32>.size {
```
kylemacomber, line 29: Should (can?) we use `UnsafeRawBufferPointer` here instead of `UnsafeRawPointer` and `size`?

kylemacomber (commented):
We should probably leave at least one test that explicitly lists out the subcommands and/or we should add a test that explicitly lists out the subcommands (and deliberately omits a nested ParsableCommand)

Sources/_Runtime/Utilities/ImageInspection.swift
Azoy: We could, but since `UnsafeRawBufferPointer` isn't representable in C, we'd have to do construction of a mock buffer over on the C side, and on the Swift side we'd have to do some bit casting to eventually get the buffer pointer.

Sources/ArgumentParser/Utilities/SubcommandDiscovery.swift
```diff
@@ -0,0 +1,106 @@
...
+    ) as! ExistentialMetadata
+    
+    parsableCommand = parsableCommandMetadata.protocols[0]
+  }
+  
+  // Grab all of the conformances to ParsableCommand in the same module that
+  // this ParsableCommand was defined in.
+  let conformances = _Runtime.getConformances(for: parsableCommand, in: module)
+  
+  // FIXME: Maybe we want to reserve some initial space here?
```
natecook1000, line 69: No need! You should only reserve initial space when you statically know a minimum size. If we preemptively reserve space here, those will be unnecessary allocations for commands that _don’t_ have subcommands.

Sources/ArgumentParser/Utilities/SubcommandDiscovery.swift
```diff
@@ -0,0 +1,106 @@
...
+    // If we don't have a context descriptor, then an ObjC class somehow
+    // conformed to a Swift protocol (not sure that's possible).
+    guard let descriptor = conformance.contextDescriptor else {
+      continue
+    }
+    
+    // This is okay because modules can't conform to protocols, so the type
+    // being referenced here is at least a child deep in the declaration context
+    // tree.
+    let parent = descriptor.parent.unsafelyUnwrapped
```
natecook1000, line 82: We don't really use `unsafelyUnwrapped` unless we've identified a specific performance issue — regular unwrapping should be fine here.

```suggestion
    let parent = descriptor.parent!
```
... or use `guard`/`fatalError` to unwrap with an error message.

Sources/_Runtime/Metadata/ClassMetadata.swift
```diff
@@ -0,0 +1,52 @@
...
+    #if _ptrauth(_arm64e)
+    let signed = __ptrauth_strip_asda(view._descriptor.pointer)
+    return ContextDescriptor(pointer: signed)
+    #else
+    return view._descriptor
+    #endif
+  }
+}
+
+struct _ClassMetadata {
```
natecook1000, line 32: These `View` declaration need to move into `_CRuntime` — Swift doesn’t guarantee type layout, so there’s no way to know / ensure that these will match the runtime’s metadata layout.

I think computed properties like `ClassMetadata.descriptor`, above, are the right way to go for anything where you need an actual type instead of an int or a pointer.

Sources/_Runtime/Metadata/MetadataAccessFunction.swift
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public struct MetadataAccessFunction {
+  let pointer: UnsafeRawPointer
+
+  public func callAsFunction() -> Any.Type {
```
natecook1000, line 15: This is a great way to implement this 👍🏻

Sources/_Runtime/Utilities/ImageInspection.swift
natecook1000: The raw buffer would also be a weird interface because we're stepping by an `Int32` at a time, but don't want to bind the memory to `Int32`. I might approach this by just iterating over the pointers, e.g.:

```swift
for start in stride(from: section, to: section + size, by: MemoryLayout<Int32>.stride) { ... }
```

Sources/_Runtime/Utilities/RelativePointers.swift
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+protocol RelativePointer {
+  associatedtype Pointee
+
+  var offset: Int32 { get }
```
natecook1000, line 15: Does this need to be part of the protocol interface?

natecook1000 (commented):
This is looking great, @Azoy! 👏🏻👏🏻

In addition to the notes below, let's make sure we still have tests covering manually specified subcommands as well as the empty override (a command with nested subcommands, but with `subcommands: []` as part of the configuration).

Sources/_Runtime/Utilities/RelativePointers.swift
Azoy: Huh, it seems I don't actually use `pointee(from:)` anywhere (which is the only default that needed offset information, unless we wanted a `isNull` or something). The more that I think about it, for ArgumentParser's use case we can get rid of this protocol entirely.

Sources/_Runtime/Metadata/ClassMetadata.swift
Azoy: Argh, you're so right. Really wishing we had a `@repr(c)` attribute or similar in Swift!

hassila:
@Azoy Did/would you consider splitting out _Runtime & _CRuntime to a separate GitHub SPM package? I'm starting to look at prototyping a dynamic loading plug-in API (https://forums.swift.org/t/swift-dynamic-loading-api/39495/21) and I think the approach you took there would be a possible nice way forward for that use case too. I was thinking of linking the plugins with that hypothetical separate package with the metadata runtime to allow for fully dynamic discovery of types conforming to the specific API protocol. It seems like generically interesting introspection capability...

PR #320, opened 2021-06-01
Fix custom completion args for bash

Frizlab:
I’ve noticed when using a custom completion in bash, the custom completion handler does not receive all the arguments as would be expected. This PR fixes that.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -149,7 +149,7 @@ struct BashCompletionsGenerator {
                 ?? arg.help.keys.first?.rawValue ?? "---"
           
           return """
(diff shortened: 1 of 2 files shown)


kylemacomber:
@swift-ci please test

kylemacomber:
@Frizlab could describe a simple concrete example where the current behavior returns the wrong result?

Frizlab:
Here is an example! https://gitlab.com/frizlab-demo-projects/swift-argument-parser-example-for-pr-320

To see the bug, load the completion script in bash, and type `./.build/debug/example --option1 a ` (with a space after the a) and then trigger autocomplete. The file `completion_debug.txt` will be created, with one line per completion call, each line showing the content of the `args` array in the completion function.

You’ll see the args array only contain one entry, but it should be containing three.

With my fix, it does contain the three elements.

As a comparison one can try zsh completion, which does indeed show the three arguments as expected.

natecook1000:
Thanks for this fix, @Frizlab! 🎉

PR #321, opened 2021-06-01
Fix default value description for URL

Frizlab:
At first I noticed a crash when running the tests of this project. It was due to weird chars in my CWD.
This lead me to discover a bug where the default value description of a URL checks for the `absoluteString` value of itself instead of the path + whether it is a file URL!

Note: I did not add a test specifically, I don’t really know how this would be tested. Maybe create a folder with a weird name and check everything is good? Also a test with an https URL as default w/ a path being the same as current CWD?
Tell me, I’ll do them if needed.

### Checklist

### DIFF
--- a/Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
+++ b/Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
@@ -25,7 +25,7 @@ extension URL: ExpressibleByArgument {
   }
 
   public var defaultValueDescription: String {
-    self.absoluteString == FileManager.default.currentDirectoryPath
+    self.path == FileManager.default.currentDirectoryPath && self.isFileURL
       ? "current directory"
       : String(describing: self)
   }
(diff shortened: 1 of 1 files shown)


kylemacomber:
After a quick skim of the documentation, it looks like this might actually be a Foundation bug. `FileManager.currentDirectoryPath` is defined as returning a type `String`, but it appears possible it will return `nil`. It should probably be defined as returning `String?`.

Frizlab:
@kylemacomber I don’t think so. Maybe it’s a separate issue, but unrelated to the one I reported.

`URL(string: FileManager.default.currentDirectoryPath)` is likely to fail, and is definitely _not_ supposed to be used to create a file URL.
It might not return `nil` if by a lucky chance the current directory path happens to be a valid URL (e.g. a path with no space in it), but 1/ the returned URL will **never** be a file URL, and 2/ if the current path contains a space or any character not allowed in a URL the init will definitely return `nil`.

The correct way to init a file URL is `URL(fileURLWithString: "/path/to whatever")`, which will never return nil in addition to correctly returning a file URL.

kylemacomber:
@swift-ci please test

PR #332, opened 2021-06-22
[WIP] Manual page generation

rauhul:
This is definitely a work in progress, but I'd like to get some comments on the approach and get ideas on how best to integrate the functionality into argument-parser. My initial approach is to introduce a new module called `ArgumentParserGenerators` which can be used to create a parallel tool to a user's main tool for generating man-pages. I think completion scripts could be moved here too in order to reduce binary bloat.

All comments/questions/suggestions are welcome!


Package.swift
```diff
@@ -14,6 +14,7 @@ import PackageDescription
 
 var package = Package(
     name: "swift-argument-parser",
+    platforms: [.macOS("11")],
```
rauhul, line 17: remove this

Package.swift
```diff
@@ -24,22 +25,37 @@ var package = Package(
...
+            dependencies: ["ArgumentParser", "MDoc"]),
         .target(
             name: "ArgumentParserTestHelpers",
             dependencies: ["ArgumentParser"]),
-
         .target(
-            name: "roll",
-            dependencies: ["ArgumentParser"],
-            path: "Examples/roll"),
+            name: "MDoc",
```
rauhul, line 35: Maybe I should fold this into `ArgumentParserGenerators`

Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
```diff
@@ -21,11 +22,26 @@ struct ArgumentDefinition {
...
+
+    var `case`: UpdateCase {
+      switch self {
+      case .nullary: return .nullary
+      case .unary: return .unary
+      }
+    }
+  }
+
+  public enum UpdateCase {
```
rauhul, line 34: missing `@_spi`

rauhul:
@tomerd do you have any ideas on if/how this could work the spm plugin stuff?

hassila:
I think splitting out completion script / man page generation to a separate tool would be nice for code size (given the other case about that). Also, are you thinking of DocC style comments inline with the code for the actual man page content? It would be nice if the same syntax could be used?

tomerd:
I assume the desired behavior here would be to create a custom command such as `swift package generate-man`, if so this needs to wait for additional features in the plugin system which we are considering to add, cc @abertelrud.

rauhul:
> I assume the desired behavior here would be to create a custom command such as `swift package generate-man`, if so this needs to wait for additional features in the plugin system which we are considering to add, cc @abertelrud.

I'm not exactly sure what the desired UX is, but I think it would be nice to build a package for distribution which would mean creating a layout maybe like this:
```bash
package
├── bin
├── lib
└── share
```

This would be a nice foil to "spm install" which maybe could install any package that is buildable for distribution. Thoughts?

rauhul:
> Also, are you thinking of DocC style comments inline with the code for the actual man page content

I think this would be an excellent future addition, but probably not as inline comments, but instead in the additional articles/tutorials format the DocC supports!

hassila:
> > Also, are you thinking of DocC style comments inline with the code for the actual man page content
> 
> 
> 
> I think this would be an excellent future addition, but probably not as inline comments, but instead in the additional articles/tutorials format the DocC supports!
> 
> 

That would be awesome.

tomerd:
@rauhul with swiftpm commands plugins now in place, would be interesting to revisit integration options

rauhul:
@tomerd I gave a plugin a shot, but I wasn't sure how to fit all the pieces together. Would you mind taking a look at the approach in the most recent commit?

tomerd (commented):
i think in this case a command plugin would be a better choice over  builtTool one. the docc plugin is an example of a medium complexity command plugin. feel free to dm for more guidance

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

natecook1000 (approved):
This is great stuff, @rauhul! :clap: :clap: We do need to add a test harness as a follow-up so that we can make sure we're not breaking this plug-in as we move forward.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #334, opened 2021-06-30
Fix version documentation in README

eneko:
README file incorrectly referenced 'minor' version when it should be 'patch' version.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ add the following line to the dependencies in your `Package.swift` file:
 ```
 
 Because `ArgumentParser` is under active development,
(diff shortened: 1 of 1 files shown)


natecook1000:
Thanks, @eneko! The current wording has the intended meaning — changes within the current minor version (which as of today is `0.4`) will not break existing users of the library. When the library moves to minor version `0.5`, the changes may break existing projects.

eneko:
Hi @natecook1000, I see, thanks. I misinterpreted the "within", as if upgrading minor version wouldn't include breaking changes. I guess it is not possible to upgrade within patch releases, since there is nothing in between 😂

Thanks for the feedback!

PR #335, opened 2021-07-08
Improvements to `--dump-help`

rauhul:
- Removes `HelpInfo` in favor of a recursively defined `CommandInfo`
  which contains more raw metadata about the source command.
  Additionally, introduces a top level `TopInfo` type with a
  serialization version to aid future tooling.

- Updates tests to match the new serialized format.

- Renames `DumpHelpInfoGenerator` to `DumpHelpGenerator` to align the
  type with the `--dump-help` flag.


rauhul:
@swift-ci please test

rauhul:
@natecook1000 what are your thoughts on the serialization format changes? I'm also wondering if the format should be evolvable/versioned or expect all clients to stay in lock step with the format? I defaulted towards the former by introducing a `serializationVersion` property.

rauhul:
@swift-ci please test

Sources/ArgumentParser/Usage/ToolInfo.swift
```diff
@@ -0,0 +1,139 @@
...
+@_spi(ToolInfo)
+public struct ToolInfoV0: Codable, Hashable {
+  public var serializationVersion: Int = 0
+  public var command: CommandInfoV0
+
+  public init(
+    serializationVersion: Int = 0,
+    command: CommandInfoV0
+  ) {
+    self.serializationVersion = serializationVersion
```
natecook1000, line 36: What's your plan for versioning this? Can `ToolInfoV0` support serialization version 1?

Sources/ArgumentParser/Usage/ToolInfo.swift
```diff
@@ -0,0 +1,139 @@
...
+    self.command = command
+  }
+}
+
+@_spi(ToolInfo)
+public struct CommandInfoV0: Codable, Hashable {
+  public var superCommands: [String]?
```
natecook1000, line 43: This seems useful 👍🏻

Sources/ArgumentParser/Usage/ToolInfo.swift
```diff
@@ -0,0 +1,139 @@
...
+
+  public var commandName: String
+  public var abstract: String?
+  public var discussion: String?
+
+  public var defaultSubcommand: String?
+  public var subcommands: [CommandInfoV0]?
+  public var arguments: [ArgumentInfoV0]?
```
natecook1000, line 51: I think it's a good idea to have these all together like this.

Sources/ArgumentParser/Usage/ToolInfo.swift
```diff
@@ -0,0 +1,139 @@
...
+
+  public var shouldDisplay: Bool
+  public var isOptional: Bool
+  public var isRepeating: Bool
+
+  public var names: [NameInfoV0]?
+  public var preferredName: NameInfoV0?
+
+  public var valueName: String?
+  public var defaultValue: String?
```
natecook1000, line 107: What do you think about capturing allowed values when that's enumerable — e.g. `--config debug` and `--config release`? That could allow for a better generated GUI that some folks have suggested.

Sources/ArgumentParser/Usage/ToolInfo.swift
```diff
@@ -0,0 +1,139 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import Foundation
+
+fileprivate extension Collection {
+  var `optional`: Self? { isEmpty ? nil : self }
```
natecook1000, line 15: Maybe name this `nonEmpty` instead?

natecook1000 (commented):
Thanks for this, @rauhul! I think the changes make sense, esp along with the serialization version, since I know there will be additional features coming that can't be captured currently (bc they don't exist). Added a variety of notes, this largely looks ready to go.

FYI I think I'm going to limit access to the API and change the flag name to `--experimental-dump-help` in the first release of this.

Sources/ArgumentParser/Usage/ToolInfo.swift
rauhul: My initial thought is that the serialization version would essentially be a semver major indicator used to present better error messages.

Evolving the serialized format could follow these rules:
- New fields can be added freely
- Many fields can be omitted removed (because they are optional)
- Changes to the encoded format can be handled by manually implementing `init(from decoder: Decoder)`

If a backwards incompatible change needs to be made, we can bump the serialization version which would have a more legible error than "decoding failure".

(However I'm not too happy with this plan, so I'd be really happy to get suggestions)

rauhul:
@swift-ci please test

natecook1000:
@swift-ci Please test

rauhul:
@natecook1000 It looks like the linux build is failing due to the use of `@_spi`, and the macOS build is failing due to the use of `XCTExpectFailure`. Any thoughts on how to proceed?

natecook1000:
Ugh — looking at whether we can use a higher version on CI to support this.

rauhul:
Would it be possible to only bump the linux version? Since swift is not ABI stable on linux, I think users have no good reason to use old versions on linux.

The macOS failure will be resolved by the other issue 340 PR

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

rauhul:
@compnerd how do you build with `cmake`?

compnerd:
You should be able to install a windows snapshot toolchain and then just use `cmake -B out -G Ninja -S . -DCMAKE_BUILD_TYPE=Release && ninja -C build`.  I'll try to see if I can find some time to test this in the next few days if you've not beat me to it.

compnerd:
This seems to have regressed the CMake build :-(

PR #339, opened 2021-07-19
Use standard path for --dump-help help/completions, remove completions

natecook1000:


natecook1000:
@swift-ci Please test

Issue #338, opened 2021-07-18
v0.4.3 crashs when using Transform closure


griffin-stewie:
It crashes(SIGABRT) on Release build when I use Transform closure.

I tested several versions bellow. Only v0.4.3 has this issue.

- ✅ 0.3.2
- ✅ 0.4.0
- ✅ 0.4.1
- ✅ 0.4.2
- ‼️ 0.4.3
- ‼️ main branch

**ArgumentParser version:** `0.4.3`
**Swift version:** Apple Swift version 5.4 (swiftlang-1205.0.26.9 clang-1205.0.19.55) Target: x86\_64-apple-darwin20.5.0

### Checklist

- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

[Here](https://github.com/griffin-stewie/sample_cal) is my sample code to reproduce this issue.

1. Build sample code with Release configuration.
2. Run this `sample_cal description --start-date 2021/07/18_11:22:33 --name aaa`

I suspect that the crash is caused by using `@Option(name: , help: , transform: )`.

One more thing, It prints a blank where `start: Date` should be printed, if I deleted `@Option() var name: String`.

### Expected behavior

No crash, and prints OptionGroup’s description.

### Actual behavior

Crashed.

Xcode shows some logs on console, like bellow:

```
sample\_cal(687,0x10026fe00) malloc: Double free of object 0x105a47cc0
sample\_cal(687,0x10026fe00) malloc: \*\*\* set a breakpoint in malloc\_error\_break to debug
```

rauhul:
I played around with your sample code in Xcode and on the command line. I tried reducing the example and had a wild amount of non deterministic results based on: clean vs incremental builds, `Date` vs `Int`, etc... To me, this screams mis-compile and not library bug. Additionally, one of crashes modes I've seen is an `EXC_BAD_ACCESS ` in the `as?` operator which is also very suspicious.

```
Apple Swift version 5.3.1 (swiftlang-1205.0.16.11 clang-1205.0.19.6)
Target: x86_64-apple-darwin20.3.0
SAP: 0.4.3 & ee32b809403dc77bded883194e3bbcb107be5104
```

natecook1000:
Thanks for the detailed report, @griffin-stewie!

griffin-stewie:
Thanks @natecook1000 and @rauhul !

[...]

PR #342, opened 2021-07-30
Work around runtime issue introduced in 0.4.3

natecook1000:
ArgumentParser version 0.4.3 added an experimental feature for hiding option groups at the command level, which is triggering a runtime crash when commands are compiled in release mode, as described in #338 and rdar://80796582. This includes changes that work around both of those crashes. I'll cherry pick this change over to the current release as 0.4.4 as well.

### DIFF
--- a/Examples/repeat/main.swift
+++ b/Examples/repeat/main.swift
@@ -11,27 +11,34 @@
 
 import ArgumentParser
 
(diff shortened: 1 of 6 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #365, opened 2021-10-27
Add test for SplitArguments.contains

keith:
This test covers the cast that an invalid long argument is not mistaken
for any other argument with a similar prefix.


natecook1000:
@keith I'm a little conflicted on this — we currently display the help screen when a user provides a help flag, even if they've also provided other, incorrect input. That is, `repeat --asdf --help` will just show the help screen.

That said, the current behavior gets in the way of a user being able to understand what they did wrong, so I agree. We should change the behavior so that short flag names only match a _whole_ input instead of matching a sub-group.

keith:
Yea, FWIW I discovered this with a custom case of my own, instead of the help case, where my flag just happened to start with `-h`, which I think is a much more confusing case than my example here

PR #377, opened 2021-12-08
Rename `...using_command` fish function, set `$cmd` in local scope

floam:
This PR has two minor changes for FishCompletionsGenerator.swift.

The first is related to the name of the fish function created that's used to determine whether to show completions or not, depending on the commandline entered thus far. Functions beginning with `__fish` **really** ought only be things internal to fish that are included with the shell. So prefix the function `_swift` instead.

Secondly, `set --local cmd` instead of `set cmd` - this ensures running the function does not accidentally clobber a global or universal variable.

### DIFF
--- a/Sources/ArgumentParser/Completions/FishCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/FishCompletionsGenerator.swift
@@ -2,8 +2,8 @@ struct FishCompletionsGenerator {
   static func generateCompletionScript(_ type: ParsableCommand.Type) -> String {
     let programName = type._commandName
     let helper = """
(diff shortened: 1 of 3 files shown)


rauhul:
@swift-ci please test

floam:
There was a test failure; I believe I force-pushed the correction.

rauhul:
@swift-ci please test

floam:
Okay, I'm finally running swift test myself now, and there are now zero failures. I'm just an old chunk of coal.

rauhul:
@swift-ci please test

natecook1000:
This is great, thanks @floam! 👏🏻

PR #385, opened 2022-01-09
Clean up help generation

rauhul:
- Adds check to ensure that `wrapped(to:wrappingIndent:)` doesn't
  attempt to retrieve a negative prefix.
- The Usage struct was composed of an array of strings which always
  contained exact one string at runtime. This struct has been removed
  and replaced with a single usage string.
- Removes HelpGenerator._screenWidthOverride in favor of explicitly
  setting the screen width in generateHelp calls.


rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

natecook1000:
@swift-ci Please test

natecook1000:
💻👍🏻

PR #387, opened 2022-01-09
Fix dump help crash

### LINKED ISSUE #375: Having non-wrapped or default property in command breaks `--experimental-dump-help`
When I was using `--experimental-dump-help` with a command that has non-wrapped or default property, it showed the error
```
ArgumentParser/DumpHelpGenerator.swift:153: Fatal error: argument.kind must not be .default
```
caused by https://github.com/apple/swift-argument-parser/blob/e1465042f195f374b94f915ba8ca49de24300a0d/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift#L153

**ArgumentParser version:** main
**Swift version:** swift-driver version: 1.26.9 Apple Swift version 5.5 (swiftlang-1300.0.31.1 clang-1300.0.29.1)
Target: x86_64-apple-macosx11.0

### Checklist

### Steps to Reproduce
It can be reproduced if you add something like
```
var i: Int?
```
in math command from the examples.

### Expected behavior
It should output the JSON without including the default property.

### Actual behavior
It fails with the fatal error shown above.

It seems that this happens because default properties are not properly removed if `argument.help.isComposite` is false. The easy fix is to filter arguments based on their `Kind` and remove default prperties but I am not sure if that is the best way to go.

rauhul:
- Fixes #375
- Fixes a bug with the experimental dump help generator which would
  occur when a ParsableCommand has non-Argument properties.

### DIFF
--- a/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
@@ -107,7 +107,7 @@ fileprivate extension CommandInfoV0 {
     let arguments = commandStack
       .allArguments()
       .mergingCompositeArguments()
-      .map(ArgumentInfoV0.init)
+      .compactMap(ArgumentInfoV0.init)
 
     self = CommandInfoV0(
       superCommands: superCommands,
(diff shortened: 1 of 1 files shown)


rauhul:
@swift-ci please test

natecook1000:
@swift-ci Please test

natecook1000:
🎉

Issue #384, opened 2022-01-05
Add help level that's never visible


natecook1000:
With the addition of a `--help-hidden` flag that displays an extended help screen (as in #366), command-line tool developers may want to specify a level of hidden-ness that isn't visible in that extended display. The `ArgumentHelp` type should be updated to include multiple levels of visibility instead of just using a Boolean value.

#### Existing functionality

```swift
// Hidden flag
@Flag(help: .hidden)
var experimentalExtraSpicy = false

// Hidden flag, but with abstract provided
@Flag(help: ArgumentHelp("Experimental extra sauce setting", shouldDisplay: false))
var experimentalExtraSauce = false
```

#### Proposed

A visibility API:

```swift
enum Visibility {
    /// Show help for this argument whenever appropriate
    case default

    /// Only show help for this argument in the extended help screen
    case hidden

    /// Never show help for this argument
    case private
}
```

...and its use at the declaration site:

```swift
// Hidden flag
@Flag(help: .hidden)
var experimentalExtraSpicy = false

// Hidden flag, but with abstract provided
@Flag(help: ArgumentHelp("Experimental extra sauce setting", visibility: .hidden))
var experimentalExtraSauce = false

// Private flag
@Flag(help: .private)
var deprecatedFlag = false
```

McNight:
Hello @natecook1000 and others,
I want to get involved in some Swift projects I particularly like and I saw that this issue was a good opportunity 😃
[I already started to implement it](https://github.com/McNight/swift-argument-parser/tree/mcnight/argumenthelp_visibility_levels), following your proposed API. However, I have some questions:
- do you think it's suitable right now for a draft pull request ?
- some tests rely on the old API where some arguments are expected to be hidden so what should we do ? modify the expectations or modify the `ParsableArguments` entities (using the new API) ? Also new tests obviously have to be added!
- what about documentation ?
Thanks!

natecook1000:
@McNight — thanks for your interest in contributing! Feel free to open a draft PR whenever you're ready for feedback.

Since #366 hasn't landed yet, the work you're doing won't have any visible effects. There shouldn't need to be any changes to existing tests, just new unit tests that exercise the new API. Note that all the API in _existing_ tests needs to stay for source compatibility, so if some of the tests are no longer compiling, that's a cue that something got lost. You can add deprecated versions of old API that call through to your new API to resolve those.

[...]

PR #390, opened 2022-01-13
ArgumentHelp.Visibility levels API

McNight:
This pull request (as a draft for now) adds a new `ArgumentHelp.Visibility` API with multiple levels allowing to indicate whether help messages should be shown in the extended help display.
It follows suggestions made by @natecook1000 in issue https://github.com/apple/swift-argument-parser/issues/384.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
+++ b/Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
@@ -11,6 +11,18 @@
 
 /// Help information for a command-line argument.
 public struct ArgumentHelp {
+  /// Visibility level of an argument's help.
+  public enum Visibility {
+      /// Show help for this argument whenever appropriate.
+      case `default`
+
+      /// Only show help for this argument in the extended help screen.
+      case hidden
+
+      /// Never show help for this argument.
+      case `private`
+  }
+
   /// A short description of the argument.
   public var abstract: String = ""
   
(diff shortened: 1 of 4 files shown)


McNight:
Some comments:

- `hidden` and `private` have the same properties for now to make sure all tests are still passing
- I added a test in `HelpGenerationTests.swift` near the bottom of the file, I don't know if it should be regrouped with the `testHelpWithHidden()` test
- I explicitly made several commits but I could squash them if needed

Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
```diff
@@ -24,26 +35,31 @@ public struct ArgumentHelp {
   ///   flags don't include a value.
   public var valueName: String?
   
-  /// A Boolean value indicating whether this argument should be shown in
+  /// A visibility level indicating whether this argument should be shown in
   /// the extended help display.
-  public var shouldDisplay: Bool = true
```
rauhul, line 29: I dont think we can eliminate this property since it is part of our public API, but it can be deprecated and forward get/set to the new visibility property you've defined

Same for the initializer below

McNight: Indeed, my bad

McNight:
I made the modifications to keep and mark the `shouldDisplay` boolean property as deprecated. I wanted to do the same for the other initializer but I had some trouble to mark it as public or even internal since it conflicts with the other one.

Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
```diff
@@ -24,9 +35,21 @@ public struct ArgumentHelp {
...
+    get {
+      return visibility == .default
+    }
+    set {
+      visibility = newValue ? .default : .hidden
+    }
+  }
   
   /// Creates a new help instance.
   public init(
```
rauhul, line 55: Could we add a depreciation to this initializer start moving away from `shouldDisplay`

McNight: Sure!

McNight: I marked this initializer as deprecated but didn't provide a deprecation message since I don't any a public alternative for now. However, I just learned (from you :)) about `@_disfavoredOverload`, do you think it could be useful here?

rauhul: I think the message should guide users to `ArgumentHelp.init(_:discussion:valueName: visibility:)`. `@_disfavoredOverload` probably is needed as well since both inits can be called without providing any arguments.

Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
```diff
@@ -11,6 +11,17 @@
 
 /// Help information for a command-line argument.
 public struct ArgumentHelp {
+  public enum Visibility {
```
natecook1000, line 14: Could you add an abstract for this type, as well?

natecook1000 (commented):
Looking good, @McNight! Don't worry about squashing commits, we do that when merging.

Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
rauhul: That's a clever solution!

Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
McNight: Sure !

natecook1000 (approved):
Changes look good! 👍🏻

Issue #381, opened 2021-12-24
Unable to set default value of `String`-typed option with `help` and `completion`


stevapple:
Let's say, it's possible to configure a `String`-typed option:
- with `help` text;
- with `completion`;
- with default value.

In fact, we can do either two of these, but combining three of these will cause compilation failure.

**ArgumentParser version:** `1.0.2`, `main`
**Swift version:** swift-driver version: 1.26.21 Apple Swift version 5.5.2 (swiftlang-1300.0.47.5 clang-1300.0.29.30)

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

Minimal code to reproduce:

```swift
import ArgumentParser

struct Run: ParsableCommand {
    @Option(help: "", completion: .file())
    var string: String = ""

    func run() {}
}
```

Similar bug with `String?`:

```swift
@Option(completion: .file())
var string: String? = ""
```

### Expected behavior

Compile successfully.

### Actual behavior

```
main.swift:5:26: error: cannot convert value of type 'String' to expected argument type '[String]'
    var string: String = ""
                         ^
main.swift:5:9: error: property type 'String' does not match 'wrappedValue' type 'Array<String>'
    var string: String = ""
        ^~~~~~
main.swift:5:9: note: cannot automatically synthesize 'Decodable' because '<<error type>>' does not conform to 'Decodable'
    var string: String = ""
        ^
main.swift:3:8: error: type 'Run' does not conform to protocol 'Decodable'
struct Run: ParsableCommand {
       ^
Swift.Decodable:2:5: note: protocol requires initializer 'init(from:)' with type 'Decodable'
    init(from decoder: Decoder) throws
    ^
```

### Workaround

For anyone who encountered such error, you can workaround with a non-op `transform`:

```swift
// ✅ It works!
@Option(help: "", completion: .file(), transform: {String($0)})
var string: String = ""
```

natecook1000:
Oooof, it looks like we have some parameter order inconsistency with the `Option` initializers — the one you want has `help` after `completion`. This code works instead:

```swift
@Option(completion: .file(), help: "")
var string: String = ""
```

We should deprecate that initializer and add a new one that has the parameters in the correct order.

McNight:
Hey, me again 😄
I allowed myself to create a PR, following your recommendations.

[...]

PR #391, opened 2022-01-17
Deprecate Option initializer and add a new one with parameters in order

McNight:
This PR marks an `Option` initializer as deprecated because it had its last 2 parameters inverted (help should be before completion, following order of all other initializers).
It also adds a new one with parameters in correct order.
It follows recommendations of @natecook1000 in issue https://github.com/apple/swift-argument-parser/issues/381.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Option.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Option.swift
@@ -127,6 +127,23 @@ extension Option where Value: ExpressibleByArgument {
   }
 
   /// Creates a property with a default value provided by standard Swift default value syntax.
+  @available(*, deprecated, message: "Swap the order of your 'help' and 'completion' arguments.")
+  public init(
+    wrappedValue: Value,
+    name: NameSpecification = .long,
+    parsing parsingStrategy: SingleValueParsingStrategy = .next,
+    completion: CompletionKind?,
+    help: ArgumentHelp?
+  ) {
+    self.init(
+      name: name,
+      initial: wrappedValue,
+      parsingStrategy: parsingStrategy,
+      help: help,
+      completion: completion)
+  }
+
+  /// Creates a property with a default value provided by standard Swift default value syntax.
   ///
   /// This method is called to initialize an `Option` with a default value such as:
   /// ```swift
@@ -138,12 +155,13 @@ extension Option where Value: ExpressibleByArgument {
   ///   - name: A specification for what names are allowed for this flag.
   ///   - parsingStrategy: The behavior to use when looking for this option's value.
   ///   - help: Information about how to use this option.
+  ///   - completion: Kind of completion provided to the user for this option.
   public init(
     wrappedValue: Value,
     name: NameSpecification = .long,
     parsing parsingStrategy: SingleValueParsingStrategy = .next,
-    completion: CompletionKind? = nil,
-    help: ArgumentHelp? = nil
+    help: ArgumentHelp? = nil,
+    completion: CompletionKind? = nil
   ) {
     self.init(
       name: name,
--- a/Tests/ArgumentParserEndToEndTests/DefaultsEndToEndTests.swift
+++ b/Tests/ArgumentParserEndToEndTests/DefaultsEndToEndTests.swift
@@ -777,3 +777,19 @@ extension DefaultsEndToEndTests {
     }
   }
 }
+
+@available(*, deprecated)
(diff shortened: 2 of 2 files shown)


Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -138,12 +138,40 @@ extension Option where Value: ExpressibleByArgument {
   ///   - name: A specification for what names are allowed for this flag.
   ///   - parsingStrategy: The behavior to use when looking for this option's value.
   ///   - help: Information about how to use this option.
+  @available(*, deprecated, message: "Use init(wrappedValue:name:parsing:help:completion:) instead.")
   public init(
     wrappedValue: Value,
     name: NameSpecification = .long,
     parsing parsingStrategy: SingleValueParsingStrategy = .next,
-    completion: CompletionKind? = nil,
```
rauhul, line 145: This change isn't ABI stable if I recall correctly, but im not sure if that matters for this library. I wonder if you could leave types as Optional with a default of `nil` and use `@_disfavoredOverload` to prefer using your new initializer. 

@natecook1000 to comment on if this is necessary

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -138,12 +138,40 @@ extension Option where Value: ExpressibleByArgument {
...
+  /// This method is called to initialize an `Option` with a default value such as:
+  /// ```swift
+  /// @Option var foo: String = "bar"
+  /// ```
+  ///
+  /// - Parameters:
+  ///   - wrappedValue: A default value to use for this property, provided implicitly by the compiler during property wrapper initialization.
+  ///   - name: A specification for what names are allowed for this flag.
+  ///   - parsingStrategy: The behavior to use when looking for this option's value.
+  ///   - help: Information about how to use this option.
```
rauhul, line 168: Could you add a description of the `completion` parameter?

McNight: Sure!

McNight: I noticed that a lot of `completionKind` parameters from other methods are not documented. What should we do about those ?

Sources/ArgumentParser/Parsable Properties/Option.swift
McNight: I didn't know about `@_disfavoredOverload`. Indeed, it seems to do the trick (when re-establishing optionals).

Sources/ArgumentParser/Parsable Properties/Option.swift
rauhul: Probably worth doing in a separate pull request (if you're up to it)

natecook1000 (commented):
Thanks so much for taking care of this, @McNight! 🎉 

In addition to the couple of notes below, could you also add a test that uses the deprecated version of the `@Option` initializer? It could go in `DefaultsEndToEndTests.swift`.

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -138,12 +138,41 @@ extension Option where Value: ExpressibleByArgument {
   ///   - name: A specification for what names are allowed for this flag.
   ///   - parsingStrategy: The behavior to use when looking for this option's value.
   ///   - help: Information about how to use this option.
+  @available(*, deprecated, message: "Use init(wrappedValue:name:parsing:help:completion:) instead.")
```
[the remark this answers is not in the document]
McNight: That's great (I sometimes lack of inspiration when writing documentation or deprecation messages 😅)

McNight:
I added a test. Please feel free to criticize it. Also:

- I took inspiration from the `testParsing_OptionPropertyInit*` tests
- what do you think of the spot where I put it ?
- what do you think of the naming ?
- it obviously generates a warning. is it an issue ?

rauhul:
I _think_ you can silence the warning by making the test deprecated and moving the command definition into the test function.

natecook1000:
Yep! If you mark the struct and test as deprecated it should eliminate the warnings. Otherwise this looks ready! 👍🏻

McNight:
Both work fine; I just marked the struct and test as deprecated because I think it reads better, if that's ok for you @rauhul :)
(Next PR will be one to harmonize documentation regarding `completionKind` in initializers)

natecook1000 (commented):
Looks great! 🎉

natecook1000:
@swift-ci Please test

PR #397, opened 2022-01-31
Fix issue with unconditional remaining properties in subcommands

natecook1000:
This extends the current behavior of `@Argument` array properties defined with the `.unconditionalRemaining` parsing strategy to work with subcommands, as well. For example, with a command hierarchy like this:

```swift
@main
struct Root: ParsableCommand {
    static var configuration: CommandConfiguration {
        .init(subcommands: [Sub.self])
    }

    @Flag var verbose = false
}

struct Sub: ParsableCommand {
    @Argument(parsing: .unconditionalRemaining)
    var remainder: [String] = []

    @OptionGroup var root: Root

    func run() {
        print("Verbose:", root.verbose)
        print("Remainder:", remainder)
    }
}
```

We want the `Sub` command to capture any arguments after the subcommand name appears in the command-line arguments. This fix enables the following behavior:

```
$ root --verbose sub foo --bar
Verbose: true
Remainder: foo --bar
$ root sub foo --bar --verbose 
Verbose: false
Remainder: foo --bar --verbose
$ root --verbose sub foo --bar --verbose 
Verbose: true
Remainder: foo --bar --verbose
```

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -31,6 +31,8 @@ public protocol ParsableCommand: ParsableArguments {
   mutating func run() throws
 }
 
+// MARK: - Default implementations
+
 extension ParsableCommand {
   public static var _commandName: String {
     configuration.commandName ??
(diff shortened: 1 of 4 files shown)


natecook1000:
@swift-ci Please test

PR #404, opened 2022-02-11
Support an `async` entry point for commands

natecook1000:
### Description
Adds an `AsyncParsableCommand` type that sports an `run() async` method, allowing for asynchronous code in a command line tool.

### Detailed Design
This includes the new `AsyncParsableCommand` protocol, which provides a `static func main() async` entry point and can call through to the root command's or a subcommand's asynchronous `run()` method. For this asynchronous execution, the root command must conform to `AsyncParsableCommand`, but its subcommands can be a mix of asynchronous and synchronous commands.

```swift
/// A type that can be executed as part of a nested tree of commands.
public protocol AsyncParsableCommand: ParsableCommand {
  mutating func run() async throws
}

extension AsyncParsableCommand {
  public static func main() async { ... }
}
```

Due to an issue in Swift 5.5, you can only use `@main` on an `AsyncParsableCommand` root command starting in Swift 5.6. This PR also includes a workaround for clients that are using Swift 5.5. Declare a separate type that conforms to `AsyncMainProtocol` and add the `@main` attribute to that type. For example:

```swift
@main enum Main: AsyncMain {
    typealias Command = <#command#>
}
```

`AsyncMainProtocol` is deprecated for Swift 5.6.

### Documentation Plan
There's a new example included (`count-lines`) that uses `AsyncParsableCommand` to read lines asynchronously, demonstrating the feature. A new guide for creating asynchronous commands is still TK.

### Test Plan
As an entry point, this feature is tested by invoking the built executable via a new test in `ArgumentParserExampleTests`.

### Source Impact
This change requires an upgrade in Swift versions to 5.5, but is otherwise source compatible. 

### Checklist


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.2
```
jpsim, line 1: If this PR ends up changing the minimum supported Swift version from 5.2 to 5.5, that will be a breaking change for any consumers who support compiling with Swift < 5.5. Is the intention to ship this as part of a 2.0 release or is breaking Swift version support not considered a semver breaking change?

natecook1000: Good question, @jpsim! The plan is to ship this change as part of the 1.1.0 release. My understanding is that this isn't a source breaking change, since clients using older versions of Swift will just continue to get version 1.0.3. That said, I do have the wrong platform setting below — will update that. Thanks!

ffried: Is there a reason why we don't wrap the new stuff in `#if swift(>=5.5)`, and in turn keep support for older Swift versions?

rauhul: @ffried Not to say that you shouldn't, what are your use cases where you need to support older swift versions like 5.2?

ffried: @rauhul I don't have any myself. I was just wondering in case this is considered a breaking change.

jpsim: Usage of older Swift versions like 5.1/5.2 is higher than you might think ([15% of this survey's respondents](https://twitter.com/daveverwer/status/1489287264043773955)). So for library authors who want to continue supporting those users, all dependencies must be compatible with all supported versions.

My original question though was more about semantic versioning and whether changing the minimum supported Swift version constitutes a breaking change.

> clients using older versions of Swift will just continue to get version 1.0.3

I could be wrong, but I don't think that's how SwiftPM works. I suspect it'll see the `// swift-tools-version:5.5` comment and fail immediately if the swift tools version is older than that.

rauhul: > Usage of older Swift versions like 5.1/5.2 is higher than you might think ([15% of this survey's respondents](https://twitter.com/daveverwer/status/1489287264043773955)).

That might be a convincing enough reason to upgrade. IMO _we_ (the community + apple package owners) should come up with some guidelines for version support. I think NIO announced they will support only the latest 2 swift versions. (Though I might be mis-remembering). 

@Lukasa what is swift-nio's police for swift version support and do you consider dropping support for a swift version a breaking change?

Lukasa: We are proposing to support [the latest Swift version and the two versions prior](https://forums.swift.org/t/swiftnio-swift-version-support/53232). We do not consider dropping Swift versions to be a breaking change, because as @natecook1000 says, SwiftPM takes the tools version into account when resolving what version of a package will be used. Users silently get the last version that supported their Swift version.

jpsim: > SwiftPM takes the tools version into account when resolving what version of a package will be used. Users silently get the last version that supported their Swift version

This is a big surprise to me since it means SwiftPM needs to check out multiple versions of a dependency in order to identify which on is the last to be compatible with the current swift version. Until today I thought SwiftPM just looked at the git tags to determine which version of a dependency to resolve.

If this is the case, then I have no objections to versioning this with 1.1.0.

@rauhul it'd be great if there were guidelines for Swift version support for Swift library authors.

natecook1000:
@swift-ci Please test

Package.swift
rauhul: @jpsim This seems like a good conversation to start on the swift forums, I'm 100% certain people have opinions... (for better or worse)

natecook1000:
@swift-ci Please test

rauhul:
@swift-ci Please test

Package.swift
```diff
@@ -14,6 +14,7 @@ import PackageDescription
 
 var package = Package(
```
rauhul, line 15: @natecook1000 can we take this opportunity to make the indentation 2 spaces?

rauhul:
@swift-ci please test

natecook1000:
@swift-ci Please test macOS platform

Package.swift
natecook1000: Oh sure. It should be 4 in code samples and example code and 2 everywhere else.

Jerry-Carter:
Shifted our test platform to 5.6 for our async version.  Wanted to let you know that we've been able to considerably simplify our code around swift-argument-parser.  This is looking good, Nate.

natecook1000:
@swift-ci Please test

Blackjacx:
I refactored my App Store Connect API tool with using the async branch and integrating async/await using the async branch of the argument parser was sooooo satisfying. Good job everyone 👍🏼 hope this makes it in the official release soon. It is soo soo great :)

For the interested of you here are some links: [twitter.com/Blackjacxxx/status/1411730725393555456](https://twitter.com/Blackjacxxx/status/1411730725393555456)

natecook1000:
@swift-ci Please test

Package.swift
Lukasa: A fun irony here is that we discussed the tools version change, but I missed that this _also_ added a `platforms` stanza, and while changing the tools version isn't breaking, changing the platforms stanza very much is.

PR #412, opened 2022-02-15
Show hidden args/opts/flags with --help-hidden

natecook1000:
This includes arguments, options, and flags that have `.hidden` visibility when a user requests the extended version of help with `--help-hidden`. Previously, only hidden option groups were included.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Usage/HelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/HelpGenerator.swift
@@ -106,7 +106,8 @@ internal struct HelpGenerator {
     if let usage = currentCommand.configuration.usage {
       self.usage = usage
     } else {
-      var usage = UsageGenerator(toolName: toolName, definition: [currentArgSet]).synopsis
+      var usage = UsageGenerator(toolName: toolName, definition: [currentArgSet])
+        .synopsis(visibility: visibility)
       if !currentCommand.configuration.subcommands.isEmpty {
         if usage.last != " " { usage += " " }
         usage += "<subcommand>"
@@ -140,7 +141,8 @@ internal struct HelpGenerator {
     /// more elements at a time.
     var args = commandStack.argumentsForHelp(visibility: visibility)[...]
     while let arg = args.popFirst() {
-      guard arg.help.visibility == .default else { continue }
+      guard arg.help.visibility.isAtLeastAsVisible(as: visibility)
+        else { continue }
       
       let synopsis: String
       let description: String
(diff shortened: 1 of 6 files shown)


natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -140,7 +141,8 @@ internal struct HelpGenerator {
     /// more elements at a time.
     var args = commandStack.argumentsForHelp(visibility: visibility)[...]
     while let arg = args.popFirst() {
-      guard arg.help.visibility == .default else { continue }
+      guard arg.help.visibility.isAtLeastAsVisible(as: visibility)
```
rauhul, line 144: Shouldn't `var args = commandStack.argumentsForHelp(visibility: visibility)[...]` make this unnecessary?

Sources/ArgumentParser/Usage/UsageGenerator.swift
```diff
@@ -37,10 +37,10 @@ extension UsageGenerator {
   /// The tool synopsis.
   ///
   /// In `roff`.
-  var synopsis: String {
+  func synopsis(visibility: ArgumentVisibility = .default) -> String {
```
rauhul, line 40: Not defaulting this argument will improve correctness at the cost of verbosity. (I think its a worthwhile trade off to remove the default)

Sources/ArgumentParser/Usage/HelpGenerator.swift
rauhul: I was thinking about this, I don't think removing it will work because the visibility in ArugmentSetProvider isn't correct for non-OptionGroups. But maybe I'm wrong

natecook1000: I updated `ArgumentSet(_:visibility:)` so that it does the filtering, as you'd expect from the spelling.

natecook1000:
@swift-ci Please test

Tests/ArgumentParserUnitTests/UsageGenerationTests.swift
```diff
@@ -15,6 +15,17 @@ import XCTest
...
 
+func _testSynopsis<T: ParsableArguments>(
+  _ type: T.Type,
+  visibility: ArgumentVisibility = .default,
+  expected: String,
+  file: StaticString = #file,
+  line: UInt = #line
+) {
+  let help = UsageGenerator(toolName: "example", parsable: T(), visibility: visibility)
+  XCTAssertEqual(help.synopsis(), expected, file: file, line: line)
```
rauhul, line 26: `visibility` doesn't get passed to `help.synopsis` here. (this is why I don't like the default value)

natecook1000:
@swift-ci Please test macOS platform

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #413, opened 2022-02-15
Change ArgumentVisibility into a struct

rauhul:
- Changes ArgumentVisibility from an enum to a struct. This will allow
  ArgumentParser to add cases in the future without breaking clients
  that could have been exhaustively switching across all cases. It also
  allows us to implement protocol on the internal type that and not
  expose their conformance on the public type.


rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

Sources/ArgumentParser/Parsable Properties/ArgumentVisibility.swift
```diff
@@ -10,7 +10,22 @@
...
+    /// Only show help for this argument in the extended help screen.
+    public static let hidden = Self(rawValue: .hidden)
+
+    /// Never show help for this argument.
+    public static let `private` = Self(rawValue: .private)
+}
+
+/// Internal implementation of `ArgumentVisibility` to allow for easier API
+/// evolution
+internal enum _ArgumentVisibility {
```
natecook1000, line 28: Let's move this to be nested inside `ArgumentVisibility` and rename to match the other types that are set up like this, e.g. https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Parsable%20Properties/Flag.swift#L126-L132

Sources/ArgumentParser/Parsable Properties/ArgumentVisibility.swift
```diff
@@ -39,8 +52,7 @@ extension ArgumentVisibility {
     }
   }
 
-  /// - Returns: true if `self` is at least as visible as the supplied argument.
-  func isAtLeastAsVisible(as other: Self) -> Bool {
```
natecook1000, line 43: I actually prefer this API over the comparable conformance, especially since we eventually want to include per-group visibility, which won't be comparable in nature. A function call like this is more explicit about the relationship between the two visibilities, too. (I can kind of picture the ordering going in either direction.)

Sources/ArgumentParser/Parsable Properties/ArgumentVisibility.swift
rauhul: Should be fixed.

Sources/ArgumentParser/Parsable Properties/ArgumentVisibility.swift
rauhul: Should be fixed.

rauhul:
@swift-ci please test

natecook1000 (approved):

PR #422, opened 2022-03-15
Fix broken links/incorrect variance calculation

natecook1000:


natecook1000:
@swift-ci Please test

PR #427, opened 2022-03-16
Remove platform requirement from Package.swift

### LINKED ISSUE #426: Packages using ArgumentParser which support older platforms can no longer compiler (since 1.1.0)
Packages depending on swift-argument-parser which support older platforms are no longer able to compile.

This is a regression introduced in 1.1.0 where supported platforms were added to the package manifest. This raised the supported platforms from the defaults. This is a breaking change as packages depending on swift-argument-parser which support platforms older than those supported by this package are no longer able to compile (SPM will not resolve to a version of the package where appropriate platforms are supported).

Packages depending on swift-argument-parser must either raise their own supported platforms, require the version of swift-argument-parser to be less than 1.1.0, or remove their requirement on swift-argument-parser. 

**ArgumentParser version:** `1.1.0`

### Checklist

natecook1000:
Adding the platform requirement is a source breaking change; this moves the requirement down to the `async` symbols instead.

Fixes #426.

### DIFF
--- a/Examples/count-lines/CountLines.swift
+++ b/Examples/count-lines/CountLines.swift
@@ -13,6 +13,7 @@ import ArgumentParser
 import Foundation
 
 @main
+@available(macOS 10.15, *)
 struct CountLines: AsyncParsableCommand {
     @Argument(
         help: "A file to count lines in. If omitted, counts the lines of stdin.",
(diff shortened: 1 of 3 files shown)


natecook1000:
@swift-ci Please test

Lukasa (approved):
Nice, LGTM

Issue #429, opened 2022-03-20
testCountLines fails on my local environment


KS1019:
testCountLines fails when running tests on my environment. It seems like arguments are not properly recognized and the default run() method is executed instead of run() in CountLines.
 
**ArgumentParser version:**  main (1.1.1)
**Swift version:** 
```
Swift version: Apple Swift version 5.6 (swiftlang-5.6.0.323.62 clang-1316.0.20.8)
Target: x86_64-apple-macosx12.0
Unix version: Darwin Kotaro-no-MacBook-Pro.local 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022
 root:xnu-8019.80.24~20/RELEASE_X86_64 x86_64

macOS version: 12.2.1
Xcode-select path: '/Applications/Xcode-13.3.0.app/Contents/Developer
Xcode: Xcode 13.3 (Build version 13E113)

```

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
Run `swift test`

### Expected behavior
Tests run successfully

### Actual behavior

```shell
Test Suite 'CountLinesExampleTests' started at 2022-03-20 00:44:05.359
Test Case '-[ArgumentParserExampleTests.CountLinesExampleTests testCountLines]' started.
/Users/kotarosuto/Documents/oss/swift-argument-parser/Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift:21: error: -[ArgumentParserExampleTests.CountLinesExampleTests testCountLines] : XCTAssertEqual failed: ("1") is not equal to ("10") - Strings have different numbers of lines.
/Users/kotarosuto/Documents/oss/swift-argument-parser/Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift:21: error: -[ArgumentParserExampleTests.CountLinesExampleTests testCountLines] : XCTAssertEqual failed: ("20") is not equal to ("USAGE: count-lines <input-file> [--prefix <prefix>] [--verbose]")
/Users/kotarosuto/Documents/oss/swift-argument-parser/Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift:22: error: -[ArgumentParserExampleTests.CountLinesExampleTests testCountLines] : XCTAssertEqual failed: ("1") is not equal to ("10") - Strings have different numbers of lines.
[...]

PR #430, opened 2022-03-22
Replace `AsyncParsableCommand.main` protocol extension

christopherweems:
Building an executable project using `AsyncParsableCommand` with the Swift development branch toolchain broke some time after the 02-25-22a snapshot. No compile-time errors show, but an executable built with the 03-13-22 or 03-22-22 trunk toolchain ignores any provided arguments, prints the list of argument subcommands, and exits.

This seems to be a bug in the compiler, which now prefers the `ParsableCommand.main()` type method to the `AsyncParsableCommand.main()` override.

I've found replacing `AsyncParsableCommand.main()` with one defined on `ParsableCommand` with an `AsyncParsableCommand` self-requirement restores the original behavior without breaking source-compatibility for Swift 5.6 (and presumably earlier versions as well). I am unsure of any ABI implications arising from this change.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
@@ -24,7 +24,7 @@ public protocol AsyncParsableCommand: ParsableCommand {
 }
 
 @available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
-extension AsyncParsableCommand {
+extension ParsableCommand where Self : AsyncParsableCommand {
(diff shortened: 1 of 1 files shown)


natecook1000:
Wow, this sounds like the issue described in #429, which I haven't been able to reproduce — thanks for finding the issue and the resolution! This library doesn't promise ABI compatibility, and this should be a source-stable change.

natecook1000:
@swift-ci Please test

natecook1000:
I'm getting ambiguity errors with this change, unfortunately. I'll have a fix/solution for the overall problem in a future update.

Issue #433, opened 2022-03-25
Debug assert to catch AsyncParseableCommand hierarchy issues


rauhul:
It would be nice to add a code path in debug mode that asserts if a user has accidentally included an AsyncParseableCommand nested under a Sync one, as the async run entry point will not be called in this case.

KeithBird:
> It would be nice to add a code path in debug mode that asserts if a user has accidentally included an AsyncParseableCommand nested under a Sync one, as the async run entry point will not be called in this case.

typo: `AsyncParseableCommand` -> `AsyncParsableCommand` 😊

KeithBird:
Does the following code correctly express this issue?

```swift
struct Math: ParsableCommand {
  static var configuration = CommandConfiguration(
    subcommands: [Add.self]
  )

  struct Add: AsyncParsableCommand {
    @Argument() var values: [Int] = []
  }
}

// Error: `Add` subcommand can't run asynchronously in `Math`
```

MarcoEidinger:
@rauhul is the issue fixed with https://github.com/apple/swift-argument-parser/pull/436 ? Then issue can be closed

rauhul:
Yes, good catch, verified locally.

PR #436, opened 2022-03-27
Fix `AsyncParseableCommand` hierarchy

KeithBird:
The feature will be invalid when override `main()` in `ParseableCommand`. I did some testing but didn't commit because our unit tests caught errors with `parse()` as the entry, skipping `main()`.

### Checklist


KeithBird:
Related to the issue #433

Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
```diff
@@ -31,6 +31,10 @@ extension AsyncParsableCommand {
   /// Instead of calling this method directly, you can add `@main` to the root
   /// command for your command-line tool.
   public static func main() async {
+#if DEBUG
+    configuration.subcommands.forEach(checkAsycHierarchy)
+#endif
+
```
natecook1000, line 37: We don't need these checks in the async main() methods — if one of these is executing, the subcommands can be any mix of synchronous or asynchronous commands.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -164,4 +171,19 @@ extension ParsableCommand {
   internal static var defaultIncludesUnconditionalArguments: Bool {
     configuration.defaultSubcommand?.includesUnconditionalArguments == true
   }
+  
+#if DEBUG
+  @available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
+  internal static func checkAsycHierarchy(_ command: ParsableCommand.Type) {
```
natecook1000, line 177: Spelling nit on this:
```suggestion
  internal static func checkAsyncHierarchy(_ command: ParsableCommand.Type) {
```

natecook1000 (commented):
Thanks for implementing this, @KeithBird! 👏🏻 Notes below.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
KeithBird: Oops!

Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
KeithBird: I want to confirm, in the following hierarchy structure, will the innermost `AsyncCommand` also be executed asynchronously?
```
AsyncCommand { Command { AsyncCommand } }
```

rauhul:
Thanks for taking this :D

KeithBird:
Thanks for staying up so late :P

rauhul:
@swift-ci please test

KeithBird:
@swift-ci please test

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -164,4 +171,19 @@ extension ParsableCommand {
...
+#if DEBUG
+  @available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
+  internal static func checkAsyncHierarchy(_ command: ParsableCommand.Type) {
+    command.configuration.subcommands.forEach(checkAsyncHierarchy)
+    
+    if command is AsyncParsableCommand.Type { return }
+    
+    for sub in command.configuration.subcommands {
+      guard !(sub is AsyncParsableCommand.Type) else {
+        fatalError("`\(sub)` subcommand can't run asynchronously in `\(command)` command")
```
natecook1000, line 184: I don't think this is giving quite the right diagnostic — if the root and first-level subcommand are synchronous, and the second-level subcommand is async, it will point at the first-level subcommand instead of the root.

```swift
@main
struct Root: ParsableCommand {
  static var configuration: CommandConfiguration {
    .init(subcommands: [SubOne.self])
  }

  struct SubOne: ParsableCommand {
    static var configuration: CommandConfiguration {
      .init(subcommands: [SubTwo.self])
    }
    
    struct SubTwo: AsyncParsableCommand {}
  }
}
```

Gives this error even though `Root` is the problem, not `SubOne`:

```
Fatal error: `SubTwo` subcommand can't run asynchronously in `SubOne` command
```

Maybe you can do a recursive search for any `AsyncParseableCommand`, and then output the error message just from this root-level context.

natecook1000: Additionally, let's update that error message to include information about how to fix the problem. Since this is a configuration issue, it would be great if this could be as big and clear a message as possible. Let's borrow the format from `directlyInitializedError`:

```
fatalError("""

  --------------------------------------------------------------------
  Asynchronous subcommand of a synchronous root.

  The asynchronous command `\(sub)` is declared as a subcommand of the synchronous root command `\(command)`.
  
  With this configuration, your asynchronous `run()` method will not be called. To fix this issue, change `\(command)`'s `ParsableCommand` conformance to `AsyncParsableCommand`.
  --------------------------------------------------------------------

  """.wrapped(to: 70))
```

KeithBird: done

PR #440, opened 2022-04-07
Fix invalid link in docs

KeithBird:
### Checklist


Sources/ArgumentParser/Documentation.docc/ArgumentParser.md
```diff
@@ -90,3 +90,4 @@ and then either calls your `run()` method or exits with a useful message.
 
 - <doc:ManualParsing>
 - <doc:ExperimentalFeatures>
+- <doc:CommandsAndSubcommands>
```
natecook1000, line 93: This article was originally in the "Essentials" section before I moved the link into `ParsableCommand`. Did you have a hard time finding it? I'd rather restore it to its original spot, under "Getting Started".

Sources/ArgumentParser/Documentation.docc/Articles/ManualParsing.md
```diff
@@ -51,7 +51,7 @@ print(chosen.joined(separator: "\n"))
 
 Manually parsing commands is a little more complex than parsing a simple `ParsableArguments` type. The result of parsing from a tree of subcommands may be of a different type than the root of the tree, so the static ``ParsableCommand/parseAsRoot(_:)`` method returns a type-erased ``ParsableCommand``.
 
-Let's see how this works by using the `Math` command and subcommands defined in [Commands and Subcommands](./CommandsAndSubcommands.md). This time, instead of calling `Math.main()`, we'll call `Math.parseAsRoot()`, and switch over the result:
+Let's see how this works by using the `Math` command and subcommands defined in [Commands and Subcommands](https://apple.github.io/swift-argument-parser/documentation/argumentparser/commandsandsubcommands). This time, instead of calling `Math.main()`, we'll call `Math.parseAsRoot()`, and switch over the result:
```
natecook1000, line 54: Instead of using this markdown link, let's use the DocC style.
```suggestion
Let's see how this works by using the `Math` command and subcommands defined in <doc:CommandsAndSubcommands>. This time, instead of calling `Math.main()`, we'll call `Math.parseAsRoot()`, and switch over the result:
```

Sources/ArgumentParser/Documentation.docc/ArgumentParser.md
KeithBird: I think there should be a page that acts as a table of contents listing all the important articles. Also I don't think an article link can't be in two places at the same time. If all articles can be found in the `ArgumentParser` directory, `CommandsAndSubcommands` will likely be ignored.

KeithBird: I found `CustomizingCommandHelp` also in the `ParsableCommand`. I think it's hard to find an article that only appears in the documentation of a symbol. Because I usually read Quick Help by Option-clicking a symbol instead of searching this symbol in the global documentation. And all text introductions can be seen in Quick Help before the launch of docc. We only need to go deep if we need to understand the methods under it.

Sources/ArgumentParser/Documentation.docc/Articles/ManualParsing.md
KeithBird: looks better

Sources/ArgumentParser/Documentation.docc/ArgumentParser.md
natecook1000: Article links can be included in multiple places. I agree that we should have the articles all visible from the root-level page — let's add links to both of those to the "Essentials" section, as well.

KeithBird: Done

natecook1000:
@swift-ci Please test

Issue #450, opened 2022-06-14
[GSoC] Partial initialization for `ParsableCommand`


KeithBird:
Before this project we need to initialize the entire `ParsableCommand` by parsing a single line of command, otherwise we will require the user to re-enter the whole command. In interactive mode, we can initialize step by step in each conversation by prompting the user.

---

**This issue is posted here as a subtask of the [GSoC project](https://github.com/apple/swift-argument-parser/issues/449) for discussion and suggestions.**

natecook1000:
It would be great to see this implemented as an initial proof of concept. Going from the `Missing expected argument` error to an `ArgumentDefinition`, prompting for input (just using `print` and `readLine` is fine at this stage), and then attempting decoding again would let us see that we have the pieces in place to work on the interface.

KeithBird:
Sorry, I clicked the wrong button, please assign me again.

KeithBird:
I am currently working on the bug that arrays (such as `@Argument var phrase: [String]`) couldn't be parsed as supplemental input.

PR #451, opened 2022-06-21
[GSoC] Partial initialization for ParsableCommand

KeithBird:
Solution to the issue #450.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
@@ -31,17 +31,8 @@ extension AsyncParsableCommand {
   /// Instead of calling this method directly, you can add `@main` to the root
   /// command for your command-line tool.
   public static func main() async {
-    await self.main(lines: nil)
-  }
-  
-  /// Executes this command, or one of its subcommands, with the program's
-  /// command-line arguments.
-  ///
-  /// Instead of calling this method directly, you can add `@main` to the root
-  /// command for your command-line tool.
-  public static func main(lines: [String]?) async {
     do {
-      var command = try parseAsRoot(nil, lines: lines)
+      var command = try parseAsRoot()
       if var asyncCommand = command as? AsyncParsableCommand {
         try await asyncCommand.run()
       } else {
@@ -70,14 +61,8 @@ extension AsyncMainProtocol {
   /// Executes the designated command type, or one of its subcommands, with
   /// the program's command-line arguments.
   public static func main() async {
-    await self.main(lines: nil)
-  }
-  
-  /// Executes the designated command type, or one of its subcommands, with
-  /// the program's command-line arguments.
-  public static func main(lines: [String]?) async {
     do {
-      var command = try Command.parseAsRoot(nil, lines: lines)
+      var command = try Command.parseAsRoot()
       if var asyncCommand = command as? AsyncParsableCommand {
         try await asyncCommand.run()
       } else {
--- a/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
@@ -31,8 +31,17 @@ extension AsyncParsableCommand {
   /// Instead of calling this method directly, you can add `@main` to the root
   /// command for your command-line tool.
   public static func main() async {
+    await self.main(lines: nil)
+  }
+  
+  /// Executes this command, or one of its subcommands, with the program's
+  /// command-line arguments.
+  ///
+  /// Instead of calling this method directly, you can add `@main` to the root
+  /// command for your command-line tool.
+  public static func main(lines: [String]?) async {
     do {
-      var command = try parseAsRoot()
+      var command = try parseAsRoot(nil, lines: lines)
       if var asyncCommand = command as? AsyncParsableCommand {
         try await asyncCommand.run()
       } else {
@@ -61,8 +70,14 @@ extension AsyncMainProtocol {
   /// Executes the designated command type, or one of its subcommands, with
   /// the program's command-line arguments.
   public static func main() async {
+    await self.main(lines: nil)
+  }
+  
+  /// Executes the designated command type, or one of its subcommands, with
+  /// the program's command-line arguments.
+  public static func main(lines: [String]?) async {
     do {
-      var command = try Command.parseAsRoot()
+      var command = try Command.parseAsRoot(nil, lines: lines)
       if var asyncCommand = command as? AsyncParsableCommand {
         try await asyncCommand.run()
       } else {
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -60,10 +60,9 @@ extension ParsableCommand {
   /// - Returns: A new instance of this type, one of its subcommands, or a
   ///   command type internal to the `ArgumentParser` library.
   public static func parseAsRoot(
-    _ arguments: [String]? = nil,
-    lines: [String]? = nil
+    _ arguments: [String]? = nil
   ) throws -> ParsableCommand {
-    var parser = CommandParser(self, lines: lines)
+    var parser = CommandParser(self)
     let arguments = arguments ?? Array(CommandLine.arguments.dropFirst())
     return try parser.parse(arguments: arguments).get()
   }
--- a/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
@@ -60,11 +60,12 @@ extension ParsableCommand {
   /// - Returns: A new instance of this type, one of its subcommands, or a
   ///   command type internal to the `ArgumentParser` library.
   public static func parseAsRoot(
-    _ arguments: [String]? = nil
+    _ arguments: [String]?,
+    lines: [String]? = nil
   ) throws -> ParsableCommand {
     var parser = CommandParser(self)
     let arguments = arguments ?? Array(CommandLine.arguments.dropFirst())
-    return try parser.parse(arguments: arguments).get()
+    return try parser.parse(arguments: arguments, lines: lines).get()
   }
   
   /// Returns the text of the help screen for the given subcommand of this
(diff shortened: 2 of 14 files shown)


Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+//
+//  Created by Keith  on 2022/6/21.
+//
+
+extension CommandParser {
+  func canInteract(error: Error, split: inout SplitArguments) -> Bool {
+    if let error = error as? ParserError {
+      switch error {
+      case let .missingValueForOption(inputOrigin, name):
```
natecook1000, line 12: Can we use `guard` for a requirement like this? You can also use `case let` pattern matching in an `if` or `guard` statement instead of a `switch` when you're just looking for one case.

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+    
+    return false
+  }
+  
+  func canInteract(error: Error, values: inout ParsedValues) -> Bool {
+    if let error = error as? ParserError {
+      switch error {
+      case let .noValue(forKey: key):
```
natecook1000, line 35: Same note as above for this method 👍🏻

Sources/ArgumentParser/Parsing/InputOrigin.swift
```diff
@@ -29,16 +29,20 @@
...
   enum Element: Comparable, Hashable {
-    /// The input value came from a property's default value, not from a
-    /// command line argument.
+    /// The input value came from a property's default value,
+    /// not from a command line argument.
     case defaultValue
     
+    /// The input value came from interactive mode,
+    /// not from the first command typed by the user
+    case interactive
```
natecook1000, line 38: 👍🏻

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -142,35 +142,48 @@ extension CommandParser {
...
-    let decoder = ArgumentDecoder(values: values, previouslyDecoded: decodedArguments)
-    var decodedResult: ParsableCommand
-    do {
-      decodedResult = try currentNode.element.init(from: decoder)
-    } catch let error {
-      // If decoding this command failed, see if they were asking for
-      // help before propagating that parsing failure.
-      try checkForBuiltInFlags(split)
-      throw error
+    var values: ParsedValues?
```
natecook1000, line 145: Let's see if we can avoid making `values`, `decodedResult`, and `decoder` optional here — is that easier to manage if you refactor these loops out into functions?

Sources/ArgumentParser/Parsing/InputOrigin.swift
```diff
@@ -109,15 +113,15 @@ extension InputOrigin {
   }
 }
 
-extension InputOrigin.Element {
```
natecook1000, line 112: If this is still compiling, this seems like leftover code that we aren't using anymore. Feel free to delete!

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,58 @@
...
+      case let .missingValueForOption(inputOrigin, name):
+        var input: String?
+        while input?.isEmpty ?? true {
+          print("Please enter value for '\(name.synopsisString)': ", terminator: "")
+          input = readLine() ?? nil
+        }
+        
+        let index = inputOrigin.elements.first!.baseIndex! + 1
+        split._elements.append(.init(value: .value(input!),
+                                     index: .init(inputIndex: .init(rawValue: index))))
```
natecook1000, line 21: Does appending the new value work if the option with the missing value is in the middle? For example:

```
repeat hello --count --include-counter
```

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,58 @@
...
+        guard label != "generateCompletionScript" else { break }
+        
+        var input: String?
+        while input?.isEmpty ?? true {
+          print("Please enter '\(label)': ", terminator: "")
+          input = readLine() ?? nil
+        }
+        
+        var element = values.elements[key]!
+        element.value = (input as Any)
```
natecook1000, line 46: `ParsedValues` stores its values after they've been parsed out to the correct type. For example, the value for `@Option var count: Int` is converted to an `Int` before it's stored here. We'll need to do that transformation (which is stored in `ArgumentDefinition.update`) before assigning to `element.value`.

natecook1000 (commented):
This is a great start, @KeithBird! Some notes below — it looks like one missing piece is retrieving the correct `ArgumentDefinition` when you get a no-value error. That will let you do the required transformation before storing the new value you received from the user.

Sources/ArgumentParser/Parsing/Interactive.swift
KeithBird: done

Sources/ArgumentParser/Parsing/Interactive.swift
KeithBird: There may be more than one case here in the future, so let's keep the `switch` for now.

Sources/ArgumentParser/Parsing/InputOrigin.swift
KeithBird: done

Sources/ArgumentParser/Parsing/CommandParser.swift
KeithBird: done

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,59 @@
...
+    let inputIndex = inputOrigin.elements.first!.baseIndex! + 1
+    split._elements.insert(.init(value: .value(input!),
+                                 index: .init(inputIndex: .init(rawValue: inputIndex))),
+                           at: inputIndex)
+    
+    for index in (inputIndex + 1) ..< split.count {
+      split._elements[index].index = .init(inputIndex: .init(rawValue: index))
+    }
+    
+    split.originalInput.insert(input!, at: inputIndex)
```
KeithBird, line 28: If we don't change `originalInput` here, even if we get the correct second input, `repeat --count --include-counter hello` will crash below:

https://github.com/apple/swift-argument-parser/blob/c61fc0f7f0b10fbbdbdc7e0b31cf1b7f8fd35ea0/Sources/ArgumentParser/Parsing/SplitArguments.swift#L271

natecook1000:
I think this could make more sense if there's only one path for interactively collecting additional input. That should be possible if instead of throwing that `.missingValueForOption` error during parsing, we defer that error by storing it in the `ParsedValues` and throwing the error during decoding. That way, we won't be trying to re-write the original input, which is surely going to be more error prone than operating on the more structured parsed value data. The first `canInteract` method would go away with this approach. What do you think of that plan?

You may need to add additional information to the errors in order to be able to recover properly. In particular, you'll need access to the `ArgumentDefinition` in your `canInteract` method.

KeithBird:
Although we defer error handling, we still can't skip modifying the `SplitArguments` because its scope escapes the `parseCurrent()`.

And since the two `canInteract()` one modifies the `SplitArguments` and the other modifies the `ParsedValues`, I don't think it's a good idea to merge them.

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  File.swift
+//
+//
+//  Created by Keith  on 2022/6/21.
+//
```
natecook1000, line 6: Let's get the correct header comment on this file.

natecook1000 (commented):
This is looking good! I think we can come back to the split arguments question. Last thing I'd like to see here is a document explaining what's implemented and how to test it, since we don't have a test harness for this yet. It can just be a root-level Interactivity.md file.

Sources/ArgumentParser/Parsing/Interactive.swift
KeithBird: done

KeithBird:
I am currently working on the bug that arrays (such as `@Argument var values: [Int]`) can't be parsed as supplemental input. Or, should I defer this work after creating a PR about `@Argument`? @natecook1000

natecook1000 (approved):
LGTM! 🚀 Okay to merge this while you work on the array issue - let's just open an issue to track that work.

Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
```diff
@@ -31,8 +31,17 @@ extension AsyncParsableCommand {
...
   public static func main() async {
+    await self.main(lines: nil)
+  }
+  
+  /// Executes this command, or one of its subcommands, with the program's
+  /// command-line arguments.
+  ///
+  /// Instead of calling this method directly, you can add `@main` to the root
+  /// command for your command-line tool.
+  public static func main(lines: [String]?) async {
```
natecook1000, line 42: It's a great idea to have a way to test the incremental parsing/initialization that you're adding, but we'll need to find a way to do it without affecting the public `ParsableCommand` API.

natecook1000 (changes requested):
The tests are an important addition here! Since this is all internal machinery, I think it's fine to provide an internal hook for testing this, with the tests creating and calling the `CommandParser` methods directly. You could add `lines` as a `CommandParser` property, with an internal initializer that lets you inject the lines to test.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -60,9 +60,10 @@ extension ParsableCommand {
   /// - Returns: A new instance of this type, one of its subcommands, or a
   ///   command type internal to the `ArgumentParser` library.
   public static func parseAsRoot(
-    _ arguments: [String]? = nil
+    _ arguments: [String]? = nil,
+    lines: [String]? = nil
```
natecook1000, line 64: We still need to revert this change to the public API.

Tests/ArgumentParserInteractiveTests/ArgumentInteractiveTests.swift
```diff
@@ -0,0 +1,82 @@
...
+// This source file is part of the Swift Argument Parser open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
```
natecook1000, line 12: If you change this to a `@testable import`, you can make a version of the `AssertParseCommand` function that creates a `CommandParser` directly, instead of going through public API.

natecook1000 (approved):
LGTM! 🚀

Issue #459, opened 2022-07-22
[GSoC] Asking to retype invalid values


KeithBird:
### Asking to retype an invalid value

When parameters similar to the following are not initialized:

```swift
@Argument var integer: Int
```

The following dialog will be generated automatically:

```
? Please enter 'integer': a
Error: The value 'a' is invalid for '<integer>'.

? Please replace 'a': 0.1
Error: The value '0.1' is invalid for '<integer>'.

? Please replace '0.1': 1
```

### Asking to retype invalid values in array

When parameters similar to the following are not initialized:

```swift
@Argument var values: [Int]
```

The following dialog will be generated automatically:

```
Please enter 'values': a 0.1
Error: The value 'a' is invalid for '<values>'.

? Please replace 'a': 1
Error: The value '0.1' is invalid for '<values>'.

? Please replace '0.1': 2
```

---

**This issue is posted here as a subtask of the [GSoC project](https://github.com/apple/swift-argument-parser/issues/449) for discussion and suggestions.**

PR #460, opened 2022-07-25
[GSoC] Asking to retype invalid values

KeithBird:
Solution to the issue #459.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/Interactive.swift
+++ b/Sources/ArgumentParser/Parsing/Interactive.swift
@@ -68,7 +68,7 @@ extension CommandParser {
         storeNormalValues(label: label, updateBy: updateBy, arguments: arguments)
       } else {
         allValues.enumerated().forEach { print("\($0 + 1). \($1)") }
-        storeCaseIterableEnums(label: label, allValues: allValues, updateBy: updateBy, arguments: arguments)
+        storeCaseIterableEnums(label: label, allValues: allValues, updateBy: updateBy)
       }
 
       return true
@@ -92,10 +92,12 @@ extension CommandParser {
         try updateBy(str)
       } catch {
         // Handle ParserError
-        guard let error = error as? ParserError else { return }
+        guard
+          let error = error as? ParserError,
+          case let .unableToParseValue(_, _, original, _, _) = error
+        else { return }
         let generator = ErrorMessageGenerator(arguments: arguments, error: error)
         let description = generator.makeErrorMessage() ?? error.localizedDescription
-        guard case let .unableToParseValue(_, _, original, _, _) = error else { return }
         print("Error: " + description + ".\n")
         replaceInvalidValue(original: original, updateBy: updateBy, arguments: arguments)
       }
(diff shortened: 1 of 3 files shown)


Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -57,51 +57,91 @@ extension CommandParser {
...
+    let input = getInput() ?? ""
+
+    do {
+      try updateBy(input)
+    } catch {
+      // Handle ParserError
+      guard let error = error as? ParserError else { return }
+      let generator = ErrorMessageGenerator(arguments: arguments, error: error)
+      let description = generator.makeErrorMessage() ?? error.localizedDescription
+      guard case let .unableToParseValue(_, _, original, _, _) = error else { return }
```
natecook1000, line 118: The code will be clearer if we group the two preconditions together. Since they have the same failure outcome (an early `return`), you can combine them into a single `guard` statement, as well.

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -57,51 +57,91 @@ extension CommandParser {
...
+      print("Error: " + description + ".\n")
+      replaceInvalidValue(original: original, updateBy: updateBy, arguments: arguments)
+    }
+  }
+
+  fileprivate mutating func storeCaseIterableEnums(label: String,
+                                                   allValues: [String],
+                                                   updateBy: (String) throws -> Void,
+                                                   arguments: ArgumentSet)
```
natecook1000, line 127: Is `arguments` used in this function?

Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -57,51 +57,91 @@ extension CommandParser {
...
-        return getCaseIterableEnum(label: label, allValues: allValues)
+    for num in strs {
+      guard let index = Int(num) else {
+        print("Error: '\(num)' is not a serial number.\n")
+        storeCaseIterableEnums(label: label, allValues: allValues, updateBy: updateBy, arguments: arguments)
+        return
       }
 
       guard range.contains(index) else {
         print("Error: '\(index)' is not in the range of \(range) \n")
```
natecook1000, line 142: We want to avoid showing programming concepts like `3...6` in error messages. Maybe something like this, or maybe two different messages for too low and too high? For example, `'x' must be at least '1'` and `'x' must be at most '7'`. What do you think?
```suggestion
        print("Error: '\(index)' is not in the range '\(range.lowerBound) - \(range.upperBound)'.\n")
```

natecook1000 (commented):
@KeithBird This is looking good! A few notes below.

Sources/ArgumentParser/Parsing/Interactive.swift
KeithBird: "is not in the range 3 - 6." provides more complete information.

natecook1000 (approved):
LGTM!

Issue #462, opened 2022-07-27
[GSoC] Add API for disabling interactive mode


natecook1000:
There should be API added to the `CommandConfiguration` type so that command authors can disable interactive mode. We'll need to choose a good name — something like `promptForMissingRequiredValues: Bool = true` is accurate but obviously too long.

KeithBird:
Is the issue fixed with #464? @natecook1000

PR #464, opened 2022-08-08
[GSoC] Add API for disabling interactive mode

KeithBird:
Solution to the issue #462.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
+++ b/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
@@ -47,7 +47,7 @@ public struct CommandConfiguration {
   public var shouldDisplay: Bool
 
   /// A Boolean value indicating whether to prompt for missing required values.
-  public var isInteractable: Bool
+  public var shouldPromptForMissing: Bool
   
   /// An array of the types that define subcommands for this command.
   public var subcommands: [ParsableCommand.Type]
@@ -90,7 +90,7 @@ public struct CommandConfiguration {
     discussion: String = "",
     version: String = "",
     shouldDisplay: Bool = true,
-    isInteractable: Bool = false,
+    shouldPromptForMissing: Bool = true,
     subcommands: [ParsableCommand.Type] = [],
     defaultSubcommand: ParsableCommand.Type? = nil,
     helpNames: NameSpecification? = nil
@@ -101,7 +101,7 @@ public struct CommandConfiguration {
     self.discussion = discussion
     self.version = version
     self.shouldDisplay = shouldDisplay
-    self.isInteractable = isInteractable
+    self.shouldPromptForMissing = shouldPromptForMissing
     self.subcommands = subcommands
     self.defaultSubcommand = defaultSubcommand
     self.helpNames = helpNames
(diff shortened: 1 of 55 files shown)


KeithBird:
https://github.com/apple/swift-argument-parser/blob/df9ee6676cd5b3bf5b330ec7568a5644f547201b/Sources/ArgumentParser/Parsable%20Types/ParsableArguments.swift#L77

When `ParsableArguments` is converted to `ParsableCommand`, it is difficult to set the configuration parameters, but testing requires turning off interactive mode. So I've set the interaction mode to turn off by default for now. @natecook1000

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -45,6 +45,9 @@ public struct CommandConfiguration {
   /// A Boolean value indicating whether this command should be shown in
   /// the extended help display.
   public var shouldDisplay: Bool
+
+  /// A Boolean value indicating whether to prompt for missing required values.
+  public var isInteractable: Bool
```
natecook1000, line 50: I don't think we're quite there yet with `isInteractable` — what other names have you considered? This is a fairly small API, especially if changing the default will hopefully be rare, so maybe something longer like `shouldPromptForMissing`?

natecook1000 (commented):
Looks good so far! Can you change the default to true if you make `_WrappedParsableCommand` have it disabled?

Sources/ArgumentParser/Parsable Types/ParsableArguments.swift
```diff
@@ -60,6 +60,28 @@ struct _WrappedParsableCommand<P: ParsableArguments>: ParsableCommand {
   @OptionGroup var options: P
 }
 
+/// A type that provides the `ParsableCommand` interface to a `ParsableArguments` type,
+/// and set the `shouldPromptForMissing` to true.
+struct _WrappedThrowableCommand<P: ParsableArguments>: ParsableCommand {
```
natecook1000, line 65: Why are we using a new type here instead of modifying `_WrappedParsableCommand`? Is the old type still used anywhere?

natecook1000 (commented):
One question about the new wrapper type for you. Let's also try to watch out for whitespace-only changes - makes it a bit harder to review. 👍

KeithBird:
> One question about the new wrapper type for you. Let's also try to watch out for whitespace-only changes - makes it a bit harder to review. 👍

My apologies. You can turn on [Hide whitespace](https://cloudfour.com/thinks/quick-tip-how-to-hide-whitespace-changes-in-git-diffs/#github) which is hard to find. Can we manage formats automatically using [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) and some configuration files? Manual formatting is a little tricky when there are a lot of files that have been slightly modified.

natecook1000 (approved):
LGTM!

natecook1000:
I'd forgotten about that setting, thanks! I need to think about adding a dependency - it would be nice to have the formatting handled.

KeithBird:
I think the [TCA](https://github.com/pointfreeco/swift-composable-architecture/blob/main/.github/workflows/format.yml) has done a good job of this. Please let me know if I can help. @natecook1000

PR #471, opened 2022-08-25
Fix generate-manual plugin authors argument

rauhul:
- Fixes an issue where the --author option without its value  was passed
  by the generate-manual plugin to the generate-manual tool resulting in
  plugin invocation failures or incorrect author information.


rauhul:
@swift-ci please test

natecook1000 (approved):

PR #473, opened 2022-08-25
Fix missing option value names in manuals

rauhul:
- Fixes a manual generation bug where options in the synopsis were not
  given an option value name and flags were.


rauhul:
@swift-ci please test

natecook1000 (approved):

rauhul:
@swift-ci please test

PR #475, opened 2022-08-25
Drop experimenal from generate-manual

rauhul:
- Removes the experimental prefix from the generate-manual plugin. The
  manual generation still is not mature, but we would like to avoid
  breaking adopters scripts by delaying changing the plugin's name in
  the future.


rauhul:
@swift-ci please test

natecook1000 (approved):
🎉

rauhul:
@swift-ci please test

PR #477, opened 2022-08-26
Unify @Argument and @Option initialization paths

### LINKED ISSUE #466: Including `transform:` parameter in `@Argument` declaration generates wrong help message for optional arguments
Declaring an argument with `@Argument(..., transform:)` makes it non-optional in the "USAGE" and doesn't include it's default value in the argument's abstract in the "ARGUMENTS" section, ***even* if the value has a default value and/or is of an optional type.**  
(The actual parsing of the optional arguments seems to be correct)

**ArgumentParser version:** `1.1.3`
**Swift version:** `swift-driver version: 1.26.9 Apple Swift version 5.5.1 (swiftlang-1300.0.31.4 clang-1300.0.29.6)`

### Checklist

### Steps to Reproduce
Declare arguments like this in a `ParsableCommand` struct:
```
struct Cmd1: ParsableCommand {
  @Argument(help: "Test argument 1", transform: { Int($0)! })
  var test1: Int = 0
  
  @Argument(help: "Test argument 2", transform: URL.init(fileURLWithPath:))
  var test2: URL?
}
```
and print the help message through `Cmd.parseOrExit([])` or `print(Cmd.helpMessage())`

### Expected behavior
The printed help string (namely USAGE and ARGUMENTS) should be the same as if the arguments were declared like this (without the `transform:` parameter):
```
struct Cmd2: ParsableCommand {
  @Argument(help: "Test argument 1")
  var test1: Int = 0
  
  @Argument(help: "Test argument 2")
  var test2: String?
}
```

### Actual behavior
Help message for arguments with `transform:` parameter looks like this:
```
USAGE: cmd1 <test1> <test2>

ARGUMENTS:
  <test1>                 Test argument 1
  <test2>                 Test argument 2
```
While help message *without* `transform:` is correctly generated:
```
USAGE: cmd2 [<test1>] [<test2>]

ARGUMENTS:
  <test1>                 Test argument 1 (default: 0)
  <test2>                 Test argument 2
```

rauhul:
- Fixes #466.
- Adds initializers to ArgumentDefinition generic over a Container type.
  The Container type must conform to a new internal protocol
  ArgumentDefinitionContainer which describes functionality like default
  set of help options for the argument defined by the property wrapper,
  etc.
- Adds overloads for Optional @Arguments and @Options with default
  values which emit deprecation warning to guide users towards using the
  non-Optional versions.

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -322,7 +322,7 @@ extension Argument {
     Optional @Arguments with default values should be declared as non-Optional.
     """)
   public init<T>(
-    wrappedValue: Value,
+    wrappedValue: Optional<T>,
     help: ArgumentHelp? = nil,
     completion: CompletionKind? = nil
   ) where T: ExpressibleByArgument, Value == Optional<T> {
@@ -403,7 +403,7 @@ extension Argument {
     Optional @Arguments with default values should be declared as non-Optional.
     """)
   public init<T>(
-    wrappedValue: Value,
+    wrappedValue: Optional<T>,
     help: ArgumentHelp? = nil,
     completion: CompletionKind? = nil,
     transform: @escaping (String) throws -> T
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -94,66 +94,7 @@ extension Argument: CustomStringConvertible {
   }
 }
 
-extension Argument: DecodableParsedWrapper where Value: Decodable {}
-
-// MARK: Property Wrapper Initializers
-
-extension Argument where Value: ExpressibleByArgument {
-  /// Creates a property with an optional default value, intended to be called by other constructors to centralize logic.
-  ///
-  /// This private `init` allows us to expose multiple other similar constructors to allow for standard default property initialization while reducing code duplication.
-  private init(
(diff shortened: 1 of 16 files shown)


rauhul:
@swift-ci please test

natecook1000 (commented):
This looks great — thanks in particular for the extra test coverage 👏🏻👏🏻👏🏻

I don't feel like we have a great way to know exactly whether APIs are being introduced / removed here. I ran the breaking changes diagnostic, and after stripping some spurious errors (it doesn't think that changing the name a generic parameter is okay), I'm still seeing these four breaks — could you take a look to see if these are real or not?
```
  💔 API breakage: constructor Argument.init(wrappedValue:parsing:help:completion:) has parameter 0 type change from Value to [Element]
  💔 API breakage: constructor Argument.init(wrappedValue:parsing:help:completion:transform:) has parameter 0 type change from Value to [Element]
  💔 API breakage: constructor Option.init(wrappedValue:name:parsing:help:completion:) has parameter 0 type change from [Element] to Value
  💔 API breakage: constructor Option.init(wrappedValue:name:parsing:help:completion:transform:) has parameter 0 type change from [Element] to Value
```

rauhul:
@natecook1000 those are all spurious AFAICT, I reran locally to verify again.

rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

natecook1000 (approved):
🚢

Issue #159, opened 2020-05-20
Allow configuration discussion to be moved to end


erica:
I know this one is a little tenuous but I thought I'd throw it out there and if you don't like it, it can be as easily closed as any other feature request:

A flag that allows the discussion text (`CommandConfiguration(discussion: "")`) to be displayed at the end of help instead of the beginning so the options are visible to the user in the first screen of `more`/`less`. The current output is heavily weighted towards the style of internal code comments instead of the tooling needs of the end-user, who often wants to review the flags, options, arguments (and okay subcommands) quickly, and then use the tool immediately. This is the pattern used for memory refresh.

Reading the full discussion is more typical for someone who is encountering an issue with the tooling and looking for deeper instruction material, and is more typically secondary to casual use. (see also: "full help")

natecook1000:
Agreed on this, but I think I might want to have three slots available, so it ends up matching the way `man` pages are often layed out:

```
$ example --help
Usage: example --foo [bar]
Abstract
Discussion
Argument/flag/option descriptions
Additional info
```

@erica Do you have any suggestions for a name for “Additional info” here?

natecook1000:
Name for this extra configuration property can be `extendedDiscussion`.

arthurcro:
Hey! :wave: Are you still interested in closing issue? I'd love to give this a try!

natecook1000:
That would be great, thanks @arthurcro!

PR #479, opened 2022-08-29
IDEA: Supplemental discussion

rauhul:
- IDEA: additional discussion for commands and arguments which is
  exported to dump-help and can be consumed by supplemental content
  generators to contains much more detailed information than you may want
  to include in a help screen.
  Maybe a better idea would be to have --help-detailed flag that would
  include this information in the help output. Suggestions are welcome.

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
+++ b/Sources/ArgumentParser/Parsable Properties/ArgumentHelp.swift
@@ -16,7 +16,11 @@ public struct ArgumentHelp {
   
   /// An expanded description of the argument, in plain text form.
   public var discussion: String = ""
-  
+
+  /// Additional description of the argument for supplemental content, in plain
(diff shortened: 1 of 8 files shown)


rauhul:
I'm kinda liking `--help-detailed` (`--detailed-help`) and renaming this property `detailedDiscussion`.

natecook1000:
I think `--help-detailed` makes a lot of sense! We could/should also detect when that's a reasonable option and add it to the help screen (along with `--help-hidden` when there are hidden args).

arthurcro:
Hey! :wave: Is this skech pull request somehow related to https://github.com/apple/swift-argument-parser/issues/159?

rauhul:
> Hey! 👋 Is this skech pull request somehow related to #159?

Yep this would close that issue once fleshed out and merged.

PR #480, opened 2022-08-30
Allow default `nil` values for optional properties

natecook1000:
This adds underscored initializers that let library users add `= nil` to declarations of optional `@Option` and `@Argument` properties. Previously, default values have been available for properties of non-optional types only.

These new initializers use `_OptionalNilComparisonType` as the wrapped value parameter, so only a `nil` literal is acceptable in the default value position. This avoids the problem of declaring an optional property with a non-`nil` default, which ends up negating the purpose of an optional.

### Checklist


natecook1000:
@swift-ci Please test

rauhul:
Just wondering, why is this desired? Are we expecting an upcoming language mode where `var x: T?` doesn't have a default value of `nil`? (I would love that change)

Issue #461, opened 2022-07-27
[GSoC] Interactive mode should only run in an interactive shell


natecook1000:
Tests are currently blocked on the `feature/interactive` branch because the interactive mode is always on. It should only run when the command is executed in an interactive shell.

numist:
This was fixed by #485, yes?

PR #485, opened 2022-09-08
[GSoC] Interactive mode should only run in an interactive shell

KeithBird:
Solution to the issue #461.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/Interactive.swift
+++ b/Sources/ArgumentParser/Parsing/Interactive.swift
@@ -9,10 +9,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if os(Linux)
+#if canImport(Glibc)
 import Glibc
-#else
+#elseif canImport(Darwin)
 import Darwin
+#elseif canImport(CRT)
+import CRT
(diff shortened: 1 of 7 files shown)


Sources/ArgumentParser/Parsing/Interactive.swift
```diff
@@ -9,6 +9,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(Linux)
+import Glibc
+#else
+import Darwin
+#endif
```
natecook1000, line 16: Let's use `canImport` for these instead of `os` checks — you can see the other files that import `Glibc`.

natecook1000 (commented):
This looks good! If we remove the `shouldPromptForMissing: false` from the example commands, those tests will verify that this is working properly. To fully test that this is enabled when appropriate we'll need to emulate an interactive shell.

Sources/ArgumentParser/Parsing/Interactive.swift
KeithBird: done

natecook1000 (approved):

PR #492, opened 2022-09-17
Add API for titling an option group

### LINKED ISSUE #267: Group and label option groups in the help screen

natecook1000:
This change lets you provide a title for option groups, which is used when generating the help screen. Titled option groups, when they exist, are placed between the ARGUMENTS and OPTIONS section of the help. Multiple option groups with the same title are coalesced into a single group.

For example, this command declaration:

```swift
struct Extras: ParsableArguments {
    @Flag(help: "Print extra output while processing.")
    var verbose: Bool = false

    @Flag(help: "Include details no one asked for.")
    var oversharing: Bool = false
}

@main
    @OptionGroup(title: "Extras")
    var extras: Extras

    @Argument var name: String?
    @Option var title: String?
}
```

yields this help screen:

    USAGE: example [--verbose] [--oversharing] [<name>] [--title <title>]

    ARGUMENTS:
      <name>

    EXTRAS:
      --verbose               Print extra output while processing.
      --oversharing           Include details no one asked for.

    OPTIONS:
      --title <title>

Resolves #267.

### Checklist
    - [x] API documentation
    - [x] Articles

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
+++ b/Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
@@ -36,6 +36,9 @@ public struct OptionGroup<Value: ParsableArguments>: Decodable, ParsedWrapper {
   // FIXME: Adding this property works around the crasher described in
   // https://github.com/apple/swift-argument-parser/issues/338
   internal var _dummy: Bool = false
+  
+  /// The title to use in the help screen for this option group.
+  public var title: String = ""
 
   internal init(_parsedValue: Parsed<Value>) {
     self._parsedValue = _parsedValue
@@ -62,12 +65,27 @@ public struct OptionGroup<Value: ParsableArguments>: Decodable, ParsedWrapper {
   }
 
   /// Creates a property that represents another parsable type, using the
-  /// specified visibility.
-  public init(visibility: ArgumentVisibility = .default) {
+  /// specified title and visibility.
+  ///
+  /// - Parameters:
+  ///   - title: A title for grouping this option group's members in your
+  ///     command's help screen. If `title` is empty, the members will be
+  ///     displayed alongside the other arguments, flags, and options declared
+  ///     by your command.
+  ///   - visibility: The visibility to use for the entire option group.
+  public init(
+    title: String = "",
+    visibility: ArgumentVisibility = .default
+  ) {
     self.init(_parsedValue: .init { _ in
-      ArgumentSet(Value.self, visibility: .private)
+      var args = ArgumentSet(Value.self, visibility: .private)
+      args.content.withEach {
+        $0.help.parentTitle = title
+      }
+      return args
     })
     self._visibility = visibility
+    self.title = title
   }
 
   /// The value presented by this property wrapper.
@@ -109,5 +127,17 @@ extension OptionGroup {
   @_disfavoredOverload
   public init() {
     self.init(visibility: .default)
+  }
+}
+
+// MARK: Deprecated
+
+extension OptionGroup {
+  @_disfavoredOverload
+  @available(*, deprecated, renamed: "init(title:visibility:)")
+  public init(
+    visibility _visibility: ArgumentVisibility = .default
+  ) {
+    self.init(title: "", visibility: _visibility)
   }
 }
--- a/Sources/ArgumentParser/Usage/HelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/HelpGenerator.swift
@@ -51,6 +51,7 @@ internal struct HelpGenerator {
       case positionalArguments
       case subcommands
       case options
+      case title(String)
       
       var description: String {
         switch self {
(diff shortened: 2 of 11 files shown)


natecook1000:
@rauhul I think we'll need an updated schema for the JSON help dump, and then we can update the man page generation to support this feature as well. Does that sound right?

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -203,8 +214,16 @@ internal struct HelpGenerator {
           abstract: command.configuration.abstract)
     }
     
+    // Combine the compiled groups in this order:
+    // - arguments
+    // - named sections
+    // - options/flags
+    // - subcommands
```
natecook1000, line 221: @rauhul @ethan-kusters Does this seem like the correct order?

Sources/ArgumentParser/Parsable Properties/OptionGroup.swift
```diff
@@ -111,3 +129,15 @@ extension OptionGroup {
     self.init(visibility: .default)
   }
 }
+
+// MARK: Deprecated
+
+extension OptionGroup {
+  @_disfavoredOverload
+  @available(*, deprecated, renamed: "init(title:visibility:)")
```
rauhul, line 137: Do we need this still, SAP isn't ABI stable and the above initializer is source compatible?

rauhul: eh, well very very minimally source breaking if someone ever used a point free version of the function: `OptionGroup.init(visibility:)`I really doubt this is the case though.

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -136,6 +139,8 @@ internal struct HelpGenerator {
     
     var positionalElements: [Section.Element] = []
     var optionElements: [Section.Element] = []
+    var titledSections: [String: [Section.Element]] = [:]
+    var sectionTitles: [String] = []
```
rauhul, line 143: is this needed? It seems like a duplicate of `titledSections.keys`

rauhul (approved):
LGTM, I'm a little unsure if setting the title on every instance of an option group is the best api. It may be valuable to have an OptionGroupConfiguration so a default title could be provided by the type. This can always be added in the future, so theres no reason to stop this change.

Sources/ArgumentParser/Usage/HelpGenerator.swift
rauhul: NVM this is so they are sorted. A comment about this would be nice!

Sources/ArgumentParser/Usage/HelpGenerator.swift
rauhul: Yeah this makes sense to me!

natecook1000:
> LGTM, I'm a little unsure if setting the title on every instance of an option group is the best api. It may be valuable to have an OptionGroupConfiguration so a default title could be provided by the type. This can always be added in the future, so theres no reason to stop this change.

This is a really good point. I think since we'll want to be able to specify a title at the `@OptionGroup` site even if we have type-level customization (i.e. as an override), it's okay to continue with this first.

Sources/ArgumentParser/Utilities/StringExtensions.swift
```diff
@@ -239,4 +238,8 @@ extension StringProtocol where SubSequence == Substring {
     guard lines.count == 2 else { return lines.joined(separator: "") }
     return "\(lines[0])\n\(lines[1].indentingEachLine(by: n))"
   }
+  
+  var nilIfEmpty: Self? {
```
rauhul, line 242: Note: this is called `nonEmpty` in ArgumentParserToolInfo

rauhul (approved):

PR #494, opened 2022-09-19
Fix help display for non-String RawRepresentables

natecook1000:
RawRepresentable types that have a non-String raw value are having values displayed in the help screen by converting the RawRep value into a string. However, these values are by default parsed by their raw value, so we should use that for display instead.

This is accomplished by adding a defaultValueDescription implementation for all ExpressibleByArgument-conforming RawValue types, and then basing the allValues implementation on that. This generalizes the existing overloads for String-based RawRep types, while also allowing users who customize their ExpressibleByArgument implementation to provide the correct help value for clients.

### Checklist


natecook1000:
@swift-ci Please test

PR #505, opened 2022-10-10
Change name of plugin target to readable form

clayellis:
This PR updates the name of the "Generate Manual" plugin so that it appears in a more readable "command" form in the context menu in Xcode.

Before:
image

After:
image

### Checklist

### DIFF
--- a/Package@swift-5.6.swift
+++ b/Package@swift-5.6.swift
@@ -20,7 +20,7 @@ var package = Package(
             targets: ["ArgumentParser"]),
         .plugin(
             name: "GenerateManualPlugin",
-            targets: ["GenerateManualPlugin"]),
+            targets: ["Generate Manual"]),
     ],
     dependencies: [],
     targets: [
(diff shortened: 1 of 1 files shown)


clayellis:
Arguably, there should be an additional parameter on the `Plugin` target type that allows for an easier "external" name configuration without having to rename actual targets. Maybe it's worth a pitch? Someone else may have to take it up, though.

rauhul:
I'm on board with making the name look better in Xcode, though I'm not a huge fan of having a space in the file path.

@neonichu do you have any thoughts, is this the best solution for improving the display name in Xcode?

clayellis:
> I'm on board with making the name look better in Xcode, though I'm not a huge fan of having a space in the file path.

This _avoids_ having a space in the file path (by explicitly declaring a `path`) and moves that to the target name.

natecook1000 (approved):
Since this keeps the command-line verb the same, it looks like the change is really kept local to the package manifest, which then gets picked up by Xcode for the action name. Given that, I think this is fine to merge, especially since it doesn't bind us with source compatibility or anything else if we do discover problems down the road.

natecook1000:
@swift-ci Please test

clayellis:
Thanks for merging this! I hope to contribute in larger ways in the future!

PR #504, opened 2022-10-10
[NFC] Sequester platform-specific code

natecook1000:

PR #506, opened 2022-10-12
ArgumentParser: repair the build after #504

compnerd:
There are references to functions from the Windows SDK rather than the C library for the console handling.  Explicitly import the necessary types and functions from the WinSDK module so that Windows can build once more with the changes to sequester the platform specific code.


compnerd:
CC: @natecook1000 @rauhul

compnerd:
@swift-ci please test

natecook1000 (approved):
Apologies — looks like I missed an `import WinSDK` when moving things over. Working on getting a better testing situation in place to prevent things like this in the future.

PR #522, opened 2022-11-16
Disfavor deprecated property wrapper inits

natecook1000:
This change is to make sure that the non-optional generic parameter type is chosen for `@Argument` and `@Option` properties that provide a default value. Previously, an optional type might be chosen depending on how the `help` parameter was spelled, due to strange overload resolution. In particular, using the `.init` caused selection of the deprecated optional overload:

```swift
// Unexpected:
// Infers `Value` as `Optional<AbsolutePath>`
@Argument(help: .init("The path"))
var path = AbsolutePath("/")

// Expected:
// Infers `Value` as `AbsolutePath`
@Argument(help: "The path")
var path = AbsolutePath("/")
```

This addresses the issue by marking the deprecated overloads as disfavored. rdar://102383455

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Argument.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Argument.swift
@@ -290,7 +290,7 @@ public struct ArgumentArrayParsingStrategy: Hashable {
 }
 
 // MARK: - @Argument T: ExpressibleByArgument Initializers
-extension Argument {
+extension Argument where Value: ExpressibleByArgument {
   /// Creates a property with a default value provided by standard Swift default
   /// value syntax.
   ///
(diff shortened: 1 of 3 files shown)


natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -309,7 +309,7 @@ extension Argument {
     wrappedValue: Value,
     help: ArgumentHelp? = nil,
     completion: CompletionKind? = nil
-  ) where Value: ExpressibleByArgument {
```
rauhul, line 312: Just curious: does this change affect overload resolution?

PR #527, opened 2022-11-21
Partially revert "Generate Manual" target name

rauhul:
- The name of "GenerateManualPlugin" was changed to "Generate Manual" to improve the plugin's display name in Xcode. This however results in the `.pluginWorkDirectory` provided by SPM to include a space (' ') in the path. The space in the output path by itself is not a deal breaker, but unfortunately `man` (at least on macOS) does not properly handle path arguments with spaces. Example:

  ```shell
  ➜  swift package generate-manual
  ...
  Generating manual for roll...
  Generated manual in '.build/plugins/Generate Manual/outputs/roll'
  ...
  ➜  man '.build/plugins/Generate Manual/outputs/roll/roll.1'
  /usr/bin/man: line 413: cd: .build/plugins/Generate: No such file or \
    directory
  ```

  IMO, easily previewing the generating manual takes precedence over the UI in Xcode; as a result, this commit partially reverts the plugin's name to "GenerateManual" to hopefully provide a good middle ground of an output path without a space and a display name in Xcode that is a little easier on the eyes than "GenerateManualPlugin" was.


rauhul:
@swift-ci please test

rauhul:
cc @clayellis @neonichu

rauhul:
@swift-ci please test

clayellis:
Sounds good, even more reason to pitch an explicit user-facing name for plugins.

PR #528, opened 2022-11-21
Cleanup Foundation usage

rauhul:
- Changes uses of import Foundation in ArgumentParser to only expose the symbols needed to avoid growing unintentional dependencies.
- Replaces direct usage of EXIT_FAILURE in MessageInfo with ExitCode, exposed as a result of the above change.


rauhul:
@swift-ci please test

rauhul:
@swift-ci please test

Issue #534, opened 2022-12-11
A generated fish script doesn't complete an option after an argument.


mtj0928:
A fish script that is generated by `--generate-completion-script` doesn't complete an option after an argument.
For example, `repeat` command which is contained in the repository as an example can run the following command.
```shell
$ repeat foo --count 10 
```
However, `--count` is not completed even though `$ repeat foo --c` is input and a tab key is pressed.

**ArgumentParser version:** `main`
**Swift version:**  
```
swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0
```

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
1. Build this repository
```shell 
$ swift build -c release
```
2. Add the release directory into your $PATH.
```shell
$ set -x PATH (string join "" (pwd) "/.build/release/") $PATH  
```

3. Generate a fish completion script.
```
$ repeat --generate-completion-script fish >> ~/.config/fish/completions/repeat.fish
```

4. Type `repeat foo --` and press a tab key to try to complete options.

### Expected behavior
Options that `repeat` supports are shown.

### Actual behavior
Nothing is shown.

hassila:
See https://github.com/apple/swift-argument-parser/issues/376 also

natecook1000:
Closed by #535.

Issue #376, opened 2021-12-08
Custom fish completions - do they work?


floam:
I was looking over some issues, and for fun checked out https://gitlab.com/frizlab-demo-projects/swift-argument-parser-example-for-pr-320 which was a test case for some problem that affected zsh and not bash. I notice that for this project the custom completions are not generated for fish at all. I expect there to be something calling `---completion …`?

```
$ example --generate-completion-script=fish
function __fish_example_using_command
    set cmd (commandline -opc)
    if [ (count $cmd) -eq (count $argv) ]
        for i in (seq (count $argv))
            if [ $cmd[$i] != $argv[$i] ]
                return 1
            end
        end
        return 0
    end
    return 1
end
complete -c example -n '__fish_example_using_command example' -f -r -l option1
$ example --generate-completion-script=zsh
#compdef example
local context state state_descr line
_example_commandname=$words[1]
typeset -A opt_args

_example() {
    integer ret=1
    local -a args
    args+=(
        '--option1:option1:'
        ':arg:{_custom_completion $_example_commandname ---completion  -- arg $words}'
        '(-h --help)'{-h,--help}'[Print help information.]'
    )
    _arguments -w -s -S $args[@] && ret=0

    return ret
}


_custom_completion() {
    local completions=("${(@f)$($*)}")
    _describe '' completions
}

_example

**Comment by natecook1000** [comment]
I’ve just gotten to the same point when investigating why the shortcuts tool doesn’t provide the kind of autocompletions you had suggested in your previous issue. It looks like the fish completions leave out positional arguments entirely, and only include completions for flags and options - not ideal!

**Comment by floam** [comment]
I ought to learn Swift before I get too dangerous making code changes in a shipping product.

**Comment by rauhul** [comment]
@floam if you describe the changes need in detail, I (if I can find time) or another contributor could implement them.

**Comment by floam** [comment]
I'm a hacker and eternal memory hole procrastinator. I'll probably fix it all up anyhow or disappear. But I'll keep it in mind!

**Comment by hassila** [comment]
I made a simple go at using the fish completion too now and couldn't figure out what I was doing wrong yet.

Just a small test project for testing some stuff:

```
hassila@max ~/D/G/d/.b/debug (main)> ./dm --help
USAGE: dm [--generate-flatbuffers] [--generate-fast-binary-encoding]

OPTIONS:
  --generate-flatbuffers  Generate Flatbuffers schema.
  --generate-fast-binary-encoding
[...]

PR #535, opened 2022-12-17
Improve fish completion (#376, #534)

mtj0928:
A generated fish script didn't complete an option after an argument.

The cause is the generated fish script doesn't accept an input text which already has arguments. 
For example, when the input is `repeat -`, the script can complete `--count`, but when the text is `repeat foo -`, the script cannot complete `repeat foo --count`, because the input text already has the argument "foo".

To fix the issue, I implemented the following steps.
1. Preprocess for the subcommand.
To support options of the root command like `math --version add`, `--version` is removed by `_swift_{command name}_preprocessor`.

2. Decide a command.
In a case when a command has some subcommands like `math stats average 1 10 100 --kind mean`, we need to decide the target command.  (In the example  case, the target command is `average`.)
To decide it, the newly generated script receives `expected command` and `subcommands`.
(e,g,)
    1. A case checking whether the target command is `stats` or not.
        - `expected command` = `math stats`
        - `subcommands` = `average stdev quantiles help`
    2. A case checking whether the target command is `average` or not.
        - `expected command` = `math stats average`
        - `subcommands` is empty (`average` doesn't have any subcommands.)`

    When the input has the expected command, and a text after the expected command is not the subcommand, the newly generated script decide the command is a target command, and suggest options for the target command.

### Related issues
- #376
- #534

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/FishCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/FishCompletionsGenerator.swift
@@ -63,8 +63,7 @@ extension ArgumentDefinition {
           !names.isEmpty
     else { return nil }
 
-    var results = names
-      .map{ $0.asFishSuggestion }
+    var results = names.map{ $0.asFishSuggestion }
 
     if !help.abstract.isEmpty {
       results += ["-d '\(help.abstract.fishEscape())'"]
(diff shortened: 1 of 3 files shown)


natecook1000 (changes requested):
Thanks so much, @mtj0928! Just a couple of small changes below, and could you update the `fish` test expectation for `testMath_CompletionScript()`?

mtj0928:
@natecook1000 Thank you for your review!
I applied your comment, and please review this PR again!

natecook1000:
@swift-ci Please test

PR #548, opened 2023-02-02
Stop removing underscores from CodingKey names in InputKey

natecook1000:
When a property wrapper is applied to a property, the property's storage is given a name with a prefixed underscore. That is, for a property named `x`, the actual storage is named `_x`.

That prefixed storage is what is visible through reflection, so when building an `ArgumentSet` from a command type's mirror, we need to remove the leading underscore. This is done when creating an `InputKey` for each property.

However, `InputKeys` are also created from `CodingKeys` during decoding of a command. These `CodingKeys` _do not_ have the leading underscore that is added by the property wrapper, so any underscores that appear are actually from the declaration of the property with an underscored name. Removing leading underscores from `CodingKey` names results in a mismatch when trying to find the decoded value. For example, for a simple command declared like this:

```
@main
struct Repro: ParsableCommand {
    @Option(name: .customLong("arg"))
    var _arg: String?

    func run() throws {
        print(self._arg ?? "nil")
    }
}
```

Running: `repro --arg foo` fails to parse with the error:
```
Error: Unknown option '--arg'. Did you mean '--arg'?
Usage: repro [--arg <arg>]
  See 'repro --help' for more information.
```

This change simplifies the `InputKey` type to use an array path instead of an indirect enum and removes the leading underscore dropping when creating an `InputKey` from a `CodingKey`.

rdar://104928743

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -132,7 +132,7 @@ struct BashCompletionsGenerator {
   ///
   /// These consist of completions that are defined as `.list` or `.custom`.
   fileprivate static func generateArgumentCompletions(_ commands: [ParsableCommand.Type]) -> [String] {
-    ArgumentSet(commands.last!, visibility: .default, parent: .root)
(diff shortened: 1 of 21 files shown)


natecook1000:
@swift-ci Please test

PR #554, opened 2023-02-20
Handle unparsed non-optional complex property types

### LINKED ISSUE #553: Crash when having a dictionary property (ArgumentDecoder.swift:89: Fatal error)
Swift argument parser crashes when adding a dictionary property - see attached reproducer.

**ArgumentParser version:** `1.2.2` (reproduced with `main` too)
**Swift version:**  `swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0
`
### Checklist

### Steps to Reproduce
Unpack the attached reproducible project and run from command line with:
`swift run swift-argument-parser-reproducer --command run`

The complete code is simply:
```swift
import ArgumentParser

typealias BenchmarkIdentifier=Int

struct BenchmarkResult: Codable {
    var x:Int
}

@main
struct BenchmarkTool: AsyncParsableCommand {
    @Option(name: .long, help: "The command to perform")
    var command: String

    var benchmarkResults: [BenchmarkIdentifier: [BenchmarkResult]] = [:] // comment this out to not crash

    mutating func run() async throws {
    }
}
```

### Expected behavior
```
hassila@ice ~/G/swift-argument-parser-reproducer> swift run swift-argument-parser-reproducer --command run
Building for debugging...
[3/3] Linking swift-argument-parser-reproducer
Build complete! (0.39s)
USAGE: benchmark-tool --command <command>

OPTIONS:
  --command <command>     The command to perform

hassila@ice ~/G/swift-argument-parser-reproducer> 
```

### Actual behavior
```
hassila@ice ~/G/swift-argument-parser-reproducer> swift run swift-argument-parser-reproducer --command run
Building for debugging...
[47/47] Linking swift-argument-parser-reproducer
Build complete! (4.50s)
ArgumentParser/ArgumentDecoder.swift:89: Fatal error
fish: Job 1, 'swift run swift-argument-pars...' terminated by signal SIGTRAP (Trace or breakpoint trap)
```

[swift-argument-parser-reproducer.zip](https://github.com/apple/swift-argument-parser/files/10731814/swift-argument-parser-reproducer.zip)

gwynne:
As initially demonstrated in #553, the existing `ArgumentDecoder` logic fails when presented with a property which:

- Is unparsed
- Is not optional
- Has a default value, and
- Requests a keyed container during decoding

This occurs because while the actual property in question is properly assessed as having `.defaultValue` as an `InputOrigin`, the request for a keyed decoding container bypasses the logic which would otherwise return said default value, and since the `ArgumentDefinition` logic is (correctly) not recursive, the type fails to decode any keys it contains due to a "missing value".

The most straightforward solution I could find, one which covers all cases without negatively affecting handling of parsed properties (to the best of my ability to determine), was to apply the "default value" handling from the `decodeIfPresent<T>(_:forKey:)` method of the keyed decoding container (which ensured that _optional_ properties of this sort already worked correctly) to the more fundamental "bottleneck" `decode<T>(_:forKey:)` method as well, modified to throw a `wrongType` error instead of returning nil on type mismatch.

I additionally added a functioning `allKeys` property accessor for additional robustness against custom `Decodable.init(from:)` implementations. (The choice not to do the same for `nestedContainer(keyedBy:)`, `nestedUnkeyedContainer()`, `superDecoder()`, and `superDecoder(forKey:)` was simply pragmatic - it's not even close to worth the effort.)

Fixes #553 

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/ArgumentDecoder.swift
+++ b/Sources/ArgumentParser/Parsing/ArgumentDecoder.swift
@@ -86,7 +86,17 @@ final class ParsedArgumentsContainer<K>: KeyedDecodingContainerProtocol where K 
   }
   
   var allKeys: [K] {
-    fatalError()
+    /// This logic selects only those element keys which satisfy both of two conditions:
+    ///
+    /// 1. The input key's path is exactly equal to this container's coding path,
+    ///    neither shallower nor deeper.
+    /// 2. An instance of this container's coding key type can be instantiated using
+    ///    the input key's name.
+    ///
+    /// This is, in essence, an iterative inverse of the `element(forKey:)` method.
+    self.decoder.values.elements.keys
+        .filter { $0.path == self.codingPath.map(\.stringValue) }
+        .compactMap { K.init(stringValue: $0.name) }
   }
   
   fileprivate func element(forKey key: K) -> ParsedValues.Element? {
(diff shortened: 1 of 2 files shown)


natecook1000 (commented):
@gwynne Thanks so much for this fix! I believe your change to `decode(_:forKey:)` is sufficient to solve the problem. Unless we can create a test that exercises the new `allKeys`, let's revert that back to a fatal error.

Can you also switch to 2-space indenting? Thanks!

gwynne:
@natecook1000 Well, I tried to come up with a test that would exercise the `allKeys` implementation, but it turns out nothing I could think of would get that far with the change to `decode(_:forKey:)` in place; I've removed it and put back the `fatalError()` - and also fixed my indentation, don't worry! 😆

natecook1000:
@swift-ci Please test

hassila:
(thanks @gwynne for fixing this, much obliged!)

PR #557, opened 2023-02-27
Provides `initial` initializer parameter for input property wrappers

tarbaiev-smg:
### Description
Currently it's only possible to use `Optional` initial value if the property type has `Optional` type (by means of the `wrappedValue` parameter).
In some cases we might need to provide an optional default value, e.g when reading it from the environment.
To allow this I'd like to propose adding the `initial` argument to initializers of non-`Optional` type property wrappers which do not contain the `wrappedValue` argument.

### Detailed Design
As an example, adding the `initial` parameter to the `Option` initializer:
```swift
  public init(
    parsing parsingStrategy: SingleValueParsingStrategy = .next,
    help: ArgumentHelp? = nil,
    completion: CompletionKind? = nil,
    initial: Value? = nil // <-- Using default value preserves backward compatibility
  ) {
    self.init(_parsedValue: .init { key in
      let arg = ArgumentDefinition(
        container: Bare<Value>.self,
        key: key,
        kind: .name(key: key, specification: name),
        help: help,
        parsingStrategy: parsingStrategy.base,
        initial: initial, // <-- injecting the initial value
        completion: completion)

      return ArgumentSet(arg)
    })
```

would allow us to use it as follows:

```swift
@Option(initial: ProcessInfo.processInfo["REQUIRED_OPTION"])
var requiredOption: String
```

The argument therefore remains required. A `nil` initial value would fail the command if the argument is not provided in the command line.

### Documentation Plan
The initialzers' documentation has been updated.

### Test Plan
Relevant tests added.

### Source Impact
The change is backward-compatible.

### Checklist


Coeur:
@tarbaiev-smg I believe you'd need to rebase your PR, since it now has conflicts that must be resolved.

tarbayev:
@Coeur I'd be happy to, but first I'd like to get a confirmation that it still has a chance to be merged. It's been a year since I opened it.

Coeur:
@tarbayev I don't know.

PR #573, opened 2023-06-13
Bash does not like '-' in function names

kennyyork:
If a ParsableCommand has a `commandName` value with a hyphen, the bash completion generator will generate a function name with a hyphen which isn't valid in BASH, eg `_programName_command_sub-command': not a valid identifier`

### Checklist


natecook1000:
Thanks so much for working on this, @kennyyork!

It looks like we're already escaping in this way for `zsh`, and it's probably safe to assume that `fish` has this same requirement. Could you rename the existing `zshEscapingCommandName` method to `shellEscapingCommandName`, and then have all three script generators use that?

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #574, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
macOS failure is unrelated and is also reproduced in an empty PR https://github.com/apple/swift-argument-parser/pull/575. @rauhul have you seen this one before?

natecook1000:
@swift-ci Please test

PR #576, opened 2023-06-26
Skip count-lines tests in CI for now

natecook1000:
CI is allowing a `guard #available(macOS 12)` in the tests to pass, but then reporting macOS version 10.16 both in ProcessInfo and in the guard in the actual `CountLines.run()` method, which results in an inconsistency. Skipping the test under these conditions until we can sort out the configuration issue.


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #579, opened 2023-08-14
build: add missing static link dependency

compnerd:
The `ArgumentParserTestHelpers` module depends on
`ArgumentParserToolInfo` but fails to indicate that dependency.  This was exposed whilst improving static linking for Windows.


compnerd:
@swift-ci please test

compnerd:
@rauhul - if we could get this merged and tagged, this would help unblock some windows work to improve static linking.

PR #582, opened 2023-08-26
Add Sendable conformance

vlm:
It can be useful to send `ParsableArgument` as an argument to some function. This is trivial to do in a non-async context. However, in `async` context the type that is sent as an argument has to be Sendable. This patch adds `Sendable` conformance so this becomes expressible and warning-free:

```
struct Args: AsyncParsableCommand, Sendable {
    func run() async throws {
        Task { await otherFunction(arguments: self) }
    }
}
```

The warnings will become errors in Swift 6, so this is a forward-looking change.

### Checklist
- [ ] I've read the [Contribution Guidelines](https://github.com/apple/swift-argument-parser/blob/main/CONTRIBUTING.md)


natecook1000:
Thanks for this, @vlm! Do we not need to declare the `Parsed` enum as sendable as well? That can indirectly store a user-supplied closure via initializers that include a `transform` parameter, which could break encapsulation. Should we look at adding `@Sendable` annotations to those closures at the same time?

vlm:
@natecook1000 looking...

vlm:
@natecook1000 these conformances doesn't appear to be required. Added a test to showcase that. No warnings.

natecook1000:
@swift-ci Please test

rauhul:
Is it possible to put the additional sendable restrictions behind conditional compilation so this change isn't source breaking?

Package@swift-5.6.swift
```diff
@@ -28,7 +28,9 @@ var package = Package(
         .target(
             name: "ArgumentParser",
             dependencies: ["ArgumentParserToolInfo"],
-            exclude: ["CMakeLists.txt"]),
+            exclude: ["CMakeLists.txt"],
+            swiftSettings: [.unsafeFlags(["-Xfrontend", "-strict-concurrency=complete"])]
```
0xTim, line 32: This is a "very bad idea" ™️ - anyone who pulls in the 5.6 manifest will see this propagated to their library or application and potentially introduce compiler errors and warnings.

We solved this in Vapor with a 5.9 specific manifest using the experimental feature flag which restricts it to that package - e.g. https://github.com/vapor/vapor/blob/main/Package%40swift-5.9.swift#L96

natecook1000: Definitely! Thanks for that note — I was planning to just remove the flag, and didn't realize that `enableExperimentalFeature` allows us to keep the setting package-local.

natecook1000:
@swift-ci Please test

PR #583, opened 2023-09-02
Platform.shellName

PopTo:
Platform.shellName returns the shell that is currently in use

### Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,17 +15,18 @@ import PackageDescription
 var package = Package(
     name: "swift-argument-parser",
     products: [
-        .library(
-            name: "ArgumentParser",
-            targets: ["ArgumentParser"]),
+        .library(name: "ArgumentParser", targets: ["ArgumentParser"]),
+        .library(name: "CTools", targets: ["CTools"])
     ],
     dependencies: [],
     targets: [
         // Core Library
+      .target(name: "CTools"),
         .target(
             name: "ArgumentParser",
-            dependencies: ["ArgumentParserToolInfo"],
-            exclude: ["CMakeLists.txt"]),
+            dependencies: ["ArgumentParserToolInfo", "CTools"],
+            exclude: ["CMakeLists.txt"],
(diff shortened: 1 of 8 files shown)


natecook1000:
Can you add a note about how you've tested this? When I run `/bin/ps`, I get a list of all my running processes, some of which are shells and some not. Is this doing extra work to find the currently executing process?

rauhul:
Im not a huge fan of running another process to find this information, it may not play well with sandboxing and other environmental constraints. Could we use posix or platform specific APIs where needed to find the parent shell?

PopTo:
@natecook1000 You're right. Maybe the echo $0 will be a better solution.

PopTo:
Yes, @rauhul. Do you have any suggestions for this API? Because I don't know anything that would help.

rauhul:
> Yes, @rauhul. Do you have any suggestions for this API? Because I don't know anything that would help.

On unix-y things you can start with `getppid()` to get the parent's pid and use a platform specific API to get the process name. For example `libproc` on Darwin platforms has a bunch of related API such as `proc_pidpath`.

PopTo:
@rauhul maybe I found the solution. It's a simple C function:

```
#include <pwd.h>
#include <unistd.h>

char* shellPath(void) { return getpwuid(geteuid())->pw_shell; }
```
It works on Darwin and Linux. probably also on Windows

What do you think?

gwynne:
The `getppid()` route is not really viable IMO; it requires potentially elevated privileges and is highly platform-specific. I think `SHELL` (maybe with a `getpwuid()->pw_shell` fallback) is as good as can be reasonably asked for - I've yet to see any utility that needs to differentiate between shells _in vivo_ do so by any means other than requiring the user to specify.

natecook1000:
@PopTo Thanks again for your work on this, but I think the end solution is getting more heavyweight than justified by the benefit of the result (slightly better shell auto detection). Going to close this PR – let me know if you'd like to discuss further.

PopTo:
@natecook1000 I agree with you.

PR #594, opened 2023-11-07
Display possible option values in help

rauhul:
Updates HelpGenerator to print possible value options as a suffix to the
user defined help string. In practice this looks like:

> Set diagnostic level to report public declarations without an
> availability attribute. (values: error, warn, ignore; default: warn)

Updates string diff printer to fix an OOB indexing bug.


rauhul:
@swift-ci please test

natecook1000 (commented):
This looks great! I'm a little concerned about how this will affect the help when a `CaseIterable` type has a large number of values. Can we update the `allValueStrings` documentation to show how to suppress/limit that number, or would it make sense to provide a configuration property to `ArgumentHelp`? (e.g. `maxCaseDisplay`)

Sources/ArgumentParser/Documentation.docc/Articles/CustomizingHelp.md
```diff
@@ -74,6 +74,82 @@ OPTIONS:
...
+struct FruitStore: ParsableCommand {
+    @Argument(help: "The fruit to purchase")
+    var fruit: Fruit
+  
+    @Option(help: "The number of fruit to purchase")
+    var quantity: Int = 1
+}
+```
+
+...and the help screen contains the derived valid values:
```
natecook1000, line 137:
```suggestion
The help screen still contains all the valid values.
```

Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
```diff
@@ -22,7 +22,9 @@ public protocol ExpressibleByArgument {
   /// An array of all possible strings that can convert to a value of this
   /// type.
```
natecook1000, line 23: Let's make clear that this is for the help screen, not a parsing requirement.

```suggestion
  /// An array of all possible strings that can convert to a value of this
  /// type, for display in the help screen.
```

natecook1000 (commented):
Updated docs look great! Added some light editing through the comments, then we can merge 👍🏻

rauhul:
@swift-ci please test

PR #595, opened 2023-11-07
Fix crash in string diff printer

rauhul:
Updates string diff printer to fix an OOB indexing bug.


rauhul:
@swift-ci please test

natecook1000 (approved):

PR #596, opened 2023-11-12
Respect the COLUMNS and LINES environment variables when present

gwynne:
As documented in both the [Linux](https://man7.org/linux/man-pages/man7/environ.7.html:~:text=COLUMNS) and [BSD](https://man.freebsd.org/cgi/man.cgi?environ(7)#:~:text=COLUMNS) versions of the `environ(7)` manpage, many console utilities respect the presence of the `COLUMNS` and `LINES` environment variables, treating them as overrides (or, less commonly, fallbacks) of the terminal's reported size (if any). This adds that same behavior to ArgumentParser's screen size logic, affecting the wrapping of the usage generator's output.

_Note: These changes follow the semantics of `ls(1)`, where `COLUMNS` and `LINES` individually override the values reported by a terminal when they are present. Setting only one of the two variables has no effect on the other. It is also worth noting that at the present time, overriding the screen height (i.e. `LINES`) has no impact whatsoever on ArgumentParser's behavior._

The new logic only takes effect on non-Windows/WASI platforms.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Utilities/Platform.swift
+++ b/Sources/ArgumentParser/Utilities/Platform.swift
@@ -120,16 +120,57 @@ extension Platform {
     (80, 25)
   }
   
-  /// Returns the current terminal size, or the default if the size is
-  /// unavailable.
-  static func terminalSize() -> (width: Int, height: Int) {
+  /// The terminal size specified by the COLUMNS and LINES overrides
+  /// (if present).
+  ///
+  /// Per the [Linux environ(7) manpage][linenv]:
+  ///
+  /// ```
+  /// * COLUMNS and LINES tell applications about the window size,
+  ///   possibly overriding the actual size.
+  /// ```
+  ///
+  /// And the [FreeBSD environ(7) version][bsdenv]:
+  ///
+  /// ```
+  /// COLUMNS    The user's preferred width in column positions for the
+  ///            terminal.  Utilities such as ls(1) and who(1) use this
+  ///            to format output into columns.   If  unset  or  empty,
+  ///            utilities  will use an ioctl(2) call to ask the termi-
+  ///            nal driver for the width.
+  /// ```
+  ///
+  /// > Note: Always ignored on Windows.
+  ///
+  /// [linenv]: https://man7.org/linux/man-pages/man7/environ.7.html:~:text=COLUMNS
+  /// [bsdenv]: https://man.freebsd.org/cgi/man.cgi?environ(7)#:~:text=COLUMNS
+  static func overridenTerminalSize() -> (width: Int, height: Int) {
+    var width = 0, height = 0
+
+#if !os(Windows) && !os(WASI)
+    if let colsCStr = getenv("COLUMNS"), let colsVal = Int(String(cString: colsCStr)) {
+      width = colsVal
+    }
+    if let linesCStr = getenv("LINES"), let linesVal = Int(String(cString: linesCStr)) {
+      height = linesVal
+    }
+#endif
+
+    return (width: width, height: height)
+  }
+  
+  /// The current terminal size as reported by the windowing system,
+  /// if available.
+  ///
+  /// Returns (0,0) if no reported size is available.
+  static func reportedTerminalSize() -> (width: Int, height: Int) {
 #if os(WASI)
     // WASI doesn't yet support terminal size
-    return defaultTerminalSize
+    return (width: 0, height: 0)
 #elseif os(Windows)
-    var csbi: CONSOLE_SCREEN_BUFFER_INFO = CONSOLE_SCREEN_BUFFER_INFO()
+    var csbi = CONSOLE_SCREEN_BUFFER_INFO()
     guard GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &csbi) else {
-      return defaultTerminalSize
+      return (width: 0, height: 0)
     }
     return (width: Int(csbi.srWindow.Right - csbi.srWindow.Left) + 1,
             height: Int(csbi.srWindow.Bottom - csbi.srWindow.Top) + 1)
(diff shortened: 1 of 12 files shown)


Sources/ArgumentParser/Utilities/Platform.swift
```diff
@@ -120,16 +120,57 @@ extension Platform {
...
+  ///            to format output into columns.   If  unset  or  empty,
+  ///            utilities  will use an ioctl(2) call to ask the termi-
+  ///            nal driver for the width.
+  /// ```
+  ///
+  /// > Note: Always ignored on Windows.
+  ///
+  /// [linenv]: https://man7.org/linux/man-pages/man7/environ.7.html:~:text=COLUMNS
+  /// [bsdenv]: https://man.freebsd.org/cgi/man.cgi?environ(7)#:~:text=COLUMNS
+  static func overridenTerminalSize() -> (width: Int, height: Int) {
```
rauhul, line 147: Could we use `Int?`/`nil` instead of `0`?

gwynne: I originally went with `0` based on the implementation of the `TIOCGWINSZ` logic, but now that you point it out, in retrospect I realize that the use of `0` was based on the semantics of the `ioctl()` itself rather than informing the code as a whole 😅. Will push an update in a few minutes!

rauhul:
@swift-ci please test

Sources/ArgumentParser/Utilities/Platform.swift
```diff
@@ -144,17 +190,38 @@ extension Platform {
...
+    return (
+      width: specifiedSize.width ?? reportedSize.width ?? defaultTerminalSize.width,
+      height: specifiedSize.height ?? reportedSize.height ?? defaultTerminalSize.height
+    )
+  }
+  
   /// The current terminal size, or the default if the width is unavailable.
   static var terminalWidth: Int {
-    terminalSize().width
+    self.terminalSize().width
```
rauhul, line 225: I appreciate the explicit `self`

rauhul (approved):

Sources/ArgumentParser/Utilities/Platform.swift
gwynne: Suffice to say I wish there was an `.enableUpcomingFeature()` flag called `"DisallowImplicitSelf"` 🤣

rauhul: +10000000

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
```diff
@@ -26,7 +26,8 @@ fileprivate struct Qux: ParsableArguments {
 fileprivate struct Quizzo: ParsableArguments {
   @Option() var name: String
   @Flag() var verbose = false
-  let count = 0
+  let count: Int
+  init() { self.count = 0 } // silence warning about count not being decoded
```
natecook1000, line 30: THANK YOU

natecook1000 (commented):
This looks great! Any ideas about how to test this? Could we set `COLUMNS` for some/one of the help generation tests to verify that it gets picked up as the default?

Tests/ArgumentParserEndToEndTests/UnparsedValuesEndToEndTest.swift
rauhul: +1000 this might allow us to enable "warnings as errors"!!

gwynne:
> This looks great! Any ideas about how to test this? Could we set `COLUMNS` for some/one of the help generation tests to verify that it gets picked up as the default?

Test added!

Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
```diff
@@ -798,3 +798,54 @@ extension HelpGenerationTests {
...
+    @Argument(help: "54 characters of help, so as to wrap when columns < 80")
+    var argument: String?
+  }
+  
+  func testColumnsEnvironmentOverride() throws {
+    #if os(Windows) || os(WASI)
+    throw XCTSkip("Unsupported on this platform")
+    #endif
+    
+    unsetenv("COLUMNS")
```
rauhul, line 813: Do we need to `defer { unsetenv("COLUMNS") }` to avoid this test colliding with other tests which assert on help screens?

I'm actually wondering if this is safely testable when swift test --parallel is used...

gwynne: Oof, that's a good point - Definitionally, it can't really be parallel-safe, in that environment variables themselves are mutable shared global state. The best I think I can do is lessen the amount of variance in the `COLUMNS` value (such as only setting it to 80 and 81, keeping it as close as possible to the default).

rauhul: I'm wondering if the best solution is to just not set the env var and skip test coverage. I know thats a gross solution, but I'd much rather have 100% reliable tests than missing test coverage (imo)

rauhul: @natecook1000

gwynne: I have a partial workaround - I can use `CommandLine.arguments` to detect when parallel testing is in effect (the commandline signatures are unique on both macOS and Linux), and use that to skip the test. The caveat is that it only works for `swift test`; I'd have to skip it unconditionally in Xcode.

gwynne: Update: I tried this solution and it worked in multiple Swift versions on both platforms, so I pushed the code.

natecook1000: Alternatively, could we have this test be the only place that checks that the default 80 column width is used, and just specify exact widths in all the other help generation tests?

natecook1000:
@swift-ci Please test

PR #598, opened 2023-11-14
Enable shell detection for Windows

stevapple:
The conventional `SHELL` environment variable is set by OpenSSH server on Windows, but we need to normalize the path slightly differently. Some popular shell names on Windows are `cmd` (Command Prompt), `powershell` (Windows Powershell), `pwsh` (Powershell Core) and `bash` (Git Bash).

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Utilities/Platform.swift
+++ b/Sources/ArgumentParser/Utilities/Platform.swift
@@ -29,14 +29,23 @@ extension Platform {
   /// The name of the user's preferred shell, if detectable from the
   /// environment.
   static var shellName: String? {
-#if os(Windows)
-    return nil
-#else
     // FIXME: This retrieves the user's preferred shell, not necessarily the one currently in use.
     guard let shellVar = getenv("SHELL") else { return nil }
-    let shellParts = String(cString: shellVar).split(separator: "/")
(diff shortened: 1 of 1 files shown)


Sources/ArgumentParser/Utilities/Platform.swift
```diff
@@ -29,14 +29,23 @@ extension Platform {
...
+    let shellStr = String(cString: shellVar)
+#if os(Windows)
+    let shellParts = shellStr.split(whereSeparator: { $0 == "\\" || $0 == "/" })
+    let shellName = shellParts.last.map {
+      if $0.hasSuffix(".exe") {
+        return String($0.dropLast(4))
+      } else {
+        return String($0)
+      }
+    }
```
compnerd, line 43: Please use Foundation for this, not some hand rolled path processing.

compnerd:
@natecook1000 no, they would not.  There is no completion support for cmd.

compnerd (commented):
I'm not sure I understand where this would be useful, could you share the underlying reason for doing this?

natecook1000:
Since ArgumentParser's shell detection is only to support the completion script request, it doesn't sound like we need this change.

PR #602, opened 2023-11-20
Update documentation

natecook1000:


natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Parsable Properties/Argument.swift
```diff
@@ -49,8 +51,8 @@ public struct Argument<Value>:
     self._parsedValue = _parsedValue
   }
   
-  public init(from decoder: Decoder) throws {
-    try self.init(_decoder: decoder)
+  public init(from _decoder: Decoder) throws {
```
rauhul, line 54: Just curious, why?

rauhul (approved):

Sources/ArgumentParser/Parsable Properties/Argument.swift
natecook1000: It's heavy-handed documentation curation – I think the `Decodable` conformance in the docs is sufficient without this ne'er-used initializer being there as well.

PR #603, opened 2023-11-20
[NFC] Suppress retroactive conformance errors

natecook1000:
These warnings are showing up in tests about types and protocols that are defined within this package, so they don't pose a problem for future library evolution. Instead of using the `@retroactive` attribute, which isn't supported by older compilers, this change fully qualifies the type and protocol in the relevant conformance declarations, which suppresses the issue.

Re: https://github.com/apple/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md


natecook1000:
@swift-ci Please test

PR #606, opened 2023-12-07
build: bump minimum CMake requirement

compnerd:
Update the CMake requirement to 3.19 as Swift requires 3.19.x to build. The upgrade allows us to remove workarounds that we were carrying for older CMake compatibility.


compnerd:
CC: @natecook1000 @rauhul

rauhul:
FWIW I have no issues with you directly merging cmake changes

CMakeLists.txt
```diff
@@ -11,15 +9,12 @@ option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)
...
 include(SwiftSupport)
 
-if(CMAKE_VERSION VERSION_LESS 3.16 AND CMAKE_SYSTEM_NAME STREQUAL Windows)
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
-else()
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-endif()
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
```
bnbarham, line 12: Isn't this too late now @compnerd? `SwiftSupport` is within `cmake/modules` and is included before this now. CC @xedin

compnerd: Hmm, that is a good point 🤔.  How did CI get away with this?

bnbarham: Ah, looking at logs CI is using SPM, not CMake

compnerd: Oh lovely :(

PR #608, opened 2023-12-08
build: correct ordering of `include`

compnerd:
Correct the ordering that @bnbarham pointed out post-commit.  Ensure that we adjust the `CMAKE_MODULE_PATH` prior to the use of `include` which will reference local modules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,6 @@ project(swift-argument-parser
 option(BUILD_EXAMPLES "Build Example Programs" TRUE)
 option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)
 
-include(CTest)
-include(SwiftSupport)
-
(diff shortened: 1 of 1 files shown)


compnerd:
@rauhul - turns out that I do not have commit access on the repository, I would appreciate it if you could assist with that as well 😀

CC: @bnbarham @xedin

compnerd:
@swift-ci please test

bnbarham (approved):

natecook1000:
LGTM! @compnerd will this require a new release?

compnerd:
@natecook1000 - I don't think that the previous change went into 1.3.0, so I think we should be okay.  If it did make it into 1.3.0, we should do a 1.3.1.

compnerd:
@natecook1000, @rauhul - anything else to do before we can merge this?

natecook1000:
Nope! Thanks @compnerd!

PR #612, opened 2024-01-03
Fix unrecognized `-help` flag with default command

natecook1000:
When there's a default subcommand that captures pass-through args, only a standalone help flag should trigger help (since the flag should otherwise go to the subcommand). This fix makes that true for single- dash help flag names, which were previously being skipped, due to single-dash flags being lexed as both one whole flag and as a group of individual short flags.

Re: https://github.com/apple/swift-package-manager/issues/7218
Re: rdar://120422808

### Checklist


natecook1000:
@swift-ci Please test

AndrewHoos (approved):

PR #614, opened 2024-01-10
Improve zsh completion for repeatable options (#564)

### LINKED ISSUE #564: Zsh completion script only shows a completion for the first time in an array.
If a command has either an option or argument with an array type and `completion` set to anything, the completions will only be used for the first item typed into the terminal. Any following items wont get any completions.

In the case of an option with parsing set to one of the single value options, the option and value will only be suggested the first time. Afterwards the option will no longer be given as a completion.

**ArgumentParser version:** 1.2.2
**Swift version:**
Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0

### Checklist

### Steps to Reproduce

```
import ArgumentParser

@main
struct PhoneticAlphabet: ParsableCommand {
    @Option(completion: .list(words))
    var option: [String]
    
    @Argument(completion: .list(words))
    var argument: [String]
    
    func run() throws {
        print("options:", option)
        print("arguments:", argument)
    }
}

private extension PhoneticAlphabet {
    static var words: [String] {
        ["alfa", "bravo", "charlie", "delta", "echo", "foxtrot", "golf", "hotel", "india", "juliett", "kilo", "lima", "mike", "november", "oscar", "papa", "quebec", "romeo", "sierra", "tango", "uniform", "victor", "whiskey", "xray", "yankee", "zulu"]
    }
}
```

1. Build the above command
2. Install the zsh completions for it using `{Command name} --generate-completion-script zsh` and putting the completions in the correct location for your terminal setup. For me it's `~/.oh-my-zsh/completions`.
3. Start typing the command using completions for the option
    1. Type `{Command name} --o` then press tab to complete the option name
    2. Type `a` then tab to complete the word `alfa`
    3. Type `--o` then press tab again
4. Or, start typing the command using completions for the argument
    1. Type `{Command name} a` then press tab to complete the word `alfa`
    2. Type `a` then press tab again

### Expected behavior
For step 3, it should complete `--option` again since it can be passed multiple times to the command to build the `option` array.

For step 4, it should complete `alfa` again since it be passed multiple strings to build the `argument` array.

### Actual behavior
After pressing tab the expected word is not completed. Either nothing is completed or possibly a different flag/option/argument is completed depending on what letter is typed, ie. it might complete `--help` depending on the terminal's completion settings.

mayoff:
Some commands allow an option to be repeated to provide multiple values. For example, ssh allows the -L flag to be repeated to establish multiple port forwardings.

ArgumentParser supports this style when the Option value is an Array and the parsingStrategy is ArrayParsingStrategy.singleValue or .unconditionalSingleValue.

Without this patch, ArgumentParser generates a zsh completion script that does not handle repeatable options correctly. The generated script suppresses completion of any option after that option's first use, even if the option is repeatable.

With this patch, ArgumentParser generates a zsh completion script that allows a repeatable option to be completed each time it is used.

The relevant zsh _arguments syntax is documented here: https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions

Specifically, a repeatable option's optspec needs to start with '*'. Furthermore, a repeatable argument must not list itself or its synonyms in its own parenthesized suppression list. (This is not clearly documented.)

fixes #564

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
@@ -137,7 +137,7 @@ extension ArgumentDefinition {
   func zshCompletionString(_ commands: [ParsableCommand.Type]) -> String? {
     guard help.visibility.base == .default else { return nil }
     
-    var inputs: String
+    let inputs: String
     switch update {
     case .unary:
       inputs = ":\(valueName):\(zshActionString(commands))"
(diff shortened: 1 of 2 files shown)


natecook1000:
Thanks @mayoff! Do we need the same fix for repeatable flags?

mayoff:
This patch also makes a repeatable `@Flag` complete correctly in zsh.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #616, opened 2024-02-15
Remove `@_implementationOnly` annotations

keith:
These annotations produce warnings when compiling swift-syntax without library evolution using Swift ≥5.10.

Replace them by `private import` when compiling using Swift ≥5.11.

Mirrors https://github.com/apple/swift-syntax/pull/2429

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
@@ -9,11 +9,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if swift(>=5.11)
+internal import ArgumentParserToolInfo
+internal import class Foundation.JSONEncoder
+#elseif swift(>=5.10)
+import ArgumentParserToolInfo
+import class Foundation.JSONEncoder
+#else
 @_implementationOnly import ArgumentParserToolInfo
 @_implementationOnly import class Foundation.JSONEncoder
+#endif
 
 internal struct DumpHelpGenerator {
-  var toolInfo: ToolInfoV0
+  private var toolInfo: ToolInfoV0
 
   init(_ type: ParsableArguments.Type) {
     self.init(commandStack: [type.asCommand])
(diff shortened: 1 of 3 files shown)


keith:
@swift-ci please test

keith:
@swift-ci please test

keith:
@natecook1000 can you review? (sorry I can't add you as a reviewer here)

rauhul:
Is there a reason to use private instead of internal?

Additionally could we use an upcoming feature flag instead of conditional compilation?

rauhul:
I wasn't sure if there was a specific reason to prefer private other than minimally scoped imports. Given the context of the swift-syntax comment, this changes lgtm, but I'll wait for Nate to do the final approval

Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
```diff
@@ -9,11 +9,16 @@
...
 //===----------------------------------------------------------------------===//
 
-@_implementationOnly import ArgumentParserToolInfo
-@_implementationOnly import class Foundation.JSONEncoder
+#if swift(>=5.11)
+private import ArgumentParserToolInfo
+private import class Foundation.JSONEncoder
+#else
+import ArgumentParserToolInfo
+import class Foundation.JSONEncoder
```
natecook1000, line 17: I think @rauhul is right that `internal import` matches the existing behavior better than `private import`. Also, can we keep the `@_implementationOnly` attribute for older Swift versions, to maintain the separation?
```suggestion
internal import ArgumentParserToolInfo
internal import class Foundation.JSONEncoder
#else
@_implementationOnly import ArgumentParserToolInfo
@_implementationOnly import class Foundation.JSONEncoder
```

keith: `internal` sounds good to me. We can't keep `@_implementationOnly` for old versions and also solve the warning that lead to me submitting this. Since with Swift 5.10 it seems like it's an unfortunate in-between version where you can't have `internal import` yet, but you also can't avoid the warning for `@_implementationOnly` in all situations. I pushed a new version here which just carves that out so old versions still get `@_implementationOnly`. What do you think?

keith:
@swift-ci please test

PR #617, opened 2024-02-28
Prefer `let` for private configurations

Coeur:
Prefer `let` for private configurations.
Otherwise, we may think that the configuration is mutable, which is not really sendable friendly.

### Checklist


rauhul:
@swift-ci please test

natecook1000:
Makes sense, thanks @Coeur!

PR #619, opened 2024-02-29
Fix for `@Option(transform:)` with optional type

### LINKED ISSUE #618: Ambiguous use of 'init(name:parsing:help:completion:transform:)' when transform throws
Ambiguous use of 'init(name:parsing:help:completion:transform:)' when `transform` throws.

This is a regression introduced with #477 (swift-argument-parser 1.2.0).

**Swift version:** swiftlang-5.11

### Checklist

### Steps to Reproduce
```swift
public struct MyError: Error {}

struct MyCommand: ParsableCommand {
	@Option(transform: {
		guard let url = URL(string: $0) else {
			throw MyError()
		}
		return url
	})
	var test: URL?
}
```

### Expected behavior
No errors.

Workarounds:
- Replacing `var test: URL?` with `var test: URL? = nil`
- Or making the `transform` non-throwing

natecook1000:
Due to the restructuring in #477, there was ambiguity between the unconstrained `@Option` initializer that uses a transform (but no initial value) and the one that is constrained to the property being optional. This marks the unconstrained version as disfavored, which allows overload resolution to select the optional version when appropriate.

Fixes #618.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Option.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Option.swift
@@ -31,7 +31,7 @@
 ///
 ///     mutating func run() {
 ///         print("\(greeting) \(name)!")
-///         if let age = age {
+///         if let age {
 ///             print("Congrats on making it to the ripe old age of \(age)!")
 ///         }
 ///     }
@@ -241,20 +241,25 @@ extension ArrayParsingStrategy: Sendable { }
 
 // MARK: - @Option T: ExpressibleByArgument Initializers
 extension Option where Value: ExpressibleByArgument {
-  /// Creates a property with a default value provided by standard Swift default value syntax.
+  /// Creates a property with a default value that reads its value from a
+  /// labeled option.
   ///
-  /// This method is called to initialize an `Option` with a default value such as:
+  /// This initializer is used when you declare an `@Option`-attributed property
+  /// that has an `ExpressibleByArgument` type, providing a default value:
+  ///
   /// ```swift
-  /// @Option var foo: String = "bar"
+  /// @Option var title: String = "<Title>"
   /// ```
   ///
   /// - Parameters:
   ///   - wrappedValue: A default value to use for this property, provided
-  ///   implicitly by the compiler during property wrapper initialization.
-  ///   - name: A specification for what names are allowed for this flag.
-  ///   - parsingStrategy: The behavior to use when looking for this option's value.
+  ///     implicitly by the compiler during property wrapper initialization.
+  ///   - name: A specification for what names are allowed for this option.
+  ///   - parsingStrategy: The behavior to use when looking for this option's
+  ///     value.
   ///   - help: Information about how to use this option.
-  ///   - completion: Kind of completion provided to the user for this option.
+  ///   - completion: The type of command-line completion provided for this
+  ///     option.
   public init(
     wrappedValue: Value,
     name: NameSpecification = .long,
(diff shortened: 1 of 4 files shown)


natecook1000:
@swift-ci Please test

Coeur:
Thank you Nate.
To be exhaustive, we also need to fix Argument.swift, and add `@_disfavoredOverload` line 417, and add the same test replacing `@Option` with `@Argument`.

rauhul (approved):
Thanks for the fix! I would've expected the previous tests I added to catch this

natecook1000:
@swift-ci Please test

Coeur (approved):

Coeur:
Question: to facilitate the understanding of the documentation, should we make such below changes?

Option.swift, Line 508, change:
```
  /// var foo: String = "bar"
```
To:
```suggestion
  /// var foo: String? = "bar"
```

Option.swift, Line 577, change:
```
  /// var foo: String
```
To:
```suggestion
  /// var foo: String?
```

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -294,18 +299,22 @@ extension Option where Value: ExpressibleByArgument {
...
-  /// This method is called to initialize an `Option` without a default value such as:
   /// ```swift
-  /// @Option var foo: String
+  /// @Option var name: String
   /// ```
   ///
   /// - Parameters:
-  ///   - name: A specification for what names are allowed for this flag.
-  ///   - parsingStrategy: The behavior to use when looking for this option's value.
+  ///   - name: A specification for what names are allowed for this option.
```
Coeur, line 312: The rename from `foo` to `name` may add to confusion, since there is also a `name` argument in the initializer.

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -498,24 +533,28 @@ extension Option {
 
 // MARK: - @Option Optional<T> Initializers
 extension Option {
-  /// Creates a property with a default value provided by standard Swift default
-  /// value syntax, parsing with the given closure.
+  /// Creates an optional property that reads its value from a labeled option,
+  /// parsing with the given closure, with an explicit `nil` default.
```
Coeur, line 537: In general, maybe yes. But it would also work with non-nil default, like:
```
var char: Character? = " "
```
The fact that it's a mutable optional would then allow it to eventually become nil later in time (when a program needs to free some memory for instance).

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -639,19 +697,23 @@ extension Option {
...
+  /// more labeled options.
+  ///
+  /// This initializer is used when you declare an `@Option`-attributed array
+  /// property without a default value:
   ///
-  /// This method is called to initialize an array `Option` with no default value such as:
   /// ```swift
-  /// @Option()
-  /// var foo: [String]
+  /// @Option var char: [Character]
```
Coeur, line 707: nit, plural: char -> chars

Same thing on lines 748 and 796

Sources/ArgumentParser/Parsable Properties/Option.swift
natecook1000: That way of declaring (optional with a non-`nil` default) ends up using a different initializer that has a deprecation warning, since the behavior usually isn't what the author intends. I hadn't considered the use case you're describing here, but in general the ArgumentParser API is designed to essentially provide a way to write a function interface via the command line; the type declarations should match the CL interface.

natecook1000:
@swift-ci Please test

PR #629, opened 2024-04-04
Abort on ambiguous subcommand names

dcantah:
Currently it's possible to have two commands with the same _commandName by mixing a combination of the automatic command name generation based on the type name, and explicit use of `CommandConfiguration`'s commandName parameter. For example, if we edit the math examples `Add` command to have a commandName of "multiply" this is allowed currently:

```
./.build/debug/math multiply 2 3
5
```

The behavior today is whatever subcommand is registered in the tree first is what will resolve for that command. So, `Multiply` in this example is permanently shadowed.

This change just makes sure there's no occurence of this happening in any level of the tree of potential subcommands, and if so we'll abort early. 

#### Note
I can't imagine this happens too often, but the detection logic will also be useful to verify clashing aliases+subcommands for this change as well https://github.com/apple/swift-argument-parser/pull/627.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Utilities/Tree.swift
+++ b/Sources/ArgumentParser/Utilities/Tree.swift
@@ -90,7 +90,11 @@ extension Tree where Element == ParsableCommand.Type {
   
   convenience init(root command: ParsableCommand.Type) throws {
     self.init(command)
+    var subcommandNames = Set<String>()
     for subcommand in command.configuration.subcommands {
+      if !subcommandNames.insert(subcommand._commandName).inserted {
+        throw InitializationError.matchingCommandNames(command, subcommand._commandName)
+      }
       if subcommand == command {
         throw InitializationError.recursiveSubcommand(subcommand)
       }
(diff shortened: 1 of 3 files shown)


Sources/ArgumentParser/Utilities/Tree.swift
```diff
@@ -90,7 +90,12 @@ extension Tree where Element == ParsableCommand.Type {
   
   convenience init(root command: ParsableCommand.Type) throws {
     self.init(command)
+    var subcommandNames = Set<String>()
     for subcommand in command.configuration.subcommands {
+      if subcommandNames.contains(subcommand._commandName) {
+        throw InitializationError.matchingCommandNames(command, subcommand._commandName)
+      }
+      subcommandNames.insert(subcommand._commandName)
```
rauhul, line 98: Tiny note, insert returns a Bool indicating if the new value was inserted, you can use this instead of checking contains

dcantah: A beautiful API 🙏, will change to this

dcantah: Done!

dcantah:
@swift-ci please test

dcantah:
Let me draft this until I have time to fix up the way we validate this

dcantah:
Closing this as adding some validations here seems very tricky, if we do want this I'm glad to brainstorm with someone

PR #636, opened 2024-05-01
Eliminate warnings re: CommandConfiguration init

natecook1000:
When the old (pre-aliases) initializer has all its default parameter
values, it is selected as the overload because it has fewer parameters
overall. Removing the default parameters allows it to still satisfy
(very niche) source compat requirements without actually being
available as an overload.

### Checklist


natecook1000:
@swift-ci Please test

PR #638, opened 2024-05-03
Prefer if-let shorthand in `CountLines.swift`

KS1019:
`CountLines.swift` uses the mix of the normal if-let and the shorthand. This PR changes the normal version to the shorthand to stay consistent.

### Checklist
- N/A I've added at least one test that validates that my change is working, if appropriate
- N/A I've updated the documentation if necessary

### DIFF
--- a/Examples/count-lines/CountLines.swift
+++ b/Examples/count-lines/CountLines.swift
@@ -31,7 +31,7 @@ struct CountLines: AsyncParsableCommand {
 extension CountLines {
     var fileHandle: FileHandle {
         get throws {
-            guard let inputFile = inputFile else {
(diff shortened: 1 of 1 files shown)


natecook1000:
Looks good to me, now that we're on 5.7 as the minimum Swift version. Thanks, @KS1019!

natecook1000:
@swift-ci Please test

rauhul:
We can merge this, but I personally really dislike this syntax because it doesn't work with property chaining. This becomes a deal breaker for me because explicit `self` is a _must_ imo.

rauhul:
@swift-ci test

Issue #633, opened 2024-04-23
Add ability to infer root command name from CLI args


bisgardo:
The command name of a (`Async`)`ParsableCommand` is [inferred from the name of the struct](https://apple.github.io/swift-argument-parser/documentation/argumentparser/commandconfiguration/commandname) by default.

For the root command, this name is never actually part of a command; it's only used in the usage string. 
You can of course override it manually, but you never know if the executable gets renamed.

In my understanding, it's common for the usage string to use the actual name the command was invoked with (i.e. `argv[0]` in C terms). And I also think that's the only appropriate default behavior.

Issue #570, opened 2023-05-20
Wrong executable name reported in `usage`.


dabrahams:
Executable name in usage messages is synthesized from the `@main` struct name.  That's incorrect by almost any measure: no executable by that name exists in most cases.  IMO the right behavior is to report the executable name as `CommandLine.arguments.first!`.  There are at least several command-line packages that are installed as a family of symlinks to a single executable, where the behavior of the command may actually depend on the name of the symlink through which it was invoked.

Replace this paragraph with a short description of the incorrect incorrect behavior. If this is a regression, please note the last version that the behavior was correct in addition to your current version.

**ArgumentParser version:** 1.2.2
**Swift version:** swift-driver version: 1.75.2 Apple Swift version 5.8 (swiftlang-5.8.0.124.2 clang-1403.0.22.11.100)
Target: arm64-apple-macosx13.0

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

Use this source to create a command-line tool with executable named "foo," then run the tool with no arguments

```
import ArgumentParser
import Foundation

@main
struct TesterOfThings: ParsableCommand {
  @Argument(help: "Paths.", transform: URL.init(fileURLWithPath:))
  var files: [URL]

  func run() throws {
    print(CommandLine.arguments)
  }
}
```

### Expected behavior
A message is printed that describes the usage of the `foo` command.

### Actual behavior
A message is printed that describes the usage of the nonexistent `tester-of-things` command.

rauhul:
Definitely a bug, I haven't at the command cluster support (e.g. `git-x` invocable as `git x`) code in a while so I'm not sure sure how much work this will take to fix offhand, though my guess is that this shouldn't be too hard.

natecook1000:
See also https://github.com/apple/swift-argument-parser/issues/295

natikgadzhi:
It looks like if the `ParsableCommand` has `configuration` that contains `usage` or `commandName`, those will be used instead of defaulting to `ParsableCommand.Type`.

`HelpGenerator` [builds `toolName` from `ParsableCommand.Type`](https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Usage/HelpGenerator.swift#L104), and then we pass that into `UsageGenerator`. 

[...]

Issue #295, opened 2021-04-03
USAGE command name should always reflect the actual binary name


eaigner:
The default `USAGE` output derives the usage command name from the `ParsableCommand` type name. Now this might be correct or not. The "correct" way would be to always use the last path component of `argv[0]`, as it would also show the correct string in case the build product name is different from the class name or the user renames the tool.

While there is `_commandName` I can use to override this, it is marked as for internal use only.

frou:
Another common occurrence of this is running the executable via a symlink, where the symlink has a different basename to the executable it's pointing to.

I guess `ProcessInfo.processInfo.processName` from Foundation already has to logic for processing `argv[0]`?

amomchilov:
@frou TIL about `ProcessInfo.processInfo.processName`, thanks for sharing!

I worked around this with:

```swift
	static var _commandName: String {
		// Workaround for https://github.com/apple/swift-argument-parser/issues/570
        ProcessInfo.processInfo.processName
	}
```

I know it's "for internal use only", but there's no other way. Mia culpa :)

https://github.com/apple/swift-argument-parser/blob/fd0db17fa290b350a06b937ba55c5e850bbfeb25/Sources/ArgumentParser/Parsable%20Types/ParsableCommand.swift#L18-L23

PR #641, opened 2024-05-31
Help text: Display argv[0] for root command by default

dcantah:
Fixes: #633 #570 #295
    
This seems like the "correct" default, although I'm sure there's some crazy edge case I'm failing to take into account. If your root command had the same name as the resulting binary you produce, this was never an issue, however if your root command was named anything else, then the current Usage/help text was kind of odd.

If you don't specify an explicit commandName in CommandConfiguration, we would take the root command name and generate a command name for you out of hyphens. So, `RootCommand` would become `root-command` and this is what would be displayed in the usage text for `/path/to/cool-binary --help`, even though intuitively you'd expect to see `USAGE: cool-binary`.

The current behavior was also strange for binaries that are classically invoked via symlinks as the same thing would happen. In this case imagine the struct name matched the actual binaries name, but because the symlink is what users actually
invoke you could end up seeing an unfamiliar name in help text regardless. Using argv[0] solves most of these problems.

The downside here is a LOT of tests need to change. I went with foregoing the new approach if the user explicitly provides a command name, so most of the existing tests that check for exact string matches of help text either need to replace the
command name with the first argument of the binary that is running the test (xctest, but it doesn't really matter what it is as I've added a helper to plop in the first argument for all of these tests), or they need to define CommandConfiguration(commandName: "name-we-want"). For any tests that implemented ParsableCommand I mostly went with the latter to make as few changes as possible.

Given this seems like mostly a UX change, I hope this seems sane.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
+++ b/Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
@@ -15,6 +15,10 @@ public struct CommandConfiguration: Sendable {
   ///
   /// If `nil`, the command name is derived by converting the name of
   /// the command type to hyphen-separated lowercase words.
+  /// 
+  /// NOTE: In help and usage texts if this is the root command and commandName is
+  /// `nil`, the name shown will be derived from the binary being invoked (argv[0]),
+  /// otherwise `commandName` will be displayed.
   public var commandName: String?
 
   /// The name of this command's "super-command". (experimental)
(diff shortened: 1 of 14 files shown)


dcantah:
I missed https://github.com/apple/swift-argument-parser/pull/501 from @smvz that already exists.. It seems we went in two directions though, his change is opt-in while this is changing the default behavior of the root command. His likely makes more sense from a back compat standpoint (and has the benefit of not needing all the test changes).

PR #644, opened 2024-06-02
Introduce subcommand grouping into the command configuration to improve help

DougGregor:
Add optional support for grouping subcommands into named groups, to help bring order to commands with many subcommands without requiring additional structure. For example, here's the help for a "subgroupings" command that has an ungrouped subcommand (m), and two groups of subcommands ("broken" and "complicated").

    USAGE: subgroupings <subcommand>

    OPTIONS:

    SUBCOMMANDS:
      m

    BROKEN SUBCOMMANDS:
      foo                     Perform some foo
      bar                     Perform bar operations

    COMPLICATED SUBCOMMANDS:
      n

      See 'subgroupings help <subcommand>' for detailed help.

The help output above is created like this:

```
  struct WithSubgroups: ParsableCommand {
    static let configuration = CommandConfiguration(
      commandName: "subgroupings",
      subcommands: [ M.self ],
      groupedSubcommands: [
        CommandGroup(
          name: "Broken",
          subcommands: [ Foo.self, Bar.self ]
        ),
        CommandGroup(name: "Complicated", subcommands: [ N.self ]),
        CommandGroup(name: "Invisible", subcommands: [ ])
      ]
    )
  }
```

Each `CommandGroup` names a new group with its subcommands (there are no groups within groups).

This structure is only cosmetic, and only affects help generation by providing more structure for the reader. It doesn't impact existing clients, who can still reason about the flattened list of subcommands if they prefer.


DougGregor:
@swift-ci please test

dickoff:
I was just looking at doing something very similar! (WIP here: https://github.com/apple/swift-argument-parser/tree/subcommand-groups).

The CommandsBuilder idea is an interesting one. However is it limiting in some of the manipulation that you can do with the existing simple array `subcommands` parameter? Currently it's quite easy to do things like dynamically alphabetically sort the subcommands array or only include some commands based on `#if` checks. I checked out this branch and played around with a it a bit and it wasn't obvious to me how you could achieve those kinds of things with the builder (but I'm very inexperienced with resultBuilders!)

dickoff:
Do we think there's value in allowing for a longer description string for the group? For example how subcommand groups are described in `git -h`. E.g. accomplishing something like this in the test, adding a description to the header title.

```
BROKEN SUBCOMMANDS: These commands are currently non-functional due to XYZ
  foo                     Perform some foo
  bar                     Perform bar operations
  ```

**Comment by DougGregor** [comment]
> The CommandsBuilder idea is an interesting one. However is it limiting in some of the manipulation that you can do with the existing simple array `subcommands` parameter?

It shouldn't be too limiting. Result builders can support more language syntax if we'd like (e.g., the ability to do `if #available` and such), although I didn't add them.

> I checked out this branch and played around with a it a bit and it wasn't obvious to me how you could achieve those kinds of things with the builder (but I'm very inexperienced with resultBuilders!)

I'd expect `#if` to work inside the builder already. If we want `if`, `if`-`else`, `switch`, and `if #available` to work, we can add that support to the builder by introducing the appropriate `buildXYZ` methods.

> Do we think there's value in allowing for a longer description string for the group? 

Yes, I think that's a good idea! It's easy enough to add an "abstract" that we put here.

**Comment by DougGregor** [comment]
@swift-ci please test

**Comment by DougGregor** [comment]
@dickoff See the latest two commits, which add an optional abstract and then extend out the result builders to handle all of the declarative syntax.

**Comment by DougGregor** [comment]
I've opened up a discussion thread on the forums for this change: https://forums.swift.org/t/grouping-subcommands/72219

File: Sources/ArgumentParser/Usage/HelpGenerator.swift [hunk 1]
```diff
@@ -76,8 +88,14 @@ internal struct HelpGenerator {
       guard !elements.isEmpty else { return "" }
       
       let renderedElements = elements.map { $0.rendered(screenWidth: screenWidth) }.joined()
-      return "\(String(describing: header).uppercased()):\n"
-        + renderedElements
+
+      var renderedHeader = String(describing: header).uppercased() + ":"
+      if let abstract = header.abstract {
+        renderedHeader += " "
+        renderedHeader += abstract.wrapped(to: screenWidth)
```
natecook1000, line 95: Do we need to add the abstract and then wrap?

dickoff:
> @dickoff See the latest two commits, which add an optional abstract and then extend out the result builders to handle all of the declarative syntax.

New changes look great! They resolve all the questions I brought up.

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -211,34 +229,77 @@ internal struct HelpGenerator {
...
+      }
+    )
+
     // Combine the compiled groups in this order:
     // - arguments
     // - named sections
     // - options/flags
-    // - subcommands
+    // - grouped subcommands
+    // - ungrouped subcommands
```
natecook1000, line 294: Looks like this order is flipped from the actual behavior.

Sources/ArgumentParser/Parsable Types/SubcommandsBuilder.swift
```diff
@@ -0,0 +1,65 @@
...
+    return component
+  }
+
+  public static func buildArray(_ components: [[Subcommand]]) -> [Subcommand] {
+    return components.flatMap { $0 }
+  }
+}
+
+/// Describes a single subcommand or a group thereof.
+public enum Subcommand: Sendable {
```
natecook1000, line 49: This name doesn't quite describe what the type represents, since a group of subcommands is not itself a subcommand. One option is `CommandGroup` – I'm more comfortable with a `CommandGroup` that just stores a single command than a `Subcommand` that stores a bunch of commands. But then we couldn't use that name for the other type, which it's perfect for.

Maybe just making it `Subcommands` is an improvement? I don't think this is a dealbreaker – do you have any other ideas for this name?

Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
```diff
@@ -514,6 +514,54 @@ extension HelpGenerationTests {
...
+    to XYZ.
+      foo                     Perform some foo
+      bar                     Perform bar operations
+
+    COMPLICATED SUBCOMMANDS:
+      n
+
+      See 'subgroupings help <subcommand>' for detailed help.
+    """)
+  }
```
natecook1000, line 564: This output looks terrific! From what I can see, empty command groups are allowed, but essentially ignored in the output, which is the right behavior since we're supporting conditionals in the builder syntax. Can we add a test with one of those to verify? Also a test with _only_ grouped subcommands, to make sure we aren't missing anything in that case.

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -62,6 +76,58 @@ public struct CommandConfiguration: Sendable {
...
+  ///     command.
+  ///   - defaultSubcommand: The default command type to run if no subcommand
+  ///     is given.
+  ///   - helpNames: The flag names to use for requesting help, when combined
+  ///     with a simulated Boolean property named `help`. If `helpNames` is
+  ///     `nil`, the names are inherited from the parent command, if any, or
+  ///     are `-h` and `--help`.
+  ///   - aliases: An array of aliases for the command's name. All of the aliases
+  ///     MUST not match the actual command name, whether that be the derived name
+  ///     if `commandName` is not provided, or `commandName` itself if provided.
```
natecook1000, line 106: Something like this:
```suggestion
  ///     if `commandName` is not provided, or `commandName` itself if provided.
  ///   - groupedSubcommands: A builder closure that defines the subcommands
  ///     and their groups for this command.
```

rauhul:
I would recommend splitting out the abstract into a different PR. If we add this functionality to CommandGroups then we should probably add them OptionGroups and that feels like scope creep for this change.

natecook1000 (commented):
Thanks so much for this, @DougGregor! In addition to the notes below, could you add initializers that just accept arrays that are parallel to the builder-based initializers for `CommandConfiguration` and `CommandGroup`? That will be more flexible for projects that set up the command hierarchy out of place with the command declarations themselves.

Sources/ArgumentParser/Usage/HelpGenerator.swift
DougGregor: Ahhh, probably. Thank you

DougGregor:
> could you add initializers that just accept arrays that are parallel to the builder-based initializers for `CommandConfiguration` and `CommandGroup`?

Hmm. If we're going to do this, I think I'd rather drop the builder-based initializers entirely from this pull request than introduce two new interfaces for the same thing. It's completely reasonable to say that the only way to get this functionality is with syntax like this:

```swift
subcommands: [
  // all of the ungrouped subcommands
],
groupedSubcommands: [
  CommandGroup(
    name: "group name",
    subcommands: [
      // subcommands in this group
    ]
  ),
  // ...
].
```

That avoids having to create the `Subcommand` type at all, and is arguably more in keeping with the current API design. It would leave the door open to a builder-based design in the future should we want it.

natecook1000:
That makes sense! I could see that fitting into a whole builder-based configuration approach (which I think we've looked at before). If we make this change, let's also make `flattenedSubcommands` (or `allSubcommands`) public API as well, since the subcommands will live in two different places.

dickoff:
> That makes sense! I could see that fitting into a whole builder-based configuration approach (which I think we've looked at before). If we make this change, let's also make `flattenedSubcommands` (or `allSubcommands`) public API as well, since the subcommands will live in two different places.

Could we not just have the existing `subcommands` still return all ungrouped and grouped subcommands together? Having `subcommands` start to only return ungrouped commands would be odd for clients I would think.

DougGregor:
> Could we not just have the existing `subcommands` still return all ungrouped and grouped subcommands together?

I think we need `subcommands` to still return all ungrouped and grouped subcommands. I can add `ungroupedSubcommands` and `groupedSubcommands` for clients that want them separate.

DougGregor:
@swift-ci please test

DougGregor:
@natecook1000 I believe I addressed all of your comments. The updated PR description contains the non-result-builder interface, which is a more straightforward extension of the existing syntax:

```swift
  struct WithSubgroups: ParsableCommand {
    static let configuration = CommandConfiguration(
      commandName: "subgroupings",
      subcommands: [ M.self ],
      groupedSubcommands: [
        CommandGroup(
          name: "Broken",
          subcommands: [ Foo.self, Bar.self ]
        ),
        CommandGroup(name: "Complicated", subcommands: [ N.self ]),
        CommandGroup(name: "Invisible", subcommands: [ ])
      ]
    )
  }
```

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -47,8 +47,27 @@ public struct CommandConfiguration: Sendable {
   public var shouldDisplay: Bool
   
   /// An array of the types that define subcommands for this command.
-  public var subcommands: [ParsableCommand.Type]
-  
+  ///
```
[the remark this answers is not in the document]
DougGregor: I was using the extra line to separate the "abstract" line from the "detail" line.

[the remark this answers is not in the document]
rauhul: Wow I can't this morning, I made like 2 other mistakes similar, sorry about the noise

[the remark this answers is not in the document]
natecook1000: It's a weird diff!

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -47,8 +47,27 @@ public struct CommandConfiguration: Sendable {
...
+      ungroupedSubcommands = newValue
+    }
+  }
+
+  /// An array of types that define subcommands for this command and are
+  /// not part of any command group.
+  public var ungroupedSubcommands: [ParsableCommand.Type]
+
+  /// The list of subcommands and subcommand groups.
+  public var groupedSubcommands: [CommandGroup]
```
natecook1000, line 69: These look great 👍🏻

natecook1000 (approved):
Looks great, thanks again!

Issue #637, opened 2024-05-02
Add option to retrieve a machine-readable list of possible arguments


matthewbastien:
### Description

The [vscode-swift](https://github.com/swift-server/vscode-swift) extension recently added support for [creating a new swift project](https://github.com/swift-server/vscode-swift/pull/755). It does so using `swift package init` with some user input to initialize the project with. Right now, it relies on parsing the output from `swift package init --help` to find the list of available project types for the `--type` option. However, this isn't really an ideal solution as the format could presumably change at any point (and has already drastically changed between swift 5.3 and 5.10).

Here's what the output looks like in Swift 5.8:
```
swift package init --help
OVERVIEW: Initialize a new package

USAGE: swift package init <options>

OPTIONS:
  --type <type>           Package type: (default: library)
        library           - A package with a library.
        executable        - A package with an executable.
        tool              - A package with an executable that uses
                            Swift Argument Parser. Use this template if you
                            plan to have a rich set of command-line arguments.
        empty             - An empty package with a Package.swift manifest.
  --name <name>           Provide custom package name
  --version               Show the version.
  -h, -help, --help       Show help information.
```

It would be nice to have a more reliable output format (e.g. JSON) that contains the enumerations for the `--type` parameter. No additional information is needed beyond what's already available in the output - the extension only needs the enum value as well as a brief description (a human readable label would be nice to include, but is not necessary as the enum value can be parsed rather easily).

Seeing as this is something that might be beneficial for other commands: it might even be a good idea to add this as a feature of swift argument parser. However, I'm raising an issue here first since this is the primary use case for the vscode-swift extension.

bnbarham:
> All the options in the JSON above should be feasible to be generated by the properties declared in the user's struct that inherits from ParsableArguments.

This is why I mentioned `--type` above, because its documentation comes from `discussion` and so the proposed `"description"` in your JSON wouldn't actually be possible in that case - there's no way to split those values at the moment.

natecook1000:
It seems like the request for a machine-readable list of options is satisfied by the `dump-help` functionality. Support for more structure in the help around values is a separate request that will need its own design – @matthewbastien, could you open that as a separate issue?

PR #647, opened 2024-06-05
Add ability to retrieve list of possible options for Option property

bripeticca:
Previously, information regarding a custom Option type's possible values was lost when generating subsequent structures in the pipeline to describe the argument. This would largely have to be defined by the user in the `discussion` string property in `ArgumentHelp` and would provide no mechanism to directly access the allowable values for a custom `Option`.

This adds an extra static property on `ExpressibleByArgument` labelled `defaultValueDescriptions`, which maps the value name to its description. It is up to the user to implement the `defaultValueDescription` for their custom `CaseIterable` type to fully support this behaviour. Further refactoring has been made downstream to the ArgumentDefinition and subsequent structures describing the argument to allow the ability to retrieve the value descriptions from the `ExpressibleByArgument` type, making information about the property more visible and accessible.

A new enumeration `ArgumentDiscussion` has been introduced to better encapsulate the discussion information for an argument, and includes cases for discussion sections that contain only static text, as well as cases that include a list of argument values that derive from a type that implements `ExpressibleByArgument`, `CaseIterable` and `RawRepresentable`.

Addresses #637

### Checklist


Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -239,6 +239,129 @@ public struct ArrayParsingStrategy: Hashable {
 
 extension ArrayParsingStrategy: Sendable { }
 
+// TODO: to format and create separate file
```
cmcgee1024, line 242: thought: I agree that this is probably best considering how the code base is structured.

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -239,6 +239,129 @@ public struct ArrayParsingStrategy: Hashable {
...
+
+// MARK: - @Option T: EnumerableOption Initializers
+extension Option where Value: EnumerableOption {
+    /// Creates a property with a default value that reads its value from a
+    /// labeled option.
+    ///
+    /// This initializer is used when you declare an `@Option`-attributed property
+    /// that has an `EnumerableOption` type, providing a default value:
+    ///
+    /// ```swift
```
cmcgee1024, line 272: praise: This is a concise and illustrative example and well placed.

Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
```diff
@@ -78,7 +78,7 @@ fileprivate extension ArgumentSet {
         argument.help.valueName = group.map(\.valueName).first { !$0.isEmpty } ?? ""
         argument.help.defaultValue = group.compactMap(\.help.defaultValue).first
         argument.help.abstract = group.map(\.help.abstract).first { !$0.isEmpty } ?? ""
-        argument.help.discussion = group.map(\.help.discussion).first { !$0.isEmpty } ?? ""
+        argument.help.discussion = group.map(\.help.discussion).first { !$0.isEmpty } ?? .staticString("")
```
cmcgee1024, line 81: thought: I wonder with these empty static strings if it might be worth having either an .empty in the Discussion enum or make these into optional types.

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -18,19 +18,72 @@ internal struct HelpGenerator {
     struct Element: Hashable {
       var label: String
       var abstract: String = ""
-      var discussion: String = ""
-      
+      var discussion: Discussion = .staticString("")
```
cmcgee1024, line 21: thought: Maybe this could be an optional discussion type?

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -18,19 +18,72 @@ internal struct HelpGenerator {
...
+      // TODO: to format
       func rendered(screenWidth: Int) -> String {
         let paddedLabel = self.paddedLabel
         let wrappedAbstract = self.abstract
           .wrapped(to: screenWidth, wrappingIndent: HelpGenerator.labelColumnWidth)
-        let wrappedDiscussion = self.discussion.isEmpty
+
+        var wrappedDiscussion = ""
+
+        if case let .staticString(discussionText) = discussion {
```
cmcgee1024, line 56: thought: If discussion is optional then this can be a regular if-left.

Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -90,7 +143,6 @@ internal struct HelpGenerator {
   var abstract: String
   var usage: String
   var sections: [Section]
-  var discussionSections: [DiscussionSection]
```
cmcgee1024, line 93: question: What happened to the discussion sections? I'm guessing they were vestigial.

Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -42,6 +42,69 @@ public struct ToolInfoV0: Codable, Hashable {
...
+    public var isEmpty: Bool {
+      switch self {
+      case .staticText(let s):
+        return s.isEmpty
+      case .enumerated(let values):
+        return values.isEmpty
+      }
+    }
+
+    public var nonEmpty: Self? {
```
cmcgee1024, line 104: thought: Same here, maybe nonEmpty isn't needed with optionality of Discussions.

Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -117,6 +180,70 @@ public struct ArgumentInfoV0: Codable, Hashable {
     case flag
   }
 
+  public enum DiscussionV0: Codable, Hashable {
```
cmcgee1024, line 183: thought: There's quite a few discussion enums around. Is it worth making one that's shared? It could be that a generic type is needed to suit the different inits, so that might need to be weighed against having the duplicated enums.

Tests/ArgumentParserUnitTests/DumpHelpGenerationTests.swift
```diff
@@ -59,15 +59,73 @@ extension DumpHelpGenerationTests {
     @OptionGroup(title: "Other")
     var options: Options
   }
-  
+
+  struct C: ParsableCommand {
```
cmcgee1024, line 63: question: Should there be a test that covers default values of enumerable option too?

cmcgee1024 (commented):
Overall this looks good to me. I've added thoughts and questions along the way. I don't feel comfortable enough with this repo to approve at this time.

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -37,21 +55,21 @@ public struct CommandConfiguration: Sendable {
 
   /// A longer description of this command, to be shown in the extended help
```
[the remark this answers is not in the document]
bripeticca: Good catch! Thanks.

Sources/ArgumentParser/Parsing/ArgumentDefinition.swift
```diff
@@ -45,13 +45,27 @@ struct ArgumentDefinition {
       static let isRepeating = Options(rawValue: 1 << 1)
     }
 
+    enum Discussion {
+      case staticString(String)
```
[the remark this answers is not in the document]
bripeticca: Ah, good catch. I missed the naming on this one - thanks! Fixed.

Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
bripeticca: I think making Discussion optional is the way to go! Fixed :)

Sources/ArgumentParser/Usage/HelpGenerator.swift
bripeticca: Fixed

Sources/ArgumentParser/Usage/HelpGenerator.swift
bripeticca: Ah, I think I accidentally deleted them. I didn't see them being used anywhere, so I may be able to remove it entirely. Will investigate as I continue this PR.

Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -42,6 +42,69 @@ public struct ToolInfoV0: Codable, Hashable {
...
+        return
+      }
+      if let values = try? container.decode([ValueV0].self) {
+        self = .enumerated(values)
+        return
+      }
+      throw DecodingError.typeMismatch(DiscussionV0.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Unexpected assosiated type for DiscussionV0."))
+    }
+
+    public var isEmpty: Bool {
```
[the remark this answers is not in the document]
bripeticca: Yeah, I think this can get redundant and it's probably better to just have Discussion properties be optional - this way I can just check if the property exists or not and assert that it contents are not empty on creation.

Sources/ArgumentParserToolInfo/ToolInfo.swift
bripeticca: Agreed!

Sources/ArgumentParserToolInfo/ToolInfo.swift
bripeticca: I think that's a good idea. In the process of fixing this, and have refactored a few of the redundant structs.

Tests/ArgumentParserUnitTests/DumpHelpGenerationTests.swift
bripeticca: Indeed - this case should be covered by the `defaultColor` property in this command! :) Was there a more robust test case you were thinking of that I missed?

Sources/ArgumentParser/Usage/HelpGenerator.swift
bripeticca: I may keep this as is just because I'll have to access the associated values for each case - did you have another approach in mind? :)

natecook1000 (commented):
Thanks so much for looking at this, @bripeticca! The updated help screens look really clear, and it's great that adoption is all opt in. I think we can cut back on the new API a little bit – if you agree, we just have a protocol name to discuss. Thanks!

Sources/ArgumentParser/Parsable Types/CommandConfiguration.swift
```diff
@@ -37,15 +37,19 @@ public struct CommandConfiguration: Sendable {
...
   /// display.
-  public var discussion: String
-  
+  ///
+  /// Can include specific abstracts about the argument's possible values (e.g.
+  /// for a custom `EnumerableOptionValue` type), or can describe
+  /// a static block of text that extends the description of the argument.
+  public var discussion: ArgumentDiscussion?
```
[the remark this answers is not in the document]
bripeticca: Ah, good point! I had changed this as I was making my way through the `DumpHelpGenerator` and updating the inits for `CommandInfoV0`, `ArgumentInfoV0`, etc. to consider `ArgumentDiscussion`. Along the way, I updated each struct that contained a `discussion` string to instead use `ArgumentDiscussion` - it seems like for `CommandConfiguration` this is unnecessary though! Since it's a breaking change, I can revert.

Sources/ArgumentParser/Parsable Types/EnumerableOptionValue.swift
```diff
@@ -0,0 +1,91 @@
...
+///         return "Red Color"
+///       case .blue:
+///         return "Blue Color"
+///       case .yellow:
+///         return "Yellow Color"
+///     }
+///   }
+/// }
+/// ```
+public protocol EnumerableOptionValue: CaseIterable, ExpressibleByArgument, RawRepresentable where RawValue: ExpressibleByArgument, AllCases == [Self] {
```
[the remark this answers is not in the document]
bripeticca: That's a great point! The naming is definitely tricky :) Perhaps something a little more generic like `EnumerableValue` could be used instead?

Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -281,6 +281,67 @@ extension Option where Value: ExpressibleByArgument {
...
+  ///     be appended to the top of the list of `EnumerableOptionValue` descriptions.
+  ///   - completion: The type of command-line completion provided for this
+  ///     option.
+  public init(
+    wrappedValue: Value,
+    name: NameSpecification = .long,
+    parsing parsingStrategy: SingleValueParsingStrategy = .next,
+    help: ArgumentHelp? = nil,
+    completion: CompletionKind? = nil
+  ) where Value: EnumerableOptionValue {
```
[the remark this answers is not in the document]
bripeticca: For sure - I just tried this out locally by adding the following to the beginning of each `ExpressibleByArgument` init:

```swift
var help = help
if let type = Value.self as? (any EnumerableOptionValue.Type) {
  help = ArgumentHelp.init(
  help?.abstract ?? "",
  discussion: help?.discussion,
  options: type, // retaining necessary type information for enumerable values here
  valueName: help?.valueName,
  visibility: help?.visibility ?? .default
  )
}
```

This way we can still retain the modified `ArgumentHelp` for types implementing `EnumerableOptionValue` (or whatever we decide to rename it to!). 

Alternatively, I can modify the `ArgumentHelp` init to take the `options` parameter and have it be optional, and can remove some code so it comes down to this:

```swift
// ArgumentHelp init:
public init(
  _ abstract: String = "",
  discussion: String? = nil,
  valueName: String? = nil,
  options: (any EnumerableOptionValue.Type)? = nil, // made this optional
  visibility: ArgumentVisibility = .default)
{
  self.abstract = abstract
  self.discussion = discussion
  self.valueName = valueName
  self.options = options
  self.visibility = visibility
} 
```

and the resulting `ExpressibleByArgument` inits would have some additional code that looks something like this:

```swift
self.init(_parsedValue: .init { key in
  let arg = ArgumentDefinition(
    container: Bare<Value>.self,
    key: key,
    kind: .name(key: key, specification: name),
    help: .init(
      help?.abstract ?? "",
      discussion: help?.discussion,
      valueName: help?.valueName,
      options: Value.self as? (any EnumerableOptionValue.Type),
      visibility: help?.visibility ?? .default
    ),
    parsingStrategy: parsingStrategy.base,
    initial: wrappedValue,
    completion: completion)

  return ArgumentSet(arg)
})
```

If this seems good to you, I can go ahead and change it!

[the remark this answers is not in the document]
natecook1000: Sorry for the delay here! I've been wrestling with whether we can avoid the `EnumerableOptionValue` protocol altogether, since it seems to be a refinement of the `ExpressibleByArgument` protocol that we're already handling for a couple slightly different uses. In particular, adding features like this one to the library is challenging in that we need to design a source-compatible addition that makes some kind of intuitive sense with the features that are already present.

The most related existing feature supported by the `ExpressibleByArgument` protocol is the ability to automatically include value names in the help description when a type is `CaseIterable`. Since the change in this PR kind of augments that to include descriptions as well as value names, I think we can piggy back on that existing design.

In particular, what would you think about adding another customization point to `ExpressibleByArgument`:

```swift
public protocol ExpressibleByArgument {
  // ...other members...
  
  // Existing static var for value strings:

  /// An array of all possible strings that can convert to a value of this
  /// type, for display in the help screen.
  ///
  /// The default implementation of this property returns an empty array. If the
  /// conforming type is also `CaseIterable`, the default implementation returns
  /// an array with a value for each case.
  static var allValueStrings: [String] { get }

  // New static var for value descriptions:
  
  static var allValueDescriptions: [String: String] { get }
}
```

We could provide a default implementation that returns an empty dictionary, and then any type that wants to provide a more expanded help screen could implement `allValueDescriptions` to provide corresponding descriptions for each element of the `allValueStrings` array. Internally, we can use similar machinery to capture those descriptions for use when displaying the help screen or generating the JSON help description.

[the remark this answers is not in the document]
bripeticca: This is a great idea! Thanks for the feedback - I tried to implement this locally and it seems to work well.

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

Sources/ArgumentParser/Parsable Properties/ArgumentDiscussion.swift
```diff
@@ -0,0 +1,138 @@
...
+///    -h, --help           Show help information
+/// ```
+///
+/// In any case where the argument type is not `EnumerableOptionValue`, the default implementation
+/// will use the `.staticText` case and will print a block of discussion text.
+enum ArgumentDiscussion {
+  case staticText(String)
+  case enumerated(preamble: String? = nil, any ExpressibleByArgument.Type)
+
+  public init?(_ text: String? = nil, _ options: (any ExpressibleByArgument.Type)? = nil) {
```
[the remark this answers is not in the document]
bripeticca: Good catch! Thank you!

bripeticca:
@swift-ci please test

Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -50,7 +116,11 @@ public struct CommandInfoV0: Codable, Hashable {
   /// Short description of the command's functionality.
   public var abstract: String?
   /// Extended description of the command's functionality.
-  public var discussion: String?
+  ///
+  /// Can include specific abstracts about the argument's possible values (e.g.
+  /// for a custom `CaseIterable` type), or can describe
+  /// a static block of text that extends the description of the argument.
+  public var discussion: Discussion?
```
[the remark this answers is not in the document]
rauhul: Yeah this will need a version bump and will be source break. Thats not a deal breaker, but can be annoying.

Alternatively we can call this `discussion2` and deprecate the old one.

Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
```diff
@@ -969,6 +969,256 @@ extension HelpGenerationTests {
...
+
+  func testLongOptionLabelAndDescriptionHelpWithOptionDescription() {
+    AssertHelp(.default, for: LongLabelHelpWithOptionDescription.self, equals: """
+          USAGE: long-label-help-with-option-description --argument <argument>
+          
+          OPTIONS:
+            --argument <argument>   A collection of cases with varying lengths of
+                                    labels/descriptions.
+                  This discussion should make the list of options wrap differently.
+                  Values:
```
rauhul, line 1206: Is there a real need for "Values:"? I dont think I like the difference in indentation with/without a "discussion"

Tools/generate-manual/DSL/MultiPageDescription.swift
```diff
@@ -18,7 +18,17 @@ struct MultiPageDescription: MDocComponent {
   var body: MDocComponent {
     Section(title: "description") {
       if let discussion = command.discussion {
-        discussion
+        if case let .staticText(text) = discussion {
```
rauhul, line 21: Can we factor this body out into a Discussion type and use it the 4 places changed here?

rauhul (changes requested):

Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
bripeticca: No real need, I can change it!

bripeticca:
@swift-ci please test

Tools/generate-manual/DSL/MultiPageDescription.swift
bripeticca: I've added a `DiscussionText` struct to the project that encapsulates this logic!

Tests/ArgumentParserUnitTests/HelpGenerationTests.swift
```diff
@@ -969,6 +969,253 @@ extension HelpGenerationTests {
...
+    @Option(help:
+        .init(
+          "An abstract.",
+          discussion: "A discussion."
+        )
+    )
+    var b: OptionValues?
+  }
+
+  func testEnumerableValuesWithPreamble() {
```
rauhul, line 1090: Im a little confused by how this list view interacts with the change introduced in https://github.com/apple/swift-argument-parser/pull/594. Do you have a test that covers both features together?

bripeticca: I can add a test that demonstrates the differences between both of these features! 

The idea is that if the type for the `@Option` property is an enum that implements `ExpressibleByArgument` and if the enum also implements the `defaultValueDescription` property for its cases, then the generated help text will enumerate each of the enum's values with the description defined in `defaultValueDescription`. 

For example:

```
enum OptionEnumerated: String, CaseIterable, ExpressibleByArgument { 
    case one
    case two
    case three

    public var defaultValueDescription: String { 
        switch self { 
            case .one:
                return "The number one."
            case .two:
                return "The number two."
            case .three:
                return "The number three."
        }
    }
}

struct MyCommand: ParsableCommand { 
    @Option(help: "The values for this option should be enumerated with a description.")
    var list: OptionEnumerated
}

// Example generated help text:
...
--list <list>         The values for this option should be enumerated with a description.
      one              - The number one.
      two              - The number two.
      three            - The number three.
...
```

Otherwise if this property is not implemented in the enum, then it will generate the help text like so:

```
enum OptionWithoutEnumeratedHelpText: String, CaseIterable, ExpressibleByArgument { 
    case one = "1"
    case two = "2"
    case three = "3"
}

struct MyOtherCommand: ParsableCommand { 
    @Option(help: "This is an option without explicit enumeration in the help text.")
    var values: OptionWithoutEnumeratedHelpText
}

// Example generated help text:
...
 --values <values>       This is an option without explicit enumeration in the
                                       help text. (values: 1, 2, 3)
...
```

bripeticca:
@swift-ci please test

bripeticca:
@swift-ci please test

rauhul (approved):
I'm pretty happy with this change! If @natecook1000 is happy with it, we can merge :)

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
Thanks so much for your patience on this, @bripeticca! 🎉🎉🎉

PR #648, opened 2024-06-05
Fix  zsh/bash completions for arguments in option groups with a custom completion

### LINKED ISSUE #646: Zsh completion missing for arguments in an option group with custom completion
Zsh completions are not shown for arguments that are added to a command using an `@OptionGroup` and that have a `custom` completion kind.

**ArgumentParser version:** `main`
**Swift version:** 
```
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0
```

### Checklist

### Steps to Reproduce

1. Create a command with an option group with an arguments with a custom completion kind.
	```swift
	import ArgumentParser
	
	struct ArgOptionGroup: ParsableArguments {
	    @Argument(completion: .custom { _ in
	        ["1-argInOptionGroup", "2-argInOptionGroup", "3-argInOptionGroup"]
	    })
	    var argInOptionGroup: String
	}
	
	@main
	struct CompletionExample: ParsableCommand {
	    @Argument(completion: .custom { _ in
	        ["1-argInCommand", "2-argInCommand", "3-argInCommand"]
	    })
	    var argInCommand: String
	    
	    @OptionGroup var optionGroup: ArgOptionGroup
	}
	```

1. Install the completions in your terminal and load them, steps will vary per environment, for me the steps are.
	1. `swift build` 
	1. Install the completions.
		```shell
		./completion-example --generate-completion-script zsh > ~/.oh-my-zsh/completions/_completion-example
		```
    1. Open a new terminal to load the new completions.
1. Type `./completion-example ` (note the trailing space)
	

1. Press tab to accept the first completion
1. Press space.
	
1. Type `1` (the first character of one of the arguments.
	

### Expected behavior
It should suggest completions for the argument from the option group, as it did for the argument in the command (step 3).

### Actual behavior
It doesn't show any completions for the arguments in the option group (step 4). When starting to type any of the options that should be suggested (step 5), it prints that there are no matches (probably from `zsh-autosuggestions`).

CraigSiemens:
## Issue
- Fixes #646 

### The issue

It was caused by a mismatch in the completion scripts and the custom completion parsing. This is following the example from the issue above.
- The completion scripts contained the name of the argument
  ```zsh
  ':arg-in-option-group:{_custom_completion $_completion_example_commandname ---completion  -- argInOptionGroup $words}'
  ```
- The parsing of the completion command failed to find the argument using `ArgumentSet.firstPositional(named:)`.
  This was caused by the function creating an `InputKey` with just a `name` and comparing to keys in the set that have their `name` and `path` set.

  In other words
  ```swift
  InputKey(name: "argInOptionGroup", parent: nil)
  != InputKey(name: "argInOptionGroup", parent: .init(named: "optionGroup", parent: nil))
  ```

There's also another issue that would occur when a command and an option group each contain an argument with the same name. In that case the completion would always match the first one in the `ArgumentSet` even if the user is trying to get completions for the second argument.

## The fix

The completions scripts were updated to contain the full path of the argument instead of just the name. Then the completions parser can create an `InputKey` that will exactly match one in the `ArgumentSet`.

To avoid having the logic spread out, an extension was added to `InputKey` to get the `fullPathString` from a key and an init to create an `InputKey` from a `fullPathString`.

I also updated the bash completion script to use `ArgumentDefinition.customCompletionCall(_:)` that the zsh script was already using to deduplicate the logic.

### Checklist
- [x] I've updated the documentation if necessary (wasn't necessary)


natecook1000 (commented):
Looks good, thanks for working on this! One nit, then we should be good to merge. 

Have you had a chance to see if this behavior occurs with `fish` completions as well?

CraigSiemens:
> Let's use the stdlib method here instead of Foundation's:

👍 That's been pushed to the branch.

> Have you had a chance to see if this behavior occurs with `fish` completions as well?

From what I can tell, it appears that completions for arguments aren't being generated currently for fish. 

This is returning `nil` if the `ArgumentDefinition.names` is empty.
https://github.com/apple/swift-argument-parser/blob/5be327c/Sources/ArgumentParser/Completions/FishCompletionsGenerator.swift#L62-L64

`names` will always be empty for positional arguments.
https://github.com/apple/swift-argument-parser/blob/5be327c/Sources/ArgumentParser/Parsing/ArgumentDefinition.swift#L109-L114

It'll take me some more time see if there's a reason for the condition in the guard and what the solution would be.

Tests/ArgumentParserExampleTests/MathExampleTests.swift
```diff
@@ -608,23 +608,30 @@ function _swift_math_using_command
 end
 
 complete -c math -n \'_swift_math_using_command \"math add\"\' -l hex-output -s x -d \'Use hexadecimal notation for the result.\'
+complete -c math -n \'_swift_math_using_command \"math add\"\' -l version -d \'Show the version.\'
```
CraigSiemens, line 611: The completions now include the `--version` flag like the bash/zsh commands.

Tests/ArgumentParserExampleTests/MathExampleTests.swift
```diff
@@ -608,23 +608,30 @@ function _swift_math_using_command
...
 complete -c math -n \'_swift_math_using_command \"math stats quantiles\"\' -l custom -r -f -a \'(command math ---completion stats quantiles -- --custom (commandline -opc)[1..-1])\'
+complete -c math -n \'_swift_math_using_command \"math stats quantiles\"\' -l version -d \'Show the version.\'
 complete -c math -n \'_swift_math_using_command \"math stats quantiles\"\' -s h -l help -d \'Show help information.\'
-complete -c math -n \'_swift_math_using_command \"math stats\" \"average stdev quantiles help\"\' -s h -l help -d \'Show help information.\'
-complete -c math -n \'_swift_math_using_command \"math stats\" \"average stdev quantiles help\"\' -f -a \'average\' -d \'Print the average of the values.\'
-complete -c math -n \'_swift_math_using_command \"math stats\" \"average stdev quantiles help\"\' -f -a \'stdev\' -d \'Print the standard deviation of the values.\'
-complete -c math -n \'_swift_math_using_command \"math stats\" \"average stdev quantiles help\"\' -f -a \'quantiles\' -d \'Print the quantiles of the values (TBD).\'
-complete -c math -n \'_swift_math_using_command \"math stats\" \"average stdev quantiles help\"\' -f -a \'help\' -d \'Show subcommand help information.\'
```
CraigSiemens, line 627: The changes seems to have also fixed another bug where it could suggest `help` following subcommands when it isn't a valid subcommand.

CraigSiemens (commented):
I've added a commit that adds the arguments to the fish completions. I tried implementing the logic to follow the bash/zsh implementations to control which arguments are shown/hidden.

natecook1000:
@swift-ci Please test

PR #650, opened 2024-06-27
Updates to support Swift 6 language version

natecook1000:
Some minor NFC changes to resolve strict concurrency checking warnings/errors and the warning about the change in meaning for `#file`.


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Plugins/GenerateManual/GenerateManualPluginError.swift
```diff
@@ -10,7 +10,11 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+#if swift(>=6)
+@preconcurrency import PackagePlugin
```
rauhul, line 14: Hmm it feels like we should file a bug on SwiftPM for this

rauhul (approved):
I dont understand the file -> filePath change but looks right overall

Plugins/GenerateManual/GenerateManualPluginError.swift
natecook1000: I can't reproduce the warnings that led me to add this anymore 😭

natecook1000:
@swift-ci Please test

PR #657, opened 2024-07-26
Add missing @Flag initializer in example

cg-soft:
The documentation as presented doesn't compile. An initializer is required for a boolean flag.

rdar://132579907

### Checklist


rauhul:
I can't remember if omitting the initial value is intended to work, @natecook1000 do you recall?

cg-soft:
@rauhul it definitely does not work. Leaving out the initial value makes it assume it's an Int, and I get this error:

```
error: property type 'Bool' does not match 'wrappedValue' type 'Int'
    var verbose: Bool
        ^~~~~~~
```

natecook1000:
Right! I think you can only omit the default `false` value when you specify an inversion (like a `--foo/--no-foo` pair), which makes specifying one of those values required.

natecook1000:
@swift-ci Please test

PR #660, opened 2024-08-19
Enable library evolution in CMake build

dabrahams:
The use of `@_implementationOnly` causes the compiler to warn of instability at runtime otherwise.


compnerd:
CC: @rauhul

rauhul:
Thoughts on this approach instead: https://github.com/apple/swift-argument-parser/pull/666

I don't really see a need to use this attribute anymore. It's really just doesn't work without library evo enabled anyways.

rauhul:
superseded by: #666

PR #666, opened 2024-09-17
Remove `@_implementationOnly` imports

rauhul:
The `@_implementationOnly` attribute does not work correctly without library evolution enabled. The only possible client with this enabled is Apple, however they have updated to the Swift 6 compiler and can use `internal import` instead of this attribute. Therefore there is no need for this package to use the attribute anymore.

### DIFF
--- a/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
@@ -9,15 +9,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.11)
+#if swift(>=6.0)
 internal import ArgumentParserToolInfo
 internal import class Foundation.JSONEncoder
-#elseif swift(>=5.10)
+#else
 import ArgumentParserToolInfo
 import class Foundation.JSONEncoder
-#else
-@_implementationOnly import ArgumentParserToolInfo
-@_implementationOnly import class Foundation.JSONEncoder
 #endif
 
 internal struct DumpHelpGenerator {
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -11,10 +11,8 @@
 
 #if swift(>=5.11)
 internal import class Foundation.ProcessInfo
-#elseif swift(>=5.10)
-import class Foundation.ProcessInfo
 #else
-@_implementationOnly import class Foundation.ProcessInfo
+import class Foundation.ProcessInfo
 #endif
 
 struct CommandError: Error {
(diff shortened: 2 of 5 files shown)


rauhul:
cc @dabrahams

rauhul:
@swift-ci test

Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
```diff
@@ -9,15 +9,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.11)
+#if swift(>=6.0)
 internal import ArgumentParserToolInfo
 internal import class Foundation.JSONEncoder
-#elseif swift(>=5.10)
+#else
 import ArgumentParserToolInfo
```
compnerd, line 16: I'm slightly concerned about this. The `ArgumentParserToolInfo` library is statically built, and therefore not distributed. However, this is going to serialise the dependency information into the swiftmodule. This might therefore break builds with swift <6.0 on Windows. I do have a pending change (https://github.com/swiftlang/swift/pull/63814) that would help with this situation, but I was running into some trouble with Linux builds ...

rauhul: Hmm implementation only import causes real miscompiles with library evolution so I think it really really needs to go.

natecook1000: @rauhul Where are we seeing this issue show up right now? Can we gate this change so that the attribute is still used on Windows/Linux?

rauhul: Theres a ton of context in this radar: rdar://78129903 (`@_implementationOnly` should either work in non-library-evolution-enabled modules, or be disallowed because we know it doesn't work) (sorry non-apple folks).

The jist of it is that this attribute has never properly worked without library evolution and introduces very hard to locate miscompiles due to different compilation units disagreeing about the layout of types.

@mikeash thoughts?

rauhul: I guess I don't understand the benefits of using this attribute at all for non library evo code

mikeash: I'm not sure myself! Looking at @compnerd's PR, it might break builds by emitting different autolinking info?

rauhul:
@compnerd how frustrating would it be to attempt merging this PR and reverting if it breaks windows?

rauhul:
@swift-ci test

rauhul:
@swift-ci test

natecook1000 (approved):

rauhul:
I'm merging this with the understanding that it may need to be reverted

Issue #668, opened 2024-09-20
help-dump structures should reflect hidden attribute for commands as well as arguments


heckj:
Following up from a [forum's thread on the topic](https://forums.swift.org/t/dump-help-question/74762/2), this bug is to capture the need to extend the dump-help serialization data structure to reflect that a command, as well as an argument, can be hidden from output. 

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
1. Establish an example command and mark it private
2. use the --experimental-dump-help on the resulting executable

### Expected behavior
A "hidden" command should be annotated as such. Additionally, `generate-manual` should exclude content for that command.

### Actual behavior
Hidden commands are exposed, and no data is included in the JSON representation that's made available in the dump-help serialization format.

PR #669, opened 2024-09-20
extend ToolInfoV0 with command visibility

heckj:
- resolves https://github.com/apple/swift-argument-parser/issues/668 by extending the ToolInfoV0 (help dump) struct to include visibility information for commands (already exists for arguments).
- updates test output to verify existing examples extend with the additional key.

### Checklist

### DIFF
--- a/Sources/ArgumentParserToolInfo/ToolInfo.swift
+++ b/Sources/ArgumentParserToolInfo/ToolInfo.swift
@@ -145,7 +145,7 @@ public struct CommandInfoV0: Codable, Hashable {
   public var discussion2: Discussion?
   
   /// Command should appear in help displays.
-  public var shouldDisplay: Bool
+  public var shouldDisplay: Bool = true
 
   /// Optional name of the subcommand invoked when the command is invoked with
   /// no arguments.
@@ -250,7 +250,7 @@ public struct ArgumentInfoV0: Codable, Hashable {
   public var shouldDisplay: Bool
   /// Custom name of argument's section.
   public var sectionTitle: String?
-  
+
   /// Argument can be omitted.
   public var isOptional: Bool
   /// Argument can be specified multiple times.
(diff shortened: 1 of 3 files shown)


Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -51,6 +51,8 @@ public struct CommandInfoV0: Codable, Hashable {
   public var abstract: String?
   /// Extended description of the command's functionality.
   public var discussion: String?
+  /// Command should appear in help displays.
+  public var shouldDisplay: Bool
```
rauhul, line 55: can you add ` = true` as a fall back for older json generated before this option was introduced?

rauhul (changes requested):
small change needed, otherwise looks good

Sources/ArgumentParserToolInfo/ToolInfo.swift
heckj: Updated now - apologies for the delay, didn't have my laptop with me for the past week.

rauhul:
@swift-ci test

natecook1000:
@swift-ci Please test

rauhul:
Sorry for the last minute nit pick, can we add a test case that results in `false` appearing in the json?

heckj:
@rauhul Sure, can definitely do that.

heckj:
I've updated the existing test - command `c` now includes a configuration that "hides" it to illustrate the hiding in effect during dump. It changed the existing output rather than adding a new test case explicitly, so I think this covers what we're after, but if you want the same idea in it's own test case, I can revert the last commit and create a `d` command that illustrates a hidden command in the output.

rauhul (approved):

rauhul:
ship it!

rauhul:
@swift-ci test

PR #684, opened 2024-11-19
build: repair the CMake based build

compnerd:
Add the missing source file to the package to allow building with CMake again.


compnerd:
@swift-ci please test

PR #685, opened 2024-11-20
[cmake] Install libraries in standard directories

Steelskin:
Previously, libraries were installed under `lib/swift/${os}/`. They should be installed in the default library directory for the relevant target system.
In addition, swiftmodules were installed in the older layout format on non-Darwin platforms. This changes to use the standard modern layout format for swiftmodules.

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,9 @@
 cmake_minimum_required(VERSION 3.19)
 
+# Due to a bug in CMake, we need to enable the C language before we can use
+# GNUInstallDirs.
 project(swift-argument-parser
-  LANGUAGES Swift)
+  LANGUAGES C Swift)
 
 option(BUILD_EXAMPLES "Build Example Programs" TRUE)
 option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)
@@ -16,9 +18,6 @@ set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
 include(CTest)
 
-# Due to a bug in CMake, we need to enable the C language before we can use
-# GNUInstallDirs.
-enable_language(C)
 include(GNUInstallDirs)
 include(SwiftSupport)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
 include(CTest)
+include(GNUInstallDirs)
 include(SwiftSupport)
 
 find_package(dispatch CONFIG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
 include(CTest)
+
+# Due to a bug in CMake, we need to enable the C language before we can use
+# GNUInstallDirs.
+enable_language(C)
 include(GNUInstallDirs)
 include(SwiftSupport)
 
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -61,6 +61,32 @@ function(get_swift_host_os result_var_name)
   endif()
 endfunction()
 
+# Returns the module triple in a variable. This is cached for future use.
+#
+# Usage:
+#   get_swift_module_triple(result_var_name)
+#
+# Sets ${result_var_name} with the module triple derived from the Swift compiler.
+# This is cached in the Swift_MODULE_TRIPLE variable.
+function(get_swift_module_triple result_var_name)
+  if(DEFINED Swift_MODULE_TRIPLE)
+    set(${result_var_name} ${Swift_MODULE_TRIPLE} PARENT_SCOPE)
+    return()
+  endif()
+
+  set(module_triple_command "${CMAKE_Swift_COMPILER}" -print-target-info)
+  if(CMAKE_Swift_COMPILER_TARGET)
+    list(APPEND module_triple_command -target ${CMAKE_Swift_COMPILER_TARGET})
+  endif()
+  execute_process(COMMAND ${module_triple_command}
+    OUTPUT_VARIABLE target_info_json)
+
+  string(JSON module_triple GET "${target_info_json}" "target" "moduleTriple")
+  set(Swift_MODULE_TRIPLE "${module_triple}" CACHE STRING "swift module triple used for installed swiftmodule and swiftinterface files")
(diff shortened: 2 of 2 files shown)


Steelskin:
@compnerd Before I get started on other libraries, is this what you had in mind?

cmake/modules/SwiftSupport.cmake
```diff
@@ -85,17 +82,10 @@ function(_install_target module)
...
-      RENAME ${swift_arch}.swiftmodule)
-  else()
-    install(FILES
-      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
-      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
-      DESTINATION lib/${swift}/${swift_os}/${swift_arch})
-  endif()
+  install(FILES $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
+    DESTINATION ${CMAKE_INSTALL_LIBDIR}/${swift}/${swift_os}/${module_name}.swiftmodule
+    RENAME ${swift_arch}.swiftdoc)
```
compnerd, line 87: I believe that these should be named as per the triple for the host not just the architecture component.

Steelskin: I am now deriving the triple from the compiler.

compnerd:
@swift-ci please test

compnerd:
This is going to require a cross-repository test to ensure that the toolchain builds don't break.

cmake/modules/SwiftSupport.cmake
```diff
@@ -61,6 +61,32 @@ function(get_swift_host_os result_var_name)
...
 endfunction()
 
+# Returns the module triple in a variable. This is cached for future use.
+#
+# Usage:
+#   get_swift_module_triple(result_var_name)
+#
+# Sets ${result_var_name} with the module triple derived from the Swift compiler.
+# This is cached in the Swift_MODULE_TRIPLE variable.
+function(get_swift_module_triple result_var_name)
```
etcwilde, line 71: Rather than making it a function, why don't we set it up early on in the build so that the value is just available for use?

Steelskin: Done, thanks!

cmake/modules/SwiftSupport.cmake
```diff
@@ -61,6 +61,30 @@ function(get_swift_host_os result_var_name)
   endif()
 endfunction()
 
+# Attempt to get the module triple from the Swift compiler.
```
etcwilde, line 64: Lets wrap this in a check to see if the variable is set so that we don't re-run the compiler each time we configure:

```cmake
if(NOT Swift_MODULE_TRIPLE)
...
endif()
```

This is what it looks like for the new stdlib build: https://github.com/swiftlang/swift/blob/25252cd6836f512ff820ff995ae8d92ab47076d1/Runtimes/Core/cmake/modules/PlatformInfo.cmake#L7-L20

cmake/modules/SwiftSupport.cmake
```diff
@@ -61,6 +61,30 @@ function(get_swift_host_os result_var_name)
...
+endif()
+execute_process(COMMAND ${module_triple_command}
+  OUTPUT_VARIABLE target_info_json)
+string(JSON module_triple GET "${target_info_json}" "target" "moduleTriple")
+
+# If we failed to get the module triple, fall back to the host arch.
+if(NOT module_triple)
+  get_swift_host_arch(host_arch)
+  set(module_triple ${host_arch})
+  message(WARNING
```
etcwilde, line 77: Perhaps `SEND_ERROR` here? I think I'd prefer to fail the build if we can't get the appropriate filename.
If someone really wants to set the triple by hand, with the `if(NOT Swift_MODULE_TRIPLE)` check above, they can pass one on the command line easily enough.

Steelskin: I used `FATAL_ERROR` here. Let me know if you'd prefer `SEND_ERROR`. I don't think there's any point in continuing here otherwise.

cmake/modules/SwiftSupport.cmake
Steelskin: Done.

cmake/modules/SwiftSupport.cmake
```diff
@@ -71,31 +93,20 @@ function(_install_target module)
     set(swift swift)
   endif()
 
-  install(TARGETS ${module}
-    ARCHIVE DESTINATION lib/${swift}/${swift_os}
-    LIBRARY DESTINATION lib/${swift}/${swift_os}
-    RUNTIME DESTINATION bin)
+  install(TARGETS ${module})
```
compnerd, line 96: This wont install to the correct location on all platforms. The DLLs need to go to `bin` on Windows and the import libraries and static libraries into `lib`.

Steelskin: Based on the [documentation](https://cmake.org/cmake/help/latest/command/install.html#targets) I believe this is already the default (scroll down for the table of default values).
Windows DLLs are `RUNTIME` artifacts, which get installed under `bin`.
Windows import and static libraries are `ARCHIVE` artifacts, which get installed under `lib`.

Steelskin: I ran this branch on our internal build here: https://github.com/thebrowsercompany/swift-build/actions/runs/13001454471
It succeeded, including with packaging.

compnerd (approved):

CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
 include(CTest)
+include(GNUInstallDirs)
```
etcwilde, line 18: `GNUInstallDirs` requires either C or C++ be enabled to work correctly.

```
CMake Warning (dev) at ...share/cmake-3.30/Modules/GNUInstallDirs.cmake:253 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:18 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

etcwilde (commented):
Feel free to leave a bug report on CMake, but at the moment, `GNUINstallDirs` doesn't work for projects that only use `Swift` due to how CMake detects the correct library install dirs.

CMakeLists.txt
Steelskin: Thank you, I enabled C.

Steelskin:
> Feel free to leave a bug report on CMake, but at the moment, `GNUINstallDirs` doesn't work for projects that only use `Swift` due to how CMake detects the correct library install dirs.

I filed https://gitlab.kitware.com/cmake/cmake/-/issues/26649 for this

compnerd:
@swift-ci please test

etcwilde (approved):

Steelskin:
I did a small update to enable C as part of the project definition. I was hitting issues with the C compiler otherwise.

compnerd:
@swift-ci please test

PR #690, opened 2025-01-22
Add singleton to indicate which shell version is requesting completion candidates

### LINKED ISSUE #689: Indicate version of shell requesting completions
A `SAP_SHELL_VERSION` environment variable should be exported by each generated shell completion script.

Its value will be the version of the shell requesting completions. e.g., bash could be `3.2.57`, `5.2.37`, etc.; zsh could be `5.3`, 5.9`, etc.; fish could be `3.7.1`, etc.

 It will only be available at completion time, not at script-generation time, so it will only be useful for `shellCommand` or `custom` `CompletionKind`s. `shellCommand` is written to the script at generation time, but the commands in its `String` argument can use `${SAP_SHELL_VERSION}`, while the function argument for `custom` is run during completion, so `${SAP_SHELL_VERSION}` will be used to set the following while while a custom completion function is running (otherwise, it will be set to `nil`):

```swift
public struct CompletionShell: … {
  …
  public internal(set) static var requestingVersion: String?
  …
}
```

This will allow custom completions to provide better completions (older shell versions might not support newer completion formats / shell commands or builtins), and will also allow them to know when they will receive different data due to shell bugs / limitations limited to certain shell versions (e.g., various issues in the fish `commandline` command will be fixed in the forthcoming fish 4.0.0, so output will be different for 4.0.0+ & prior versions).

rgoldberg:
Add singleton to indicate which shell version is requesting completion candidates.

Create a `String?` singleton named `CompletionShell.requestingVersion` that indicates which shell version is requesting completion candidates.

It will be set to the correct value while a Swift custom completion function is executing to offer completions for a word from a command line (e.g., while `customCompletion` from `@Option(completion: .custom(customCompletion))` executes). Otherwise, it will be set to `nil`.

The requesting shell version is communicated to the Swift app via an environment variable named `SAP_SHELL_VERSION`, which is exported by each of the generated completion scripts.

Improve some nearby DocC.

Resolve #689

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -57,7 +57,9 @@ struct BashCompletionsGenerator {
     // that other command functions don't need.
     if isRootCommand {
       result += """
-        export \(CompletionShell.environmentVariableName)=bash
+        export \(CompletionShell.shellEnvironmentVariableName)=bash
+        \(CompletionShell.shellVersionEnvironmentVariableName)="$(IFS='.'; printf %s "${BASH_VERSINFO[*]}")"
+        export \(CompletionShell.shellVersionEnvironmentVariableName)
         cur="${COMP_WORDS[COMP_CWORD]}"
         prev="${COMP_WORDS[COMP_CWORD-1]}"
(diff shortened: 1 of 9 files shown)


natecook1000:
@swift-ci Please test

PR #691, opened 2025-01-22
Allow `@Option`s with tuple types

natecook1000:
This adds `@Option` initializers that use variadic generics to work with tuples of any size.

**From the symbol documentation:**
> For example, you could use this property wrapper to capture the three dimensions for a package:
> 
> ```swift
> struct Package: ParsableArguments {
>     @Option
>     var dimensions: (Double, Double, Double)
> }
> ```
> 
> A user would then specify the three dimensions after the `--dimensions` key:
> 
> ```
> $ package --dimensions 5 4 12
> ```

### Still to do
- [ ] Decide on naming/etc on the help screen
- [ ] Include in the serialized help info
- [ ] Include in completion scripts
- [ ] Document in guides

### Checklist


Sources/ArgumentParser/Parsable Properties/Option.swift
```diff
@@ -861,3 +861,145 @@ extension Option {
...
+  }
+}
+
+// MARK: Variadic tuple support
+
+fileprivate struct InitFailure: Error {
+  var index: Int
+}
+
+/// Returns a variadic tuple, with values generated by zipping the provided
```
rauhul, line 975: variadic generic code is damn near impossible to read...

PR #692, opened 2025-02-04
Add tests parsing dump help output

rauhul:
Adds a new test target called ArgumentParserToolInfoTests which contains example json output from dump-help. The test code attempts to parse all json files in the 'Examples' subdirectory. This will help us catch cases where ArgumentParser is unable to parse json generated by older versions.

We still need to add a test that asserts old ArgumentParser versions can parse the output of new versions.


rauhul:
@swift-ci test

rauhul:
@swift-ci test

PR #698, opened 2025-02-06
Extend snapshot testing to completion scripts

rauhul:
As the title suggests this commit moves the completion script tests to
use the snapshot testing recently introduced.


rauhul:
@swift-ci test

PR #700, opened 2025-02-06
Move additional tests to snapshots

rauhul:
Moves some examples test to use snapshot files instead of inline multiline strings.


rauhul:
@swift-ci test

PR #705, opened 2025-02-06
Fix concurrency warnings

rauhul:
Moves static read-only vars behind a mutex so they are currency safe. Users of the api don't need to worry about this because the mutex is hidden behind an accessor.


rauhul:
@swift-ci test

Issue #704, opened 2025-02-06
documentation doesn't build cleanly on main


rauhul:
```
| Converting documentation...
| warning: '--level' is deprecated. Use '--coverage-summary-level' instead.
| warning: '--index' is deprecated. Use '--emit-lmdb-index' indead.
| note: The majority of content should be under level-3 headers under the "Overview" section
| Articles/CustomizingCommandHelp.md:5:1: Overview section starts here
|   --> Articles/CustomizingCommandHelp.md:62:1-62:33
| 60 | ```
| 61 |
| 62 + ## Modifying the Help Flag Names
|    | ╰─suggestion: Change the heading to a level-3 heading
| 63 |
| 64 | Users can see the help screen for a command by passing either the `-h` or the `--help` flag, by default. If you need to use one of those flags for another purpose, you can provide alternative names when configuring a root command.
|
| note: The majority of content should be under level-3 headers under the "Overview" section
| Articles/CustomizingCommandHelp.md:5:1: Overview section starts here
|    --> Articles/CustomizingCommandHelp.md:126:1-126:19
| 124 | ```
| 125 |
| 126 + ## Hiding Commands
|     | ╰─suggestion: Change the heading to a level-3 heading
| 127 |
| 128 | You may not want to show every one of your command as part of your command-line interface. To render a command invisible (but still usable), pass `shouldDisplay: false` to the ``CommandConfiguration`` initializer.
|
| note: The majority of content should be under level-3 headers under the "Overview" section
| Articles/CustomizingCommandHelp.md:5:1: Overview section starts here
|    --> Articles/CustomizingCommandHelp.md:130:1-130:41
| 128 | You may not want to show every one of your command as part of your command-line interface. To render a command invisible (but still usable), pass `shouldDisplay: false` to the ``CommandConfiguration`` initializer.
| 129 |
| 130 + ## Generating Help Text Programmatically
|     | ╰─suggestion: Change the heading to a level-3 heading
| 131 |
| 132 | The help screen is automatically shown to users when they call your command with the help flag. You can generate the same text from within your program by calling the `helpMessage()` method.
|
| note: The majority of content should be under level-3 headers under the "Overview" section
| Articles/CustomizingHelp.md:5:1: Overview section starts here
|   --> Articles/CustomizingHelp.md:38:1-38:34
| 36 | ```
| 37 |
| 38 + ## Customizing Help for Arguments
|    | ╰─suggestion: Change the heading to a level-3 heading
| 39 |
[...]

PR #708, opened 2025-02-07
Enable Documentation check

rauhul:
Enables the documentation check from the swiftlang soundness workflow and resolves errors/notes emitted by the check.

Fixes: #704


rauhul:
@swift-ci test

Issue #679, opened 2024-10-31
Some `CompletionKind`s are not properly implemented for some shells


rgoldberg:
Some `CompletionKind`s are not properly implemented for some shells.

e.g., `file` & `directory` are not implemented for bash argument completion. `file` is improperly implemented for fish (at least for fish 3.7.1).

I will work on a PR to fix these issues. If I find additional problems, I'll add them to this issue.

The `CompletionKind` / `CompletionKind.Kind` documentation is also sometimes wrong or confusing, so I will update that along with the associated code.

My PR will fix the problems mentioned in #566, but won't implement its proposed solution.

It looks like #376 has already been fixed, but it if somehow resurfaces under, I'll fix the issue in my PR.

**ArgumentParser version:**
`main` branch

**Swift version:**
```text
swift-driver version: 1.62.15 Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
Target: x86_64-apple-macosx12.0
```

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
Generate completion scripts using the aforementioned `CompletionKind`s.

Try to use them.

### Expected behavior
Completion works.

### Actual behavior
Some completion either does nothing, or outputs an error.

rgoldberg:
I am partway through a PR for this, with many improvements for all 3 shells.

What is the oldest version of each shell with which the respective script must maintain compatibility? Here are some guesses:

- bash 3.2.57 (included in macOS 10.13)
- zsh 5.3 (included in macOS 10.13)
- fish 3.7.1 (latest fish for almost a year; fish is not included with any macOS, so we can expect users to know about upgrading it)

I know that SAP can be used on OSes other than macOS (sorry to be Mac-centric, but I don't know shell versions distributed with / popular for other OSes).

Supporting bash 3.2.57 might be annoying because it doesn't support associative arrays. Being able to require some bash 4.x or (better yet) 5.x might simplify things.

natecook1000:
If we _can_ make things work with the default-installed Bash version, I think we should. I know it's a bit of a pain, but it would feel like a mismatch (and probably generate more issues) to require Bash v4 or v5. The other two shell versions both seem reasonable!

rgoldberg:
@natecook1000 thanks. Will do.

A few more questions:

[...]

PR #727, opened 2025-02-13
Improve zsh completion script generation

### LINKED ISSUE #726: Improve zsh completion script generation
Improve zsh completion script generation.

There are many issues with script generation. Some of them are documented in #679 & its comments.

This issue is for a first batch of zsh completion script fixes.

I will submit a PR. Each commit should note the fix(es) that it contains.

If you want documentation of the fixes directly in the GitHub issue and/or PR without having to look at the commit messages, I can include info from the commit messages here and/or in the PR description.

If you want the PR for this split up, parts removed or changed, etc., please let me know, and I'll make the requisite changes.

After this PR is merged, I will sequentially create an issue & submit a PR each for initial bash & fish fixes, then one or more PRs for cross-shell changes, then I will resume working on more fixes.

rgoldberg:
Improve zsh completion script generation.

There are many issues with script generation. Some of them are documented in #679 & its comments.

This PR is for a first batch of zsh completion script fixes.

Each commit should note the fix(es) that it contains.

If you want documentation of the fixes directly in the GitHub issue #726 and/or in this PR without having to look at the commit messages, I can include info from the commit messages here and/or in the issue description.

If you want the this PR split up, parts removed or changed, etc., please let me know, and I'll make the requisite changes.

After this PR is merged, I will sequentially create an issue & submit a PR each for initial bash & fish fixes, then one or more PRs for cross-shell changes, then I will resume working on more fixes.

Resolve #726

### Checklist


natecook1000:
@swift-ci

natecook1000:
@swift-ci Please test

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
```diff
@@ -98,6 +98,14 @@ public struct CompletionShell: RawRepresentable, Hashable, CaseIterable {
...
   /// The environment variable is set in generated completion scripts.
   static let shellVersionEnvironmentVariableName = "SAP_SHELL_VERSION"
+
+  public func format(completions: [String]) -> String {
+    var completions = completions
+    if self == .zsh {
+      completions.append("END_MARKER")
+    }
+    return completions.joined(separator: "\n")
+  }
```
natecook1000, line 108: It looks like the updated zsh completion script will always drop this sentinel, but I'm thinking people with completion scripts already installed might start getting `END_MARKER` in their suggested completions if they update the executable without updating their script. Does that sound right? Is there a different value for this sentinel that wouldn't show up as a completion if it weren't stripped out?

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
```diff
@@ -164,11 +172,39 @@ extension ParsableCommand {
   }
 }
 
+extension [ParsableCommand.Type] {
+  // Include default 'help' subcommand in nonempty subcommand list iff no help subcommand already exists.
+  mutating func addHelpSubcommandIffMissing() {
```
natecook1000, line 177: I know what "iff" means, but past experience tells me we'll get multiple PRs "correcting" this spelling. I think the meaning of just plain "if" is fine here instead (in the comment and the method name).

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
natecook1000: Also, my reading of this design is that `format(completions:)` is called automatically by `CommandParser`, it isn't something that CLI tool authors need to think about. Is it `public` for testing purposes? If so, let's at least underscore the API.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
```diff
@@ -164,11 +172,39 @@ extension ParsableCommand {
...
+    if !isEmpty && allSatisfy({ $0._commandName != "help" }) {
+      append(HelpCommand.self)
+    }
+  }
+}
+
 extension Sequence where Element == ParsableCommand.Type {
```
natecook1000, line 184: These can probably all be in the array extension…

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
```diff
@@ -9,217 +9,236 @@
...
 //===----------------------------------------------------------------------===//
 
-struct ZshCompletionsGenerator {
+extension [ParsableCommand.Type] {
   /// Generates a Zsh completion script for the given command.
-  static func generateCompletionScript(_ type: ParsableCommand.Type) -> String {
-    let initialFunctionName = [type].completionFunctionName()
+  var zshCompletionScript: String {
+    """
+    #compdef \(first?._commandName ?? "")
```
natecook1000, line 16: It would be an internal error if `self` were empty here, so I'd rather fully fail than end up with an unusable script. Can you bind `first` before rendering the string, and then trap otherwise?

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
```diff
@@ -9,217 +9,236 @@
...
-  {
-    guard let type = commands.last else { return "" }
-    let functionName = commands.completionFunctionName()
-    let isRootCommand = commands.count == 1
+  private var completionFunctions: String {
+    guard let type = last else { return "" }
+    let functionName = completionFunctionName()
+    let isRootCommand = count == 1
+
+    let argumentSpecsAndSetupScripts = argumentsForHelp(visibility: .default)
```
natecook1000, line 42: I know this matches what we're doing now, but perhaps we should include `.hidden` arguments in the completion script, just not `.private`. Let's discuss later…

natecook1000 (commented):
Thanks so much for this awesome work, @rgoldberg! From testing it looks flawless, just have a few minor notes/questions below.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: > It looks like the updated zsh completion script will always drop this sentinel, but I'm thinking people with completion scripts already installed might start getting `END_MARKER` in their suggested completions if they update the executable without updating their script. Does that sound right?

Correct. I just assumed that install routines / people will regenerate the completion scripts.

> Is there a different value for this sentinel that wouldn't show up as a completion if it weren't stripped out?

I will check if a single space (or any other value) will be ignored. The value `END_MARKER` is completely unimportant, as the new scripts just lop off the last element, regardless of its value. I just thought that `END_MARKER` would be easier for someone else to understand if they were reading the code / debugging things; I assumed that the minor inconvenience of having `END_MARKER` show up only for `custom` completions, and only when scripts hadn't been regenerated, would be worth it in the long run, but I understand your concerns about migration hassles.

> Also, my reading of this design is that format(completions:) is called automatically by CommandParser, it isn't something that CLI tool authors need to think about. Is it public for testing purposes? If so, let's at least underscore the API.

Correct. It was only for testing purposes. Being a Swift newbie, I forgot about `@testable import`. I'll make it internal instead of public, but use `@testable import` for testing.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: Will replace. Or should I just include a comment like the following?

```swift
// "Iff" is not a misspelling; it means "if & only if"
```

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
rgoldberg: I don't know the exact difference between `.hidden` & `.private`, but, making an educated guess, that sounds like it could make sense.

Since I don't know SAP very well, I tried to limit my changes to fixing shell issues & fixing SAP interactions that I know are broken, but it's a great idea to revisit existing assumptions throughout the completion script generation process.

I assume we should leave this particular improvement for a subsequent PR so it can be mulled over & discussed, but I'm happy to include it here if you so instruct me.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rauhul: lets stick with "if"

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
rgoldberg: Is a `fatalError()` in a `guard` block OK?

If so, is `first!` OK as long as I silence lint warnings with a `// swift-format-ignore: NeverForceUnwrap` comment?

I now notice that's used 24 times elsewhere under `Sources`.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: Do you want me to move all into:

```swift
extension [ParsableCommand.Type]
```

Or:

```swift
extension Sequence where Element == ParsableCommand.Type
```

The following also built on Swift 5.7.1 & passed all tests (I'm stuck on Swift 5.7.1 & Intel until the M4 Mac Studios are released…); I assume that the prior code is an older syntax, and that the following code is a newer equivalent syntax, but please let me know if I'm wrong (if correct, should a codebase-wide PR be done to migrate to the cleaner syntax?):

```swift
extension Sequence<ParsableCommand.Type>
```

The first is most succinct (& presumably handles all current use cases), but I assume the latter 2 are broader.

I'd probably vote for the third version, as long as it is correct…

rgoldberg:
@natecook1000 Thanks for the feedback. Do you want me to update things via new commits, or by fixing existing commits & force pushing? I assume the former, since it seems like SAP squashes PRs, but I wanted to check with you first.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: FYI: turns out that neither `public` nor `@testable import` are required; one of the two was required for one way in which I attempted to set environment variables for testing, but the way that I ended up using doesn't seem to require either of them. I just forgot to remove `public`…

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: Pushed fix.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: Pushed fix for unnecessary public function. Still must investigate replacing `END_MARKER`.

rgoldberg: Pushed fix to use ` ` instead of `END_MARKER` as the zsh pseudo-completion.

rgoldberg:
@natecook1000 I've submitted fixes as new commits for the conversations that I've resolved. Waiting for your instructions on the open conversations. Thanks for all the feedback & help.

rgoldberg:
Rebased on the new `main` as there is a simple conflict that needed to be resolved.

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
natecook1000: This is definitely for later, just something I noticed when reviewing… `.hidden` args aren't included in the default `--help` output, but are visible with `--help-hidden`. `.private` args are never included.

Sources/ArgumentParser/Completions/ZshCompletionsGenerator.swift
natecook1000: `first!` is OK with both the format-ignore comment and an explanatory comment (e.g. `// Precondition: self is guaranteed to be non-empty`).

Switching from `?? ""` to the force unwrap sounds good where it makes sense – if you need it multiple times in a function I'd rather you bind the unwrapped value to a local variable. Let's keep the format-ignore comments as local as possible.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
natecook1000: Re my original suggestion, I was just noticing that this was probably an unnecessarily broad extension, as I think these are all just used on arrays. It doesn't seem necessary to have the similar functionality broken across different extensions, so if it still builds with them all together, then that's a better strategy.

Those two syntaxes are equivalent, but the second has both Swift and runtime version requirements. The feature to support it (primary associated types) was added via [this proposal in Swift 5.7](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0358-primary-associated-types-in-stdlib.md), so we can use it without conditions now. (And a separate PR for that conversion would be great!)

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
natecook1000: Thanks so much for finding that solution!

natecook1000:
Feel free to handle commits/updating however's most comfortable for you – we'll squash on merge, so the in-process commits aren't terribly important.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
rgoldberg: I actually need `completionFunctionName()` in the `Sequence` extension for changes that I've made to generated bash completion scripts in commits that I haven't yet submitted, so I will leave the two members where they are.

rgoldberg:
@natecook1000 I've finished all the fixes. Please note that I cannot move the members from the `extension Sequence` to the `extension […]` because in some not-yet submitted commits for bash completion script generation, I need one of the members in `extension Sequence`, so I should leave both in case we need the other there in the future, too.

I've also rebased on `main`.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testColorDoccReference().md
```diff
@@ -32,3 +32,8 @@ color help [<subcommands>...]
 ```
 
 **subcommands:**
+
+
+
+
+
```
natecook1000, line 39: Likewise in these snapshots…

natecook1000 (commented):
Code changes look great! Just a little bit of cleanup, then we're good to merge 🎉

Tests/ArgumentParserExampleTests/RepeatExampleTests.swift
```diff
@@ -61,6 +64,8 @@ final class RepeatExampleTests: XCTestCase {
         --count <count>         The number of times to repeat 'phrase'.
         --include-counter       Include a counter with each repetition.
         -h, --help              Show help information.
+
+
```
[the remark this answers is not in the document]
rgoldberg: The tests break if the newlines are removed.

The newlines are generated by the code; since I removed whitespace trimming, they must be included in the expected output.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testColorDoccReference().md
rgoldberg: The tests break if the newlines are removed.

The newlines are generated by the code; since I removed whitespace trimming, they must be included in the expected output.

rgoldberg:
@natecook1000 Just replied to your 2 comments.

Tests/ArgumentParserExampleTests/RepeatExampleTests.swift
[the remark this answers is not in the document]
natecook1000: Ah, in `AssertExecuteCommand` and friends - got it. I've verified that this doesn't mark a change in the produced help output, so we're good to go. Thanks for clarifying!

natecook1000:
@swift-ci Please test
@swift-ci Please test

PR #735, opened 2025-02-15
Improve bash completion script generation

### LINKED ISSUE #734: Improve bash completion script generation
Improve bash completion script generation.

There are many issues with script generation. Some of them are documented in #679 & its comments.

This issue is for a first batch of bash completion script fixes.

I will submit a PR. Each commit should note the fix(es) that it contains.

After this PR is merged, I will create an issue & submit a PR for initial fish fixes, then one or more PRs for cross-shell changes, then I will resume working on more fixes.

rgoldberg:
Improve bash completion script generation.

There are many issues with script generation. Some of them are documented in #679 & its comments.

This issue is for a first batch of bash completion script fixes.

Each commit should note the fix(es) that it contains.

Note that I use "iff" in some Swift & bash comments, but not in code or in DocC. If you want "iff" removed from comments, please let me know.

Will rebase if `main` is updated (especially if #726 is merged).

Resolve #734

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -365,7 +365,8 @@ extension [ParsableCommand.Type] {
     case .custom:
       // Generate a call back into the command to retrieve a completions list
       return """
-        COMPREPLY+=($(compgen -W "$("${COMP_WORDS[0]}" \(arg.customCompletionCall(self)) "${COMP_WORDS[@]}")" -- "${cur}"))
+        \(addCompletionsFunctionName) -W\
+         "$("${COMP_WORDS[0]}" \(arg.customCompletionCall(self)) "${COMP_WORDS[@]}")"
 
         """
     }
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -335,9 +335,9 @@ extension [ParsableCommand.Type] {
         """
 
     case .file(let extensions):
-      let exts = extensions.map { $0.shellEscapeForSingleQuotedString() }
-        .flatMap { [$0, $0.uppercased()] }
-        .joined(separator: "|")
+      let exts =
+        extensions
+        .map { $0.shellEscapeForSingleQuotedString() }.joined(separator: "|")
       return """
         \(addCompletionsFunctionName) -o plusdirs -fX '!*.@(\(exts))'
 
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -71,16 +71,12 @@ extension [ParsableCommand.Type] {
                   -*)
                       # ${word} is a flag or an option
                       # If ${word} is an option, mark that the next word to be parsed is an option value
-                      # TODO: handle joined-value options (-o=file.ext), stacked flags (-aBc), legacy long (-long), combos
-                      # TODO: if multi-valued options can exist, support them
                       local option
                       for option in "${options[@]}"; do
                           [[ "${word}" = "${option}" ]] && is_parsing_option_value=true && break
                       done
 
                       # Remove ${word} from ${flags} or ${options} so it isn't offered again
-                      # TODO: handle repeatable flags & options
-                      # TODO: remove equivalent options (-h/--help) & exclusive options (--yes/--no)
                       local not_found=true
                       local -i index
                       for index in "${!flags[@]}"; do
@@ -107,7 +103,6 @@ extension [ParsableCommand.Type] {
               fi
 
               # ${word} is neither a flag, nor an option, nor an option value
-              # TODO: can SAP be configured to require options before positionals?
               if [[ "${positional_number}" -lt "${positional_count}" ]]; then
                   # ${word} is a positional
                   ((positional_number++))
@@ -126,7 +121,6 @@ extension [ParsableCommand.Type] {
 
           unparsed_words=("${unparsed_words[@]}")
 
-          # TODO: offer flags & options after all positionals iff they're allowed after positionals
           if\\
               ! "${was_flag_option_terminator_seen}"\\
               && ! "${is_parsing_option_value}"\\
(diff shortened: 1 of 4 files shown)


Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
```diff
@@ -9,257 +9,397 @@
...
 
-          """
-          .indentingEachLine(by: 8)
-      }
-      result += "    esac\n"
+            # Offer positional completions
+            case "${positional_number}" in
+        \(positionalArguments
+          .compactMap { arg in
+            position += 1
```
natecook1000, line 256: I generally try to have a transform like this be a pure function, without the side effect of mutating a captured variable (i.e. `position` here). If you zip `positionalArguments` with an open range, you can iterate over both the position and the arguments at the same time:

```suggestion
        \(zip(1..., positionalArguments)
          .compactMap { position, arg in
```

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
```diff
@@ -9,257 +9,397 @@
...
-      // Subcommands have their own case statement that delegates out to
-      // the subcommand completion functions.
-      result += "    case ${COMP_WORDS[\(dollarOne)]} in\n"
-      for subcommand in subcommands {
-        result += """
-          (\(subcommand._commandName))
-              \(functionName)_\(subcommand._commandName) \(subcommandArgument)
-              return
-              ;;
+    if !positionalArguments.allSatisfy({ bashValueCompletion($0).isEmpty }) {
```
natecook1000, line 248: What would you think about _always_ doing the `compactMap` on lines 254+, and then, when that's not empty, adding the result plus the wrapper text (instead of this check)?

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
```diff
@@ -9,257 +9,397 @@
...
+  }
 
-        return """
-          \(arg.bashCompletionWords().joined(separator: "|")))
-          \(arg.bashValueCompletion(commands).indentingEachLine(by: 4))
-              return
-          ;;
-          """
+  /// Returns flags for the last command of the given array.
+  private var flags: [String] {
```
natecook1000, line 312: Thoughts on naming this something like `flagCompletions` instead? Reading the source above it looked like it would be the same type as `positionalArguments` (an array of `ArgumentDefinition`). Same thoughts for `options`, below.

Sources/ArgumentParser/Completions/CompletionsGenerator.swift
```diff
@@ -164,6 +164,24 @@ extension ParsableCommand {
...
+      return []
+    }
+    return ArgumentSet(command, visibility: .default, parent: nil)
+      .filter(\.isPositional)
+  }
+
+  /// Include default 'help' subcommand in nonempty subcommand list if & only if
+  /// no help subcommand already exists.
+  mutating func addHelpSubcommandIfMissing() {
+    if !isEmpty && allSatisfy({ $0._commandName != "help" }) {
```
natecook1000: Let's rewrite to flip the polarity:
```suggestion
    if !isEmpty && !contains(where: { $0._commandName == "help" }) {
```

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
```diff
@@ -9,257 +9,397 @@
...
+                      if [[ "${word_index}" -ne "${unparsed_word_indices[${#unparsed_word_indices[@]} - 1]}" ]]; then
+                          was_flag_option_terminator_seen=true
+                      fi
+                      continue
+                      ;;
+                  -*)
+                      # ${word} is a flag or an option
+                      # If ${word} is an option, mark that the next word to be parsed is an option value
+                      # TODO: handle joined-value options (-o=file.ext), stacked flags (-aBc), legacy long (-long), combos
+                      # TODO: if multi-valued options can exist, support them
```
natecook1000, line 75: Let's move these implementation notes out of the generated script – if there's a place for them in the Swift source, that's okay; otherwise let's just capture in issues.

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
```diff
@@ -9,257 +9,397 @@
...
+                          done
+                      fi
+                      unset "unparsed_words[${word_index}]"
+                      continue
+                      ;;
+                  esac
+              fi
+
+              # ${word} is neither a flag, nor an option, nor an option value
+              # TODO: can SAP be configured to require options before positionals?
```
natecook1000, line 110: This isn't a feature currently – options/flags can come anywhere in the list of arguments.

rgoldberg:
@natecook1000 Just rebased on the new `main` with the zsh PR merged.

FYI: there are tons of TODOs in the bash completion code, as I haven't had a chance to get to all of the issues that I've seen. There are other issues that I've identified (in bash, fish, or zsh) for which I haven't included TODOs, too. I basically only noted issues in new code that I wrote, rather than in existing code; I completely rewrote the main completion logic for bash, so there is a lot of new code…

natecook1000 (commented):
@rgoldberg This is great, again! 👏🏻 👏🏻 👏🏻 I really appreciate you starting from first principles for a rewrite of this size – the generated script is working without issues for me locally, so I just have these few implementation notes.

rgoldberg:
@natecook1000 Thanks for the feedback. Will try to get to it all tomorrow, then also submit a fish PR.

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
rgoldberg: I've never heard of `zip` before. Thanks for the tip. Being a Swift newbie has its drawbacks…

Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
rgoldberg: Makes sense. I put the notes in situ since I used them as a guide for what to work on. Once the code is in GitHub, however, I can use code links in issues, which I couldn't do before.

rgoldberg:
@natecook1000 Subimtted fixes for all the comments.

rgoldberg:
@natecook1000 Rebased on new `main`.

natecook1000:
@swift-ci Please test

natecook1000 (approved):
🚢

PR #745, opened 2025-02-23
Make NameSpecification and its element ExpressibleByStringLiteral

natecook1000:
This simplifies declaring alternate names by allowing just a string representation of an option or flag name. The string is parsed to split on spaces, and then validates that each name has either a one or two dash prefix and is ASCII only.

Existing style:

```swift
@Flag(name: [.customLong("hex-output"), .customShort("x")])
var hexadecimalOutput = false
```

New style:

```swift
@Flag(name: "--hex-output -x")
var hexadecimalOutput = false
```

### Checklist


natecook1000:
@rgoldberg No, we don't have any of that validation in place. There's a certain degree of self-enforcement in place, in that things will just not work properly if you do things like defining an option with a space in its name. But we should add some of that, and can be more stringent in this case, since it's a new behavior.

dickoff:
Can one combine styles here? For instance this to achieve both `--hex-output` and `-x`?
```swift
@Flag(name: ["-x", .long])
var hexOutput = false
```

rauhul:
should this be a variadic or an array instead? e.g.
```
@Option(name: "--foo", "-f", more-args: ...)
@Option(name: ["--foo", "-f"], more-args: ...)
```

natecook1000:
@rauhul As designed right now, you can specify a group as either a space-delimited string or with string elements in an array. To @dickoff's question, you can also mix and match, since both the `NameSpecification` type and its elements are `ExpressibleByStringLiteral`.

Issue #721, opened 2025-02-12
generate-docc-reference: use term lists for arguments


rauhul:
term lists in docc provide a better UX for argument names

PR #753, opened 2025-04-03
[GenerateDocCReference] Improve filtering and categorization of arguments in documentation with sectionTitle

Chamepp:
### Summary
The recent modifications improve the readability and structure of documents generated using the **GenerateDocC** plugin. This update addresses https://github.com/apple/swift-argument-parser/issues/722 by enhancing how arguments are categorized within sections using the sectionTitle property. 

Now, arguments—including required arguments, optionals, and flags—are grouped under their respective **sectionTitles** based on their type. Since **sectionTitle** is an optional value, some arguments may not have one explicitly provided. To handle this, we introduced logic that assigns a default section when no custom **sectionTitle** is specified. These default sections are:  

- **Arguments** – for required arguments  
- **Options** – for optional values  
- **Flags** – for boolean flags  

Each argument is placed under its corresponding section based on its type. This ensures a well-structured and organized documentation format while allowing users to define their own custom **sectionTitles** when needed.

### Changes

- Listing arguments under custom sectionTitles
  - If a custom sectionTitle is provided, arguments are correctly grouped under it.
- Default section titles for missing values
  - When no custom sectionTitle is given, arguments are automatically grouped based on their type:
      - Positional arguments → "Arguments"
      - Options → "Options"
      - Flags → "Flags"

- Better argument categorization and organization
   - Arguments are now properly grouped and listed under designated sections.
   - Ensures all arguments appear under an appropriate category, whether a custom or default sectionTitle is provided.

- Fixed missing dash in argument output
   - The previous method incorrectly displayed arguments like -option instead of --option.
   - Now, options are displayed correctly with a double dash (--option) by wrapping the argument in (‘’)

- Updated unit tests and snapshot recordings

   - Re-ran and updated snapshot tests for:
      - Documentation snapshots
      - Dump help snapshots

   - Since section titles now have default values, output consistency has been verified.

- Subcommands now have their own category
   - Ensured subcommands are listed under a dedicated section to maintain consistency across all argument categories.

(Closes https://github.com/apple/swift-argument-parser/issues/722)"

### Checklist

### DIFF
--- a/Sources/ArgumentParserToolInfo/ToolInfo.swift
+++ b/Sources/ArgumentParserToolInfo/ToolInfo.swift
@@ -218,7 +218,23 @@ public struct ArgumentInfoV0: Codable, Hashable {
     self.kind = kind
 
     self.shouldDisplay = shouldDisplay
-    self.sectionTitle = sectionTitle
+
+    // The section title helps categorize arguments
+    // for better organization in documentation.
+
+    // The switch statement checks the kind of argument
+    // and assigns a default section title related to the argument type
+    // if the provided one is nil.
+    switch kind {
+    case .positional:
+      self.sectionTitle = sectionTitle ?? "Arguments"
+
+    case .option:
+      self.sectionTitle = sectionTitle ?? "Optionals"
+
+    case .flag:
+      self.sectionTitle = sectionTitle ?? "Flags"
+    }
 
     self.isOptional = isOptional
     self.isRepeating = isRepeating
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollDoccReference().md
@@ -6,32 +6,38 @@
 roll [--times=<n>] [--sides=<m>] [--seed=<seed>] [--verbose] [--help]
 ```
 
-**--times=\<n\>:**
+## Flags
+
+**`--verbose`**
+
+*Show all roll results.*
+
+
+**`--help`**
+
+*Show help information.*
+
+
+## Optionals
+
+**`--times=\<n\>`**
 
 *Rolls the dice <n> times.*
 
 
-**--sides=\<m\>:**
+**`--sides=\<m\>`**
 
 *Rolls an <m>-sided dice.*
 
 Use this option to override the default value of a six-sided die.
 
 
-**--seed=\<seed\>:**
+**`--seed=\<seed\>`**
 
 *A seed to use for repeatable random generation.*
 
 
-**--verbose:**
-
-*Show all roll results.*
-
-
-**--help:**
-
-*Show help information.*
-
+## Subcommands
 
 ## roll.help
 
@@ -41,9 +47,6 @@ Show subcommand help information.
 roll help [<subcommands>...] 
 ```
 
-**subcommands:**
+## Arguments
 
-
-
-
-
+**`subcommands`**
--- a/Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
+++ b/Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
@@ -59,25 +59,70 @@ extension CommandInfoV0 {
     }
 
     if let args = self.arguments {
-      for arg in args {
-        guard arg.shouldDisplay else {
-          continue
+      // Group arguments by sectionTitle
+      // This is done to organize the arguments into categories (e.g., optionals, flags)
+      let groupedArgs = Dictionary(
+        grouping: args.filter {
+          $0.shouldDisplay
         }
+      ) { $0.sectionTitle ?? "General" }
 
-        result += "**\(arg.identity()):**\n\n"
-        if let abstract = arg.abstract {
-          result += "*\(abstract)*\n\n"
+      // Iterate through the grouped arguments, sorted by section title
+      // Sorting ensures that the sections appear in a clear, predictable order in the final documentation.
+      // The sections are listed alphabetically based on their title, enhancing readability.
+      for (section, arguments) in groupedArgs.sorted(by: { $0.key < $1.key }) {
+        // Add section title as a Markdown header to the result
+        // Adding section titles as Markdown headers helps separate sections in the documentation,
+        // making it easier for users to navigate and understand the structure of the arguments.
+        result += "## \(section)\n\n"
+
+        // Iterate through each argument in the section
+        // Iterating through each argument allows us to add details for each argument under its respective section.
+        for arg in arguments {
+          // Add the argument identity in bold Markdown format
+          // The argument identity is emphasized in bold to make it stand out in the documentation.
+          result += "**`\(arg.identity())`**\n\n"
+
+          // If the argument has an abstract, add it in italic Markdown format
+          // Including the abstract provides a brief description of the argument's purpose.
+          // The italic formatting helps to distinguish the abstract from other parts of the documentation.
+          if let abstract = arg.abstract {
+            result += "*\(abstract)*\n\n"
+          }
+
+          // If the argument has a discussion, add it directly
+          // If a discussion is available, it is added to provide further explanation on how the argument works.
+          // This additional context is helpful for users to understand the full usage of the argument.
+          if let discussion = arg.discussion {
+            result += discussion + "\n\n"
+          }
+
+          // Add an empty line for separation between arguments
+          // The empty line improves the visual structure of the documentation,
+          // making it easier to read and separating each argument for clarity.
+          result += "\n"
         }
-        if let discussion = arg.discussion {
-          result += discussion + "\n\n"
-        }
-        result += "\n"
       }
     }
 
-    for subcommand in self.subcommands ?? [] {
-      result += subcommand.toMarkdown(path + [self.commandName]) + "\n\n"
+    if let subcommands = self.subcommands, !subcommands.isEmpty {
+      // Add a separate section for subcommands.
+      // Subcommands are treated differently from regular arguments, so they deserve their own section in the documentation.
+      // This helps to visually distinguish subcommands from regular arguments, enhancing readability.
+      result += "## Subcommands\n\n"
+
+      // Iterate through each subcommand and convert it to Markdown format
+      // Each subcommand is processed and added to the documentation under the "Subcommands" section.
+      // This ensures that users can easily identify and understand the subcommands available for the command.
+      for subcommand in subcommands {
+        result += subcommand.toMarkdown(path + [self.commandName])
+      }
     }
+
+    // Trim any unnecessary trailing newline that could have been added inadvertently
+    // By trimming, we prevent extra lines at the end of the generated document
+    // which is important for snapshot comparison.
+    result = result.trimmingCharacters(in: .newlines)
 
     return result
   }
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testColorDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testColorDoccReference().md
@@ -6,22 +6,28 @@
 color --fav=<fav> [--second=<second>] [--help]
 ```
 
-**--fav=\<fav\>:**
+## Flags
+
+**`--help`**
+
+*Show help information.*
+
+
+## Optionals
+
+**`--fav=\<fav\>`**
 
 *Your favorite color.*
 
 
-**--second=\<second\>:**
+**`--second=\<second\>`**
 
 *Your second favorite color.*
 
 This is optional.
 
 
-**--help:**
-
-*Show help information.*
-
+## Subcommands
 
 ## color.help
 
@@ -31,9 +37,6 @@ Show subcommand help information.
 color help [<subcommands>...] 
 ```
 
-**subcommands:**
+## Arguments
 
-
-
-
-
+**`subcommands`**
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testCountLinesDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testCountLinesDoccReference().md
@@ -6,25 +6,33 @@
 count-lines [<input-file>] [--prefix=<prefix>] [--verbose] [--help]
 ```
 
-**input-file:**
+## Arguments
+
+**`input-file`**
 
 *A file to count lines in. If omitted, counts the lines of stdin.*
 
 
-**--prefix=\<prefix\>:**
+## Flags
 
-*Only count lines with this prefix.*
-
-
-**--verbose:**
+**`--verbose`**
 
 *Include extra information in the output.*
 
 
-**--help:**
+**`--help`**
 
 *Show help information.*
 
+
+## Optionals
+
+**`--prefix=\<prefix\>`**
+
+*Only count lines with this prefix.*
+
+
+## Subcommands
 
 ## count-lines.help
 
@@ -34,9 +42,6 @@ Show subcommand help information.
 count-lines help [<subcommands>...] 
 ```
 
-**subcommands:**
+## Arguments
 
-
-
-
-
+**`subcommands`**
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testMathDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testMathDoccReference().md
@@ -8,15 +8,19 @@ A utility for performing maths.
 math [--version] [--help]
 ```
 
-**--version:**
+## Flags
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
 *Show help information.*
 
+
+## Subcommands
 
 ## math.add
 
@@ -26,29 +30,28 @@ Print the sum of the values.
 math add [--hex-output] [<values>...] [--version] [--help]
 ```
 
-**--hex-output:**
+## Arguments
 
-*Use hexadecimal notation for the result.*
-
-
-**values:**
+**`values`**
 
 *A group of integers to operate on.*
 
 
-**--version:**
+## Flags
+
+**`--hex-output`**
+
+*Use hexadecimal notation for the result.*
+
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
-*Show help information.*
-
-
-
-
-## math.multiply
+*Show help information.*## math.multiply
 
 Print the product of the values.
 
@@ -56,29 +59,28 @@ Print the product of the values.
 math multiply [--hex-output] [<values>...] [--version] [--help]
 ```
 
-**--hex-output:**
+## Arguments
 
-*Use hexadecimal notation for the result.*
-
-
-**values:**
+**`values`**
 
 *A group of integers to operate on.*
 
 
-**--version:**
+## Flags
+
+**`--hex-output`**
+
+*Use hexadecimal notation for the result.*
+
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
-*Show help information.*
-
-
-
-
-## math.stats
+*Show help information.*## math.stats
 
 Calculate descriptive statistics.
 
@@ -86,15 +88,19 @@ Calculate descriptive statistics.
 math stats [--version] [--help]
 ```
 
-**--version:**
+## Flags
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
 *Show help information.*
 
+
+## Subcommands
 
 ### math.stats.average
 
@@ -104,29 +110,30 @@ Print the average of the values.
 math stats average [--kind=<kind>] [<values>...] [--version] [--help]
 ```
 
-**--kind=\<kind\>:**
+## Arguments
 
-*The kind of average to provide.*
-
-
-**values:**
+**`values`**
 
 *A group of floating-point values to operate on.*
 
 
-**--version:**
+## Flags
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
 *Show help information.*
 
 
+## Optionals
 
+**`--kind=\<kind\>`**
 
-### math.stats.stdev
+*The kind of average to provide.*### math.stats.stdev
 
 Print the standard deviation of the values.
 
@@ -134,24 +141,23 @@ Print the standard deviation of the values.
 math stats stdev [<values>...] [--version] [--help]
 ```
 
-**values:**
+## Arguments
+
+**`values`**
 
 *A group of floating-point values to operate on.*
 
 
-**--version:**
+## Flags
+
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
-*Show help information.*
-
-
-
-
-### math.stats.quantiles
+*Show help information.*### math.stats.quantiles
 
 Print the quantiles of the values (TBD).
 
@@ -159,44 +165,43 @@ Print the quantiles of the values (TBD).
 math stats quantiles [<one-of-four>] [<custom-arg>] [<values>...]     [--file=<file>] [--directory=<directory>] [--shell=<shell>] [--custom=<custom>] [--version] [--help]
 ```
 
-**one-of-four:**
+## Arguments
+
+**`one-of-four`**
 
 
-**custom-arg:**
+**`custom-arg`**
 
 
-**values:**
+**`values`**
 
 *A group of floating-point values to operate on.*
 
 
-**--file=\<file\>:**
+## Flags
 
-
-**--directory=\<directory\>:**
-
-
-**--shell=\<shell\>:**
-
-
-**--custom=\<custom\>:**
-
-
-**--version:**
+**`--version`**
 
 *Show the version.*
 
 
-**--help:**
+**`--help`**
 
 *Show help information.*
 
 
+## Optionals
+
+**`--file=\<file\>`**
 
 
+**`--directory=\<directory\>`**
 
 
-## math.help
+**`--shell=\<shell\>`**
+
+
+**`--custom=\<custom\>`**## math.help
 
 Show subcommand help information.
 
@@ -204,9 +209,6 @@ Show subcommand help information.
 math help [<subcommands>...] 
 ```
 
-**subcommands:**
+## Arguments
 
-
-
-
-
+**`subcommands`**
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRepeatDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRepeatDoccReference().md
@@ -6,25 +6,33 @@
 repeat [--count=<count>] [--include-counter] <phrase> [--help]
 ```
 
-**--count=\<count\>:**
+## Arguments
 
-*The number of times to repeat 'phrase'.*
-
-
-**--include-counter:**
-
-*Include a counter with each repetition.*
-
-
-**phrase:**
+**`phrase`**
 
 *The phrase to repeat.*
 
 
-**--help:**
+## Flags
+
+**`--include-counter`**
+
+*Include a counter with each repetition.*
+
+
+**`--help`**
 
 *Show help information.*
 
+
+## Optionals
+
+**`--count=\<count\>`**
+
+*The number of times to repeat 'phrase'.*
+
+
+## Subcommands
 
 ## repeat.help
 
@@ -34,9 +42,6 @@ Show subcommand help information.
 repeat help [<subcommands>...] 
 ```
 
-**subcommands:**
+## Arguments
 
-
-
-
-
+**`subcommands`**
--- a/Tests/ArgumentParserUnitTests/Snapshots/testADumpHelp().json
+++ b/Tests/ArgumentParserUnitTests/Snapshots/testADumpHelp().json
@@ -29,6 +29,7 @@
           "kind" : "long",
           "name" : "enumerated-option"
         },
+        "sectionTitle" : "Optionals",
         "shouldDisplay" : true,
         "valueName" : "enumerated-option"
       },
@@ -61,6 +62,7 @@
           "kind" : "long",
           "name" : "enumerated-option-with-default-value"
         },
+        "sectionTitle" : "Optionals",
         "shouldDisplay" : true,
         "valueName" : "enumerated-option-with-default-value"
       },
@@ -78,6 +80,7 @@
           "kind" : "long",
           "name" : "no-help-option"
         },
+        "sectionTitle" : "Optionals",
         "shouldDisplay" : true,
         "valueName" : "no-help-option"
       },
@@ -96,6 +99,7 @@
           "kind" : "long",
           "name" : "int-option"
         },
+        "sectionTitle" : "Optionals",
         "shouldDisplay" : true,
         "valueName" : "int-option"
       },
@@ -115,6 +119,7 @@
           "kind" : "long",
           "name" : "int-option-with-default-value"
         },
+        "sectionTitle" : "Optionals",
         "shouldDisplay" : true,
         "valueName" : "int-option-with-default-value"
       },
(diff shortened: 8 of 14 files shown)


Sources/ArgumentParserToolInfo/ToolInfo.swift
```diff
@@ -218,7 +218,23 @@ public struct ArgumentInfoV0: Codable, Hashable {
     self.kind = kind
 
     self.shouldDisplay = shouldDisplay
-    self.sectionTitle = sectionTitle
+
```
rauhul, line 221: This should not be part of ToolInfo. Instead the consumer of tool info can choose how to interpret the missing title. (the consumers of toolinfo are/will be: generate–docc-reference, generate-manual, generate-shell-completion, help-generation, usage-generation)

rauhul (changes requested):

Sources/ArgumentParserToolInfo/ToolInfo.swift
Chamepp: Thank you for your feedback dear Rauhul. I handled the logic in the generate-docc command.

Tools/generate-docc-reference/GenerateDoccReference.swift
```diff
@@ -126,4 +153,19 @@ struct GenerateDoccReference: ParsableCommand {
       print(page)
     }
   }
+
+  func assignSectionTitle(to argument: ArgumentInfoV0) throws -> String {
```
natecook1000, line 157: I don't see any errors that this function is throwing, which would eliminate the need for the new `argumentAssignmentFailed` error above.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -59,26 +59,71 @@ extension CommandInfoV0 {
...
-        if let abstract = arg.abstract {
-          result += "*\(abstract)*\n\n"
-        }
-        if let discussion = arg.discussion {
-          result += discussion + "\n\n"
+      ) { $0.sectionTitle ?? "General" }
+
+      // Iterate through the grouped arguments, sorted by section title
+      // Sorting ensures that the sections appear in a clear, predictable order in the final documentation.
+      // The sections are listed alphabetically based on their title, enhancing readability.
```
natecook1000, line 72: Is this alphabetical sorting the same as used when printing help for the command-line tool?

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -59,26 +59,71 @@ extension CommandInfoV0 {
...
+      // Subcommands are treated differently from regular arguments, so they deserve their own section in the documentation.
+      // This helps to visually distinguish subcommands from regular arguments, enhancing readability.
+      result += "## Subcommands\n\n"
+
+      // Iterate through each subcommand and convert it to Markdown format
+      // Each subcommand is processed and added to the documentation under the "Subcommands" section.
+      // This ensures that users can easily identify and understand the subcommands available for the command.
+      for subcommand in subcommands {
+        result += subcommand.toMarkdown(path + [self.commandName])
+      }
```
natecook1000, line 119: I don't think this change to how subcommands are listed is an improvement, since it ends up hard to see the hierarchy between commands (see the `math` docs). Could you keep this PR focused on grouping the args/options/flags by section?

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -59,26 +59,71 @@ extension CommandInfoV0 {
...
+          $0.shouldDisplay
         }
-
-        result += "**\(arg.identity()):**\n\n"
-        if let abstract = arg.abstract {
-          result += "*\(abstract)*\n\n"
-        }
-        if let discussion = arg.discussion {
-          result += discussion + "\n\n"
+      ) { $0.sectionTitle ?? "General" }
```
natecook1000, line 68: Why is the `"General"` fallback needed here? It looks like the section titles might only be added for the root-level command (see the subcommands in the `math` docs for example).

natecook1000 (commented):
Thanks for working on this, @Chamepp! I think it would be great to focus this PR on just categorizing the arguments/options/flags, rather than also changing the subcommand display.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
Chamepp: Sure. The reason I changed this alongside the argument grouping was because after grouping the arguments, the subcommands started to feel a bit confusing. Since I was already organizing arguments into sections, I thought it might make sense to give subcommands their own section as well. And since both changes were aligned with the same purpose which is grouping by section, I thought it would be reasonable to include this small change in the same PR. I’ve now reverted the changes related to subcommands. 

Let me know if you'd like a shorter or more formal version too.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
Chamepp: The reason behind this fallback was that in case the property doesn’t have a custom sectionTitle and doesn’t fall into arguments default categories, it will be placed  under the General category. I thought this will be great to handle this edge case and we can spot problems when the General section appears alongside throwing an error. I think I have developed a bad habit of over engineering. From now on I will focus on what's necessary and essential.

Chamepp: Yes I checked the Math documentation output. This wasn’t an issue when I was assigning the default value in the ToolInfo. Since subcommands also uses the `toMarkDown()` function from the `GenerateDocC` extension, we should handle the default value of the section in the CommandInfoV0 extension. By doing this, every single command that uses the `toMarkdown()` can benefit from this feature. I have implemented this and will commit my new changes soon.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
Chamepp: The way the documentation generation sorts this is: Arguments, Flags, Options.
If documentation generation sorts this without alphabetical sorting the placement of default section titles will not be consistent. We will get section titles sorted randomly every time.

When we are printing help for the repeat command, which includes all three types of values(arguments, flags and options), the output is sorted in this way: Arguments, Options. Then Options section include options listed first and then flags next.

I think with the addition of custom sectionTitles this is going to be a good implementation since it makes the documentation easier to work with. But as you mentioned I think it's important that we match the help output, and in some way we do.

I will wait for your feedback on this. If you think this is a good choice I will commit my changes. If not I will find another solution to make sure we list Arguments, Options and Flags in a consistent way that follows the help output pattern. Also I can implement alphabetical sorting for custom section titles.

Chamepp: I assume you might be busy so I went ahead and committed my changes and will request for review. Let me know if you think we should change the sorting logic.

Another problem that arises if we don’t maintain a consistent section creation via sorting is that the snapshot and the generated document will always differ since the document generation produces random results. This will cause the tests to fail every time.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -59,26 +59,61 @@ extension CommandInfoV0 {
...
+      ) { self.assignSectionTitle(to: $0) }
+
+      // Iterate through the grouped arguments, sorted by section title
+      // Sorting ensures that the sections appear in a clear, predictable order in the final documentation.
+      // The sections are listed alphabetically based on their title, enhancing readability.
+      for (section, arguments) in groupedArgs.sorted(by: { $0.key < $1.key }) {
+        // Add section title as a Markdown header to the result
+        // Adding section titles as Markdown headers helps separate sections in the documentation,
+        // making it easier for users to navigate and understand the structure of the arguments.
+        result += "## \(section)\n\n"
```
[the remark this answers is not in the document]
Chamepp: That's a great point, thank you. I will adjust it shortly.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -59,26 +59,61 @@ extension CommandInfoV0 {
...
+
+        // Iterate through each argument in the section
+        // Iterating through each argument allows us to add details for each argument under its respective section.
+        for arg in arguments {
+          // Add the argument identity in bold Markdown format
+          // The argument identity is emphasized in bold to make it stand out in the documentation.
+          result += "**`\(arg.identity())`**\n\n"
```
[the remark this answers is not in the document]
Chamepp: Thank you for taking the time to review this. I agree that using term lists is beneficial. I had actually noticed a separate issue related to this #721 and was planning to open a new pull request for it. I saw that you already addressed it, appreciate that.

natecook1000:
@Chamepp It looks like the merge lost some of the output, at least according to the test fixtures – can you take a look?

Chamepp:
@natecook1000 Sure. Solved merge conflicts again and force pushed.

natecook1000 (commented):
@Chamepp, apologies for the delayed response! I think if we step back, this change would make the most sense if it aligned the structure of the generated DocC and man pages with what you see in the help screen for a command. Specifically:

- "Arguments" is the right category to come first
- "Options" should be the next category, including both `@Flag` and `@Option` properties – that distinction is internal to the library, it's not an external one
- Arguments/options/flags that are in a titled option group should be pulled out into a separate section with that title – see this test file for examples (AFAICT it doesn't look like this feature is covered in the tests for DocC or man page generation) https://github.com/apple/swift-argument-parser/blob/main/Tests/ArgumentParserUnitTests/HelpGenerationTests%2BGroupName.swift
- We should have some kind of default help text for the `command help <subcommand>` pseudo-argument, since there's no place to customize that.

Thanks again for your contribution here!

Chamepp:
@natecook1000 Yes, I believe it's a great idea to keep things consistent in design. Keeping the categorization structure unified across DocC, the man page, and the help screen will provide clarity for the user through that consistency. I think it's definitely more user friendly.

I'll revise the PR with the changes you suggested and get back to you.

Chamepp:
Dear @natecook1000,

I’ve updated my solutions according to the requested modifications and added a default help message for the subcommands argument. Personally, I see the `subcommands` argument default help as more of a temporary fix. I believe it would be better to add the default help message directly to the help command’s subcommands argument itself.

```swift
@Argument(help: "The subcommand(s) you want help for.")
var subcommands: [String] = []
```

Since this PR mainly focuses on documentation generation, I’m happy to make that change in a separate PR. However, if you think it’s fine to include it here, I can go ahead and make the adjustments.

Please let me know what you prefer and what you think of this solution.

Chamepp:
I also noticed that dear @heckj is working on an [improvement](https://github.com/apple/swift-argument-parser/pull/791) that might cause a conflict with this PR. I’ll wait for their PR to be merged first, and then I’ll resolve any conflicts and revisit this.

In the meantime I’ll think of some scenarios where adding the default help message during documentation generation could cause issues. It’s also possible I’m missing something. Perhaps there are challenges with adding the help message directly to the subcommands argument which I don't see yet.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -78,25 +78,64 @@ extension CommandInfoV0 {
...
+          // Add the argument identity in bold Markdown format
+          // The argument identity is emphasized in bold to make it stand out in the documentation.
+
+
+          // If the argument has an abstract, add it in italic Markdown format
+          // Including the abstract provides a brief description of the argument's purpose.
+          // The italic formatting helps to distinguish the abstract from other parts of the documentation.
+          switch markdownStyle {
+          case .docc:
+            result += "- term **\(arg.identity()):**\n\n"
```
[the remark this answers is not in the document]
d-ronnqvist: FYI: these trailing new lines aren't necessarily and can make it harder to associate the argument with its description. For example, consider this argument from one of the tests

```md
- term **--seed=\<seed\>**:

*A seed to use for repeatable random generation.*
``` 

That could also have been written as either;

```md
- term **--seed=\<seed\>**:
  *A seed to use for repeatable random generation.*
```

or even 

```md
- term **--seed=\<seed\>**: *A seed to use for repeatable random generation.*
```

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -78,25 +78,64 @@ extension CommandInfoV0 {
...
+          case .github:
+            result += "**\(arg.identity()):**\n\n"
+          }
+
+          if let abstract = arg.abstract {
+            result += "*\(abstract)*\n\n"
```
d-ronnqvist, line 116: I have two comments about this:
1. With the blank line from above, can you verify that this text does in fact associate with the term-definition element correctly? I would think that the blank line and the lack of leading indentation would cause this to parse as a separate paragraph.
2. If basically all the text in the documentation is emphasized, that emphasis looses its importance and only result in a page where all the text is slanted.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollDoccReference().md
```diff
@@ -7,29 +7,31 @@ roll [--times=<n>] [--sides=<m>] [--seed=<seed>] [--verbose]
...
 
 
-- term **--seed=\<seed\>**:
+- term **--seed=\<seed\>:**
 
 *A seed to use for repeatable random generation.*
 
 
-- term **--verbose**:
+- term **--verbose:**
```
d-ronnqvist, line 29: Question: does this render as expected? I know that term-definition lists split on the colon but I don't know of the top of my head what happens if the colon is inside a formatted element.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRepeatMarkdownReference().md
```diff
@@ -6,6 +6,15 @@
 repeat [--count=<count>] [--include-counter] <phrase> [--help]
 ```
 
+### Arguments
+
+**phrase:**
```
d-ronnqvist, line 11: Question: should this argument also be listed as a term-definition element?

Tools/generate-docc-reference/GenerateDoccReference.swift
```diff
@@ -110,13 +109,14 @@ struct GenerateDoccReference: ParsableCommand {
       throw GenerateDoccReferenceError.unableToParseToolOutput(error: error)
     }
 
-    let toolInfo: ToolInfoV0
+    var toolInfo: ToolInfoV0
```
d-ronnqvist, line 112: Just curious: what's the reason for making this property mutable? At a glance I couldn't see any mutations in the code.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -138,8 +182,23 @@ extension CommandInfoV0 {
     }
     return multilineString
   }
+
+  // Assign a default section title based on the arguments types
+  func assignSectionTitle(to argument: ArgumentInfoV0) -> String {
```
d-ronnqvist, line 187: I find this method and documentation to be a bit strangely named and designed. "assign" is a verb phrase that I read as "this method will _assign_ a section title to the argument". The documentation seems to imply the same interpretation of this function. However, this function doesn't perform an action (_assign_ a value), rather it returns value without mutating the argument. 

One way to clarify this would be to rename the method to represent the returned value and update the documentation to reflect what the method does:
```suggestion
  // Returns the default section title for the provided the argument.
  func sectionTitle(for argument: ArgumentInfoV0) -> String {
```

However, because the method isn't using any state outside of the passed argument, it might make for a clearer call site to move this to be a private extension on `ArgumentInfoV0` that adds a `defaultSectionTitle` computed property.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollMarkdownReference().md
```diff
@@ -6,6 +6,8 @@
 roll [--times=<n>] [--sides=<m>] [--seed=<seed>] [--verbose] [--help]
 ```
 
+### Options
```
d-ronnqvist, line 9: Question: how are arguments, flags, and options organized in the command-line help text? Are they organized into categories there as well? I feel that the documentation's organization should follow the command-line help text's organization.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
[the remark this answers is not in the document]
d-ronnqvist: Question: is there an example from the tests where there's a flag/option has more then one name (for example `-h` and `--help`)? How does this code format such a flag/option?

d-ronnqvist (commented):
My main question is how these changes makes the rendered documentation look. If you could include a few screenshots that show the rendered documentation with these changes, then that would make it easier to review.

Chamepp:
> My main question is how these changes makes the rendered documentation look. If you could include a few screenshots that show the rendered documentation with these changes, then that would make it easier to review.

@d-ronnqvist Here are screenshots demonstrating how the generated document categorizes the math add subcommand, matching its help output:

Screenshot 2025-09-19 at 12 03 16 PM
Screenshot 2025-09-19 at 12 05 21 PM

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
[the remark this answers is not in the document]
Chamepp: > I personally am not convinced that bold/strong text is the preferred style to represent a command line flag in a term-definition list. In my DocC proof of concept ([here](https://github.com/swiftlang/swift-docc/pull/837)) I used "code voice" / monospaced text for the flags/options/arguments and I think that looks better on the rendered page.

Since our goal is to align with DocC’s style for documentation generation, I agree that using the `“code voice” / monospaced style` you suggested is a better way to represent arguments, options, and flags in the generated documentation. I’ll add this to my task list, and once I’ve compiled all the new required modifications, I’ll share the list with you for approval before implementing the changes.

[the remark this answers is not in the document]
Chamepp: > FYI: these trailing new lines aren't necessarily and can make it harder to associate the argument with its description. For example, consider this argument from one of the tests
> 
> ```
> - term **--seed=\<seed\>**:
> 
> *A seed to use for repeatable random generation.*
> ```
> 
> That could also have been written as either;
> 
> ```
> - term **--seed=\<seed\>**:
>   *A seed to use for repeatable random generation.*
> ```
> 
> or even
> 
> ```
> - term **--seed=\<seed\>**: *A seed to use for repeatable random generation.*
> ```

That’s a great point. Since we are aligning with the DocC generation style, I agree that this is a better implementation.

> ```
> - term **--seed=\<seed\>**:
>   *A seed to use for repeatable random generation.*
> ```

I think following this style is the better choice. I’ll also add it to my task list.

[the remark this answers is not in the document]
Chamepp: > Question: is there an example from the tests where there's a flag/option has more then one name (for example `-h` and `--help`)? How does this code format such a flag/option?

The `--hex-output` option for the `math add` subcommand is a good example of the issue you mentioned, which I’ve highlighted in the screenshot earlier. For now, we are only displaying the long name in the terms list. My goal was to focus on the categorization and filtering of arguments for now and then plan to address the finer details in subsequent PR's to avoid complication and confusion.

In order to solve this we have to focus on this area and support short names for the arguments in the terms list alongside the long name setting. Then we could follow the DocC style you mentioned [earlier](https://github.com/swiftlang/swift-docc/pull/837) in the PR you shared, to present them in a way that matches the expected output.

The area for the development:
```swift
    // TODO: default values, short, etc.

    var inner: String
    switch self.kind {
    case .positional:
      inner = "<\(names.joined(separator: "|"))>"
    case .option:
      inner = "--\(names.joined(separator: "|"))=<\(self.valueName ?? "")>"
    case .flag:
      inner = "--\(names.joined(separator: "|"))"
    }
```

Final output:
```swift
- term `-x`, `--hex-output`:
  *A seed to use for repeatable random generation.*
```

If you think this can be handled in this PR, I can add it to my task list. If not, we can address it in the next PR. I’ll wait for your approval before adding it as a task for now.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
Chamepp: 1. I checked this and you’re right. After doing some research I realized that the extra newline isn’t necessary. Markdown treats a blank line as a paragraph break, so the *abstract* may be interpreted as a separate paragraph rather than part of the term definition list. In other words, DocC could render it disconnected from the term instead of as its definition. I’ll add this to my tasks and refine the solution. That was a great point, thank you for the feedback.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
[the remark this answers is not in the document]
Chamepp: I can also go into more detail on how the solution would be implemented to give you a clearer picture, if you think that would be a good idea. We can better decide whether it makes sense to include it here or in a separate PR.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
```diff
@@ -78,25 +78,64 @@ extension CommandInfoV0 {
...
+          if let abstract = arg.abstract {
+            result += "*\(abstract)*\n\n"
+          }
+
+          // Inject a default help description for the the `help` command positional argument 'subcommands'.
+          // The 'subcommands' argument is missing an abstract in ArgumentParser.
+          // This results in incomplete help documentation, which may confuse users.
+          // Since it's a positional argument named 'subcommands', it can be easily confused with actual subcommands.
+          if arg.identity() == "subcommands" && arg.abstract == nil {
+            result += "*The subcommand(s) you want help for.*\n\n"
```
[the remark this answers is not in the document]
Chamepp: As you know, every command and subcommand has a default subcommand called help. This help command takes an argument called subcommands, which can cause confusion during development because it might be mistaken for the actual subcommands section.

The subcommands argument always appears at the end of each generated documentation under the help section.

> We should have some kind of default help text for the help <subcommand> pseudo-argument, since there’s no place to customize that.

Previously, there was no default help text for this argument, so we added it manually. However, from your feedback I realized that the abstraction I introduced differs from the `math add --help` → help `subcommands` argument. I’ll add this to my tasks and refine the solution accordingly.

Screenshot 2025-09-19 at 3 43 14 PM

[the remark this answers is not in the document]
d-ronnqvist: > As you know, every command and subcommand has a default subcommand called help. 

If this is the same content for every command, is there much value in repeating a (more or less) identical automatic page for every CLI?

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRepeatMarkdownReference().md
Chamepp: Well since its a positional argument and we should match the help output structure for filtering and categorization, yes it should be listed as a term definition.

Tools/generate-docc-reference/GenerateDoccReference.swift
Chamepp: You’re right. I don’t actually mutate toolInfo, so it makes sense to make it a let. I’ll add it to my tasks.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
Chamepp: I agree. I think this naming is clearer and better represents what the function does. I’ll also move it into a private extension on `ArgumentInfoV0` as a computed property to make the call site cleaner. I will add this to my tasks list.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollMarkdownReference().md
Chamepp: Good point, I actually covered this earlier with the screenshot and explanation. The organization in the documentation directly follows the command line help text’s organization, so they stay consistent.

Chamepp:
@d-ronnqvist Here’s the task list I put together based on your earlier suggestions. Thank you again for your feedback, it really helped shape a much more mature solution.

- [x] Use `code voice / monospaced text` for arguments, options, and flags instead of bold. This better matches DocC documentation style.
- [x] Move the colon outside of formatted elements in term definition lists. The colon is a **syntactic separator** for DocC, and wrapping it prevents proper parsing.
- [x] Remove unnecessary italics from argument abstracts. Keeping abstracts in plain text (and reserving italics only for meaningful emphasis) makes the rendered output clearer, easier to scan, and more consistent with DocC style.
- [x] Refine argument rendering to remove unnecessary newlines between terms and abstracts. Blank lines are treated as paragraph breaks in Markdown, which can cause abstracts to be rendered disconnected from their corresponding term.
- [x] Change `toolInfo` from var to let since it is never mutated, improving clarity and signaling immutability.
- [x] Rename `assignSectionTitle(to:)` to `sectionTitle(for:)` and update documentation.
   - [x] move it to a private `ArgumentInfoV0` extension as a computed property. This better expresses that the function returns a value without mutating state.
- [x] Update the injected default abstract for the subcommands argument in the help subcommand so it matches exactly the text shown in the CLI --help output.

The only task I’m waiting on is supporting short names alongside long names in the documentation generation, which I mentioned earlier. I’ll wait for your approval before working on it to see if you consider it appropriate for this PR. Otherwise, I can handle it in a subsequent PR. In the meantime, I’m continuing work on the tasks above.

- [ ] ~~Supporting short name alongside the long name of the arguments in the documentation generation~~(This task will be handled in the subsequent PR's)

Chamepp:
@d-ronnqvist One thing I’ve realized is that I don’t yet have a deep understanding of the DocC documentation generation process or how the parser works at its core. Since you’ve spent a lot of time building it, you clearly have a more mature perspective and expertise than I do. You’re spotting connections and areas for improvement that I’ve been missing.

Because of that, I thought it might be great if you could highlight some issues or areas in DocC that I could contribute to. By working on those, I’d be able to build enough knowledge to design more mature solutions in the future.

So far, my approach has been to build on Swift Argument Parser and check for DocC compatibility along the way. But I believe first developing a stronger understanding of DocC would make the process much faster and lead to better solutions. If you consider this a good approach, I’d love the chance to contribute to DocC as well and keep learning through that.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollMarkdownReference().md
d-ronnqvist: Where can I find that screenshot? I don't see any in the PR description.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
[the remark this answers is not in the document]
Chamepp: I never thought about it that way, that’s a smart approach. When it comes to decision making for design, I personally focus on essentialism too. If it’s not a 100 percent yes, then it’s a no, in order to increase focus and avoid clutter and noise. Doing this keeps the design simple. I also agree that in terms of software maintenance, avoiding duplication can be helpful. But since users already expect the `--help` flag and help subcommand to be present for every command, and because we want the command help output to be consistent, I believe the user’s cognitive load is lower by avoiding surprises. In this case, consistency outweighs the small cost of duplication.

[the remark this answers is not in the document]
Chamepp: I believe keeping help in the same place consistently reduces mental effort. Even if the text or page is repetitive, it makes the documentation easier to navigate. Users don’t have to guess where help is documented, it’s always there, in the same way. When I think about the end result, skipping or centralizing it could lead to an uneven experience, where some commands have help documented and others don’t.

Tools/generate-docc-reference/GenerateDoccReference.swift
Chamepp: I double checked this. I ran a clean build with:
```swift
swift package clean
swift build -Xswiftc -warnings-as-errors
```
so any warnings introduced by my changes would fail the build.

Screenshot 2025-09-25 at 11 35 54 AM
<br />
<br />

The build did fail, but it was due to an existing deprecation warning, `customDeprecated` in `DumpHelpGenerator.swift`, not something introduced in this PR.

To confirm, I ran:
```swift
git blame Sources/ArgumentParser/Usage/DumpHelpGenerator.swift -L 213,217
```

and saw that this code was added in commit 42585ad6 (2025-05-07) by Ross Goldberg, so it predates my changes.

So to clarify, my PR doesn’t introduce new warnings.

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollMarkdownReference().md
Chamepp: @d-ronnqvist I have already included screenshots of both the generated documentation output and the CLI help output for the `math add` subcommand [here](https://github.com/apple/swift-argument-parser/pull/753#issuecomment-3311210196). For better understanding and in order to make the comparison clearer, I’ll also share an example of the roll command you mentioned below.

Our solution currently follows and matches the command help output structure and categorization. In the help output, the organization is consistent. **Positional** arguments appear first under the `Arguments` section, followed by the `Options` section, where **optional** arguments are listed first and **flags** are listed afterward.

Since the roll command doesn’t have any positional arguments, there is no `Arguments` category for it in the generated help text. The `math add` help output is a better reference because it includes both positional arguments and options, and shows how they are listed in the help output.

Screenshot 2025-09-25 at 11 52 01 AM
Screenshot 2025-09-25 at 11 53 41 AM

natecook1000: AFAICT all of those issues are existing ones for the DocC reference output, not new ones that result from @Chamepp's changes. We should track these separately.

Tools/generate-docc-reference/Extensions/ArgumentParser+Markdown.swift
natecook1000: It doesn't look like any of our Docc/Markdown generation tests include titled option groups, which means we aren't seeing the difference in section ordering. Could you add a test that includes a couple titled option groups, so that we can verify the results of this change?

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testRollMarkdownReference().md
Chamepp: Thank you for the great suggestions you made @d-ronnqvist 🙏 This is a great roadmap for us to walk towards a more mature solution in the subsequent PR's.

Chamepp:
Sure I will add the test shortly and get back to you

Chamepp:
@natecook1000 I tested the sorting functionality and its working properly. Earlier we suggested that the generated documentation should follow the same structure and order as the commands help output. We have already covered this for default section titles, which the `Arguments` comes first and then we have `Options`. However, when our command have custom section titles, the generated documentation section sorting is different from the help output ordering.

For testing this, instead of creating another example command and adding more tests, which makes maintenance more complicated for just testing the sorting functionality, i came up with the idea of adding custom titles to an existing command instead. I chose the `Math` command for testing since we mostly refer to it for checking the documentation generation output for its great coverage.
Screenshot 2025-10-03 at 11 42 58 PM
Screenshot 2025-10-04 at 1 11 37 AM

The `math.stats.quantiles` was the best option for this, because it includes a lot of arguments and options and setting custom section titles for categorization was great for this. We can also test the sorting alongside this. Now I will work on my testing method and refine my solution to match help output sorting even when we have custom section titles and get back to you.

Chamepp:
@natecook1000 The generated documentation sorting is now matching the help output.

Screenshot 2025-10-04 at 9 39 06 PM
Screenshot 2025-10-04 at 9 38 28 PM

PR #754, opened 2025-04-19
display arguments as term lists in docc-flavored markdown generation

### LINKED ISSUE #721: generate-docc-reference: use term lists for arguments

heckj:
- added a Boolean to distinguish regular (GH-flavored) markdown from docc-flavored markdown
- represent the arguments in generated files with term lists for the docc-flavored variant
- updated the snapshot tests to work both GH and DocC flavored markdown generation
- removed an added "\n" in the recording of content for snapshot tests, which caused the snapshot tests some issues (recorded tests with new values wouldn't pass on second-run)

resolves #721 

### Checklist

### DIFF
--- a/Sources/ArgumentParserTestHelpers/TestHelpers.swift
+++ b/Sources/ArgumentParserTestHelpers/TestHelpers.swift
@@ -530,7 +530,7 @@ extension XCTest {
     } else {
       command = [
         "generate-docc-reference", commandURL.path,
-        "--output-directory", "-"
+        "--output-directory", "-",
       ]
     }
     let actual = try AssertExecuteCommand(
--- a/Sources/ArgumentParserTestHelpers/TestHelpers.swift
+++ b/Sources/ArgumentParserTestHelpers/TestHelpers.swift
@@ -525,7 +525,7 @@ extension XCTest {
       command = [
         "generate-docc-reference", commandURL.path,
         "--output-directory", "-",
-        "--docc-flavored", "true",
+        "--style", "docc",
       ]
     } else {
       command = [
--- a/Sources/ArgumentParserTestHelpers/TestHelpers.swift
+++ b/Sources/ArgumentParserTestHelpers/TestHelpers.swift
@@ -448,7 +448,7 @@ extension XCTest {
       ProcessInfo.processInfo.environment["RECORD_SNAPSHOTS"] != nil
 
     if record || recordEnvironment || !snapshotExists {
-      let recordedValue = actual + "\n"
+      let recordedValue = actual
       try FileManager.default.createDirectory(
         at: snapshotDirectoryURL,
         withIntermediateDirectories: true,
@@ -507,8 +507,9 @@ extension XCTest {
       line: line)
   }
 
-  public func assertGenerateDoccReference(
+  public func assertGeneratedReference(
     command: String,
+    doccFlavored: Bool,
     record: Bool = false,
     test: StaticString = #function,
     file: StaticString = #filePath,
@@ -519,10 +520,19 @@ extension XCTest {
     #endif
 
     let commandURL = debugURL.appendingPathComponent(command)
-    let command = [
-      "generate-docc-reference", commandURL.path,
-      "--output-directory", "-",
-    ]
+    let command: [String]
+    if doccFlavored {
+      command = [
+        "generate-docc-reference", commandURL.path,
+        "--output-directory", "-",
+        "--docc-flavored", "true",
+      ]
+    } else {
+      command = [
+        "generate-docc-reference", commandURL.path,
+        "--output-directory", "-"
+      ]
+    }
     let actual = try AssertExecuteCommand(
       command: command,
       file: file,
--- a/Tools/generate-docc-reference/GenerateDoccReference.swift
+++ b/Tools/generate-docc-reference/GenerateDoccReference.swift
@@ -36,6 +36,14 @@ extension GenerateDoccReferenceError: CustomStringConvertible {
   }
 }
 
+/// The flavor of generated markdown to emit.
+enum OutputStyle: String, EnumerableFlag, ExpressibleByArgument {
+  /// DocC-supported markdown
+  case docc
+  /// GitHub-flavored markdown
+  case github
+}
+
 @main
 struct GenerateDoccReference: ParsableCommand {
   static let configuration = CommandConfiguration(
@@ -53,7 +61,7 @@ struct GenerateDoccReference: ParsableCommand {
   @Option(
     name: .shortAndLong,
     help: "Use docc flavored markdown for the generated output.")
-  var doccFlavored: Bool = false
+  var style: OutputStyle = .github
 
   func validate() throws {
     if outputDirectory != "-" {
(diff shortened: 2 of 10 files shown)


Sources/ArgumentParserTestHelpers/TestHelpers.swift
```diff
@@ -448,7 +448,7 @@ extension XCTest {
       ProcessInfo.processInfo.environment["RECORD_SNAPSHOTS"] != nil
 
     if record || recordEnvironment || !snapshotExists {
-      let recordedValue = actual + "\n"
```
heckj, line 451: I couldn't see an impact of removing this, but I'm not sure why it was there originally. With the extra new-line added to the content before it's written, the snapshot mechanism fails because a new line isn't added prior to asserting the equivalent results when the file exists. I thought the best option might be to remove this, but I can adjust in the other direction (add a newline before comparison) on the flip side if that's preferred.

Tools/generate-docc-reference/GenerateDoccReference.swift
```diff
@@ -101,22 +111,28 @@ struct GenerateDoccReference: ParsableCommand {
...
+  /// Generates a markdown file from the CommandInfoV0 object you provide.
+  /// - Parameters:
+  ///   - command: The command to parse into a markdown output.
+  ///   - directory: The directory to save the generated markdown file, printing it if `nil`.
+  ///   - doccFlavored: A Boolean value the indicates whether to generate docc-flavored markdown.
+  func generatePages(from command: CommandInfoV0, savingTo directory: URL?, doccFlavored: Bool)
     throws
   {
-    let page = command.toMarkdown([])
+    let page = command.toMarkdown([], doccFlavored: doccFlavored)
```
Chamepp, line 139: I don’t have the same level of expertise as you and the other mentors to make the final decision, but I thought it would be great to share my insights with you.

In my opinion this is a great idea in general. But I believe the single responsibility that the `GenerateDoccReference.swift` follows, is to generate docc compatible markdown files. We should also keep in mind that the `ArgumentParser+Markdown.swift` is also an extension to this plugin and `toMarkdown()` should follow the same purpose.

Also from a long term perspective of the project and the direction it follows, our goal is to generate documentation that integrates with DocC and in the last stages of the project we want to generate a symbol graph that is compatible with docc and adding the cli documentation generation support from docc instead of plugins.

The two problems that may arise by adding GFM support are:
1. GFM can be previewed but DocC is what will ship inside .doccarchive, SwiftPM websites and Xcode documentation.
2. Supporting both formats adds complexity especially when DocC evolves.

But maybe Im missing a point. Could you please elaborate more on why we need GFM output and what value does it bring if our primary goal is feeding DocC?

If GitHub presentation is still a concern, we might consider another solution that adapts the DocC flavored markdown for the GitHub interface rather than changing the source format itself.

heckj: I added this option based on an earlier conversation with @cmcgee1024 and @rauhul  about intentions for the content - was it aimed at creating markdown to host in Github, or DocC specific. The conclusion at the time was that it would be nice to serve both purposes, and since Github's markdown display doesn't support term lists, I didn't want to break one side to enable nicer formatting on the other.

I agree that this adds complexity, but given that we're generating DocC flavored markdown I thought the patterns that might be in one or the other would be sufficiently limited to warrant it. (And yeah, it's also a little weird to have the name of the tool be DocC specific if we want just GH flavored markdown from it)

Sources/ArgumentParserTestHelpers/TestHelpers.swift
Chamepp: Yes. I faced the same issue when I was working on https://github.com/apple/swift-argument-parser/issues/755. I opened an issue for this, including steps to reproduce, possible reasons why this line is added and possible solutions on how we can solve this [here](https://github.com/apple/swift-argument-parser/issues/755) so others have an easier time reviewing this. I think we should include this new line at the end and I believe your second solution is the better one.

Tools/generate-docc-reference/GenerateDoccReference.swift
heckj: Chris and Rauhul joined the DocC workgroup meeting in February to talk about this plugin shortly after they upstreamed it from swiftly into the argument parser library. The meeting wasn't recorded, but there are notes at https://forums.swift.org/t/documentation-workgroup-meeting-february-10th-2025/77710/2 

This was a far more informal conversation, not a "decision making" meeting.

Tools/generate-docc-reference/GenerateDoccReference.swift
```diff
@@ -50,6 +50,11 @@ struct GenerateDoccReference: ParsableCommand {
     help: "Directory to save generated docc reference. Use '-' for stdout.")
   var outputDirectory: String
 
+  @Option(
+    name: .shortAndLong,
+    help: "Use docc flavored markdown for the generated output.")
+  var doccFlavored: Bool = false
```
rauhul, line 56: I think I might prefer an option more like `--style <github|docc>`

Tools/generate-docc-reference/GenerateDoccReference.swift
```diff
@@ -71,6 +76,8 @@ struct GenerateDoccReference: ParsableCommand {
 
   func run() throws {
     let data: Data
+    // runs the tool with the --experimental-dump-help argument to capture
```
rauhul, line 79: these comments are nice, but the entire body of this file is nearly identical to the version in generate manual... might be worth duplicating them?

Tools/generate-docc-reference/GenerateDoccReference.swift
rauhul: this is definitely a good direction, I think the plugin should be renamed to generate-reference (like @cmcgee1024's original impl was named)

rauhul (approved):

Tools/generate-docc-reference/GenerateDoccReference.swift
heckj: Easy change, I'll set it up.

Tools/generate-docc-reference/GenerateDoccReference.swift
heckj: I didn't even look there, but I'll go review - this was all me commenting to future me about what was happening and in play.

heckj:
I went through and renamed the plugin to `generate-reference` (from `generate-docc-reference`), but that cascaded to a whole bunch of API breakage that just didn't seem warranted for this, so I backed out that commit entirely.

I'm totally game to run through and rename it again in the future, but we'll probably want to wrangle that alongside a major semver version bump, since we're renaming the whole executable of the plugin.

rauhul (approved):

heckj:
The soundness checks asserted it was a breaking change, and I wanted to make the rest of this easier rather than piling up deltas. I'll open a new PR to change the name of the plugin. I _think_ swiftly is already using it as it stands, so there may need to be a coordinated PR with them to keep from breaking their CI workflows as we adjust. I'll dig and get it set up.

rauhul:
@swift-ci please test

PR #695, opened 2025-02-05
Refactor Bash completions to use ToolInfo

rauhul:
Rebases the implementation of the BashCompletionGenerator to use
ToolInfo from ArgumentParserToolInfo instead of digging through the
command structure. This helps us decouple the implementation of Argument
parsing from the generation of supplemental content such as docs,
man-pages, completion scripts, help menus and more.

Issue #671, opened 2024-09-27
Problems with auto-created `--help` flag, `-h` flag, & `help` subcommand in generated completion scripts & help output


rgoldberg:
The following problems from an SAP-1.5.0-generated zsh completion script for the auto-created `--help` flag, the auto-created `-h` flag, & the auto-created `help` subcommand might occur for SAP-generated bash and/or fish scripts, too, but I'll just detail the problems seen in zsh.

I'm writing a command-line app that has subcommands (let's call it `command`).

SAP automatically created a `help` subcommand for `command`.

SAP also automatically created synonymous `--help` & `-h` top-level flags that output the same usage info as the `help` subcommand.

### `command --help|-h|help` output
The auto-created `help` subcommand is not included in the subcommand list output by `command --help`, `command -h`, or `command help`. It should be inserted alphabetically into the list for each of the 3 commands.

### `command --help|-h <tab>` completion
When I tab complete after a space after `command --help` or `command -h`, the completion list contains all the subcommands (including `help`).

All arguments after `command --help` or `command -h`, however, are ignored by SAP when the command is run; the command just outputs the top-level help, i.e. the list of subcommands along with top-level options.

The completion list after `command --help` or `command -h` should thus be empty.

I'd imagine that other top-level flags / options are ignored even if they're before `--help` or `-h`; if so, then the presence of any other flag / option / whatever should move `--help` & `-h` from any subsequent completion lists for the command / subcommand. 

### `command subcommand --help|-h <tab>` completion
SAP also auto-created `--help` & `-h` flags for each of my subcommands. The behavior of the top-level-command `--help` & `-h` flags should be replicated for the same two flags for each subcommand.

### `command help <tab>` completion
When I tab complete after a space after `command help`, the completion list is empty. If I manually type in `command help subcommand`, I get the help for `subcommand`.

The completion list after `command help` should thus include the list of subcommands.

PR #764, opened 2025-05-07
Refactor completion script generation to use ToolInfoV0

### LINKED ISSUE #758: Refactor completion script generation to use ToolInfoV0
@rauhul @natecook1000

Refactor completion script generation to use ToolInfoV0 for all 3 supported shells: bash, fish &
zsh.

@rauhul previously submitted PR #695 that refactored bash. He graciously sidelined his PR to
allow me to merge my numerous pending completion improvements, which have now all been
merged.

My WIP draft PR refactors all 3 shells to use ToolInfoV0, except it has 3 issues that I think require ToolInfoV0
to be updated:

1. ToolInfoV0 must include `parsingStrategy` in `ArgumentInfoV0` (needed for zsh).
2. The `HelpCommand` that is injected into ToolInfoV0 has a `--version` flag, but it is not seen
   by the completion code. Might be useful to try to resolve all issues from #671, too.
3. Nonexclusive flags implemented via an array of enum cases are represented as different names
   for the same `ArgumentInfoV0`, but each case should have its own `ArgumentInfoV0`. If such
   enum cases can have multiple names each, then all names for a single case should be in the
   same `ArgumentInfoV0`. Example enum & property (from `CompletionScriptTests.swift`):
    ```swift
    enum Kind: String, ExpressibleByArgument, EnumerableFlag {
        case one, two
        case three = "custom-three"
    }

    @Flag var allowedKinds: [Kind] = []
    ```

Bonus issue: the following seems similarly inconsistent in all of SAP that I've seen, not just
in ToolInfoV0. `case three = "custom-three"` from above is considered as `--three` for a flag
from `allowedKinds`, but as `custom-three` as an option value for `--kind` from
`@Option() var kind: Kind`. This is the case in the original completion scripts, in ToolInfoV0
& even in SAP's command-line argument verification. It seems to me that it should be consistent
throughout; of the 2 options, `custom-three` seems more sensible.

ToolInfoV0 and/or SAP might have other issues, but I haven't isolated them.

Note that my PR for this no longer replaces `-` with `_` in bash shell function names because:

- it's unnecessary (`-` is supported in function names; `-` is not supported only in variable names)
- bash is now consistent with the other 2 shells, and can use the same Swift code to generate function names
- it avoids (albeit exceedingly unlikely) potential name clashes between, e.g., subcommands named `a-b` & `a_b`

I can split it out into a separate PR, if necessary…

rgoldberg:
@rauhul @natecook1000

Refactor completion script generation to use ToolInfoV0 for all 3 supported shells: bash, fish &
zsh.

@rauhul previously submitted PR #695 that refactored bash. He graciously sidelined his PR to
allow me to merge my numerous pending completion improvements, which have now all been
merged.

My WIP draft PR refactors all 3 shells to use ToolInfoV0. It builds, and passes the format check,
but the generated completion scripts for each shell aren't what they should be, I think because 3
issues that I think require ToolInfoV0 to be updated. The test scripts contain what I believe to be
the desired output.

The 3 issues that I think exist in ToolInfoV0 are:

1. ToolInfoV0 must include `parsingStrategy` in `ArgumentInfoV0` (needed for zsh).
2. The `HelpCommand` that is injected into ToolInfoV0 has a `--version` flag, but it is not seen
   by the completion code. Might be useful to try to resolve all issues from #671, too.
3. Nonexclusive flags implemented via an array of enum cases are represented as different names
   for the same `ArgumentInfoV0`, but each case should have its own `ArgumentInfoV0`. If such
   enum cases can have multiple names each, then all names for a single case should be in the
   same `ArgumentInfoV0`. Example enum & property (from `CompletionScriptTests.swift`):
    ```swift
    enum Kind: String, ExpressibleByArgument, EnumerableFlag {
        case one, two
        case three = "custom-three"
    }

    @Flag var allowedKinds: [Kind] = []
    ```

Bonus issue: the following seems similarly inconsistent in all of SAP that I've seen, not just
in ToolInfoV0. `case three = "custom-three"` from above is considered as `--three` for a flag
from `allowedKinds`, but as `custom-three` as an option value for `--kind` from
`@Option() var kind: Kind`. This is the case in the original completion scripts, in ToolInfoV0
& even in SAP's command-line argument verification. It seems to me that it should be consistent
throughout; of the 2 options, `custom-three` seems more sensible.

ToolInfoV0 and/or SAP might have other issues, but I haven't isolated them.

Note that this PR no longer substitutes `_` in place of `-` in bash shell function names because:

- it's unnecessary (`-` is supported in function names; `-` is not supported only in variable names)
- bash is now consistent with the other 2 shells, and can use the same Swift code to generate function names
- it avoids (albeit exceedingly unlikely) potential name clashes between, e.g., subcommands named `a-b` & `a_b`

I can split out that change into a separate PR, if necessary…

Please let me know if the problems I've mentioned are indeed in ToolInfoV0, or if I'm just using it wrong.

Also, please let me know who might be able to fix ToolInfoV0, and if they'll be able to do so anytime soon.

If it won't take too much time, it would be nice to get this in to 1.6.0.

Also, for the 1.6.0 release notes, I'd very much appreciate it if you could mention that it is known that preexisting bugs from 1.5.0 still exist in completion script generation, and that they will be fixed in subsequent releases. I don't want people to think that I think that the current state is close to bug free. Thanks.

Resolve #758 #780

### Checklist


rgoldberg:
@natecook1000 @rauhul Do my theorized issues seem to exist in ToolInfoV0? Or must I modify my PR in some way instead of ToolInfoV0 being changed?

If the former, will anyone else work on ToolInfoV0 updates soon? I'm working on some other completion fixes, so I'd prefer the code not to diverge too much, and I'd prefer to work using the nicer ToolInfoV0 interface, if possible.

If you don't have the bandwidth to handle ToolInfoV0 now, I completely understand.

rauhul:
@rgoldberg sorry about the radio silence, speaking for myself im super super busy leading up to early June and can't take a close look at SAP changes until deadlines pass. Sorry :(

rgoldberg:
@rauhul No problem. Completely understandable.

If no one else can look into ToolInfoV0, then I might delve in.

Good luck with your other work!

rgoldberg:
@rauhul @natecook1000 I confirmed & fixed the 3 issues that I identified in ToolInfoV0. I don't know if my changes will cause any issues elsewhere for ToolInfoV0, but all the tests pass (after I applied some updates to expected test output that all seem correct to me).

I have not looked into the SAP-wide issue because that existed before the attempt to switch to ToolInfoV0. I'll make a separate GitHub issue for that issue in the future.

I don't know the order in which my 3 extant PRs should be reviewed & merged to simplify merging. I guess probably this first, if you have the time to review it. If not, review & merge in whatever order you prefer.

Thanks again.

rgoldberg:
@jaredgrubb This PR already fixes #780, so #781 should be unnecessary. I think the solution here is more optimal, but, if people disagree, please let me know.

Tests/ArgumentParserUnitTests/Snapshots/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*]
```
rauhul, line 1: I'd like to avoid this. Instead the code generator should not emit trailing whitespace and should generate a trailing newline. We can make this its own GitHub issue to resolve later to avoid making this even more complicated.

Tests/ArgumentParserUnitTests/Snapshots/testADumpHelp().json
```diff
@@ -25,6 +25,7 @@
             "name" : "enumerated-option"
           }
         ],
+        "parsingStrategy" : "default",
```
rauhul, line 28: Can we factor the Toolset.jso changes into a different PR, that would be a lot easier to review in isolation.

Tests/ArgumentParserUnitTests/Snapshots/.editorconfig
rauhul: We can keep it for now, I agree its annoying to deal with the editor messing up the file during development.

Tests/ArgumentParserUnitTests/Snapshots/testADumpHelp().json
rgoldberg: @rauhul The ToolInfo changes that result in JSON changes are necessary for the migrated completion scripts using ToolInfo to generate the same scripts as before (except for the fix to #780, which outputs improved scripting).

If you want, I can split it out into a new PR, but you can also just look at each commit of this PR individually, as they're sliced to be separate units of work described by their commit messages. The first 3 commits are updates to ToolInfo; the next 3 commits are updating bash, fish, and then zsh to use ToolInfo; the next commit removes a vestigial function; and the last commit adds the `.editorconfig`s.

If you want a separate PR, can I put the first 3 commits in it? Or must I submit a separate PR per commit?

Can I keep bash, fish, zsh, function removal & `.editorconfig`s in this PR?

natecook1000:
@swift-ci Please test

PR #767, opened 2025-05-12
Improve flag & option name quoting in completion generation for all 3 shells

### LINKED ISSUE #761: valueName's containing colons are not escaped in completion scripts
Here's an example of an option [from Tart](https://github.com/cirruslabs/tart/blob/280a31f7079e490b36e4cce91a860015e47d366c/Sources/tart/Commands/Run.swift#L119-L143), the option has a `valueName` that contains a colon:

```swift
@Option(help: ArgumentHelp("""
Additional disk attachments with an optional read-only and synchronization options (e.g. --disk="disk.bin", --disk="ubuntu.iso:ro", --disk="/dev/disk0", --disk "ghcr.io/cirruslabs/xcode:16.0:ro" or --disk="nbd://localhost:10809/myDisk:sync=none")
""", valueName: "path[:options]"))
var disk: [String] = []
```

This is a common notation for Unix-like CLI tools meaning that you can either specify a `path`, or additionally have `options`, but after a colon.

The resulting Zsh completion looks like this:

```shell
_tart_run() {
    local -i ret=1
    local -ar arg_specs=(
        [...]
        '*--disk[Additional disk attachments with an optional read-only and synchronization options (e.g. --disk="disk.bin", --disk="ubuntu.iso:ro", --disk="/dev/disk0", --disk "ghcr.io/cirruslabs/xcode:16.0:ro" or --disk="nbd://localhost:10809/myDisk:sync=none")]:path[:options]:'
        [...]
    )
    _arguments -w -s -S : "${arg_specs[@]}" && ret=0

    return "${ret}"
}
```

Here, the colon slips into the completion unescaped, so attempting to complete the `tart run --disk` results in the following Zsh error:

```shell
% tart run --disk _arguments:465: command not found: options]
```

Happens on the `main` branch of the ArgumentParser.

rgoldberg:
Improve flag & option name quoting in completion generation for all 3 shells.

Improve zsh quoting.

Many quoting issues remain.

To fix them, there should first be enforced & documented limits on the acceptable characters in various values throughout SAP to avoid unnecessary quoting (see #768).

Otherwise, there's a massive rabbit hole of quoting issues. Given the way that tests are currently setup, adding tests for a fixed issue in one shell requires the same issue to be fixed in all 3 shells, which is cumbersome.

e.g., I haven't ensured that fish's `argparse` function will work with shell characters that this PR now quotes. Beforehand, they would just break the whole script by causing syntax errors. Now, they might just mess up an option spec for `argparse`, which, while it could mess up argument parsing, will not cause syntax errors.

It will be a process of iterative improvement, rather than fixing everything at once.

Resolve #761

### Checklist


natecook1000:
@rgoldberg Thanks again for this fix!

Issue #555, opened 2023-02-21
Permit asynchronous code to be given to the custom completion type


thafner0:
At the moment the `custom` case of the `CompletionKind` enumeration supports being given a custom closure to generate completions on the command line. However, since the provided closure must be synchronous, it is difficult to incorporate functionality that requires swift's concurrency features. The lack of this functionality prevents certain tasks from being possible with these completions and causes other tasks to be possible, but only in a non-thread safe way. Examples of this include working with databases (such as those managed by Apple's Core Data framework).

Example:

```swift
struct SomeCommand: AsyncParseableCommand {

@Argument(help: "<Helpful help>", completion: .custom({ arguments in
return try? await completionGenerator(arguments: arguments) ?? []
}))
var argument: String

mutating func run() async throws {
// Code that makes the command Run
}
}

@MainActor
private func completionGenerator(arguments: [String]) async throws -> [String] {
// Code that generates completions that has some asynchronous system in use and/or must be run on the main actor.
}
```

luispadron:
This would be great, along with `completion` we could really use this for `transform`. 

@natecook1000 is this something you think would be wanted? I could work on this if you don't think it's too complicated as a first task

rgoldberg:
@natecook1000 @rauhul Is this issue on the roadmap for imminent development? Would it be difficult to implement?

Is there any public roadmap for development?

Thanks for any info.

rgoldberg:
I have code that implements this that I will submit as a PR after #770 has been merged.

PR #770, opened 2025-05-13
Replace the cursor index within the completing word parameter of custom-completion closures with a `String` completion prefix parameter

### LINKED ISSUE #769: Replace second index parameter of custom-completion closures with a String prefix parameter
@natecook1000 @rauhul I think this is an urgent fix that should be merged before any new version is released.

I will make a PR now. 

Replace the cursor index within completing word parameter of custom-completion closures with a `String` `completionPrefix` parameter.

Most of the time, a completion function will just want the prefix of the word being completed before the cursor.

Code shouldn't have to be complex to get it, but I didn't realize how cumbersome `Int` indices are for Swift `String`s (I'm fairly new to Swift).

Code can easily get the index from the prefix. This should be a much nicer experience for users.

rgoldberg:
Replace the cursor index within the completing word parameter of custom-completion closures with a `String` completion prefix parameter.

Resolve #769

@natecook1000 @rauhul I think that this is an urgent improvement.

The Swift substring API is exceedingly verbose, and this will make using custom completions much easier.

We don't want to release the current API with the 2 indices to users because then it will need to be supported indefinitely, despite being suboptimal.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -465,12 +465,7 @@ extension CommandParser {
       let completingPrefix: String
       if let completingArgument = args.last {
         completingPrefix = String(
-          completingArgument.prefix(
-            upTo: completingArgument.index(
-              completingArgument.startIndex,
-              offsetBy: cursorIndexWithinCompletingArgument
-            )
-          )
+          completingArgument.prefix(cursorIndexWithinCompletingArgument)
         )
       } else if cursorIndexWithinCompletingArgument == 0 {
         completingPrefix = ""
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -456,16 +456,32 @@ extension CommandParser {
       }
 
       guard
-        let s = args.popFirst(),
-        let cursorIndexWithinCompletingArgument = Int(s)
+        let arg = args.popFirst(),
+        let cursorIndexWithinCompletingArgument = Int(arg)
       else {
+        throw ParserError.invalidState
+      }
+
+      let completingPrefix: String
+      if let completingArgument = args.last {
+        completingPrefix = String(
+          completingArgument.prefix(
+            upTo: completingArgument.index(
+              completingArgument.startIndex,
+              offsetBy: cursorIndexWithinCompletingArgument
+            )
+          )
+        )
+      } else if cursorIndexWithinCompletingArgument == 0 {
+        completingPrefix = ""
+      } else {
         throw ParserError.invalidState
       }
 
       completions = complete(
         Array(args),
         completingArgumentIndex,
-        cursorIndexWithinCompletingArgument
+        completingPrefix
       )
     case .customDeprecated(let complete):
       completions = complete(args)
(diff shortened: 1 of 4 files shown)


rgoldberg:
@natecook1000 @rauhul I also have a PR for #555 (supporting `async` functions as custom completion handlers). I will submit it after this PR has been merged because it is branched from this PR's branch. It's not large at all, so shouldn't take long to review.

Please don't release a new version of Swift Argument Parser without checking with me first, because I might have other useful submissions queued up. Obviously, if I don't respond in a timely manner, go ahead & release new versions whenever you want, but if you could check with me first, I'd appreciate it, and users might get more functionality in the new release. Most importantly, if I realize that a new interface is suboptimal, it should be cleaned up ASAP before being officially released so it won't need to be maintained indefinitely, and so it won't be unpleasant for users.

Thanks again.

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -456,16 +456,32 @@ extension CommandParser {
...
 
+      let completingPrefix: String
+      if let completingArgument = args.last {
+        completingPrefix = String(
+          completingArgument.prefix(
+            upTo: completingArgument.index(
+              completingArgument.startIndex,
+              offsetBy: cursorIndexWithinCompletingArgument
+            )
+          )
```
natecook1000, line 473: There's a prefix overload that just takes the count of elements to include, so this can be simplified to `.prefix(cursorIndexWithinCompletingArgument)`.

natecook1000:
@rgoldberg This change makes sense, since the conversion from `Int` to substring can be error-prone. Just to make sure I'm understanding this correctly: if a user has the following text on the command line, with their cursor as marked:

```
$ example --foo 1 --bar 2 baz bizzle
                           ^ cursor here, on 'a' in 'baz'
```

...then the parameters passed to the custom completion closure would be as if it were called like:

```swift
complete(["example", "--foo", "1", "--bar", "2", "baz", "bizzle"], 5, "b")
```

Is that right?

rgoldberg:
@natecook1000 My zsh cursor is always displayed between characters, not on them. Maybe you can switch cursor modes, but for the cursor in-between character mode, it would be between the `b` & the `a`. Typing backspace would delete the `b`. Typing delete would delete the `a`. Typing a normal character would insert it between the 2. In that case, the arguments you listed are correct. (sorry to be pedantic, but I want to ensure that I am as unambiguous as possible).

Thanks for looking into this.

Sources/ArgumentParser/Parsing/CommandParser.swift
rgoldberg: I will update the PR now. I'm still new to Swift, so my API knowledge is limited. I didn't even realize how verbose it is to get substrings in Swift. Thanks for pointing this out. All similar tips will be greatly appreciated, as it will produce much nicer code & get me up to speed much faster. Thanks.

rgoldberg: @natecook1000 Pushed the fix. Thanks again for the tip.

natecook1000:
@swift-ci Please test

rgoldberg:
The failing API breakage check should be OK because the replaced API was never released in a new version.

natecook1000:
@rgoldberg We're on the same page! The default cursor indicator for me is a block, even though I think of that as the cursor for overwrite mode (though I can't remember what platform that was on). In any case, you're describing what I meant to.

And exactly, merging over the API breakage check.

Issue #780, opened 2025-05-17
Bash completion: Does not handle sub-commands with hyphens


jaredgrubb:
The bash completion doesn't handle subcommands that have hyphens in them correctly.

**ArgumentParser version:** `1.5.0` and main
**Swift version:** N/A

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce
The unit tests have a subcommand and the generated bash completion doesn't actually work. So the tests already reproduce the issue if the bash completion is loaded.

### Expected behavior
Subcommand completion should be correct.

### Actual behavior
```
_base_test_sub-command: command not found
```

PR #781, opened 2025-05-17
Bash Completion: fix bug in dash vs underscore (#780)

jaredgrubb:
The generated script tries to call sub-functions that have hyphens in them instead of underscores.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ Add new items at the end of the relevant section under **Unreleased**.
 
 ## [Unreleased]
 
-*No new changes.*
+### Fixes
+
+- Bash completion scripts correctly handles subcommands with hyphens. ([#780])
 
 ---
 
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -290,7 +290,7 @@ extension [ParsableCommand.Type] {
             case "${subcommand}" in
             \(subcommands.map { $0._commandName }.joined(separator: "|")))
                 # Offer subcommand argument completions
-                "\(functionName)_${subcommand}"
+                "\(functionName)_${subcommand//-/_}"
                 ;;
             *)
                 # Offer subcommand completions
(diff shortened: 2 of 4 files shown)


jaredgrubb:
@rgoldberg I see you've been adjusting completions lately. I want to make sure you see this PR.

rgoldberg:
@jaredgrubb thanks for the heads up. Just to double check: the problem exists in both 1.5.0 & in main?

rgoldberg:
@jaredgrubb Also, I think that the solution is more likely to be to ensure that generated function names do not replace `-` with `_`.

Otherwise, you can have a name conflict for the functions between subcommand `a-b` & `a_b`. It's unlikely, but possible that someone will have both.

rgoldberg:
@jaredgrubb FYI, my PR to redo completions using ToolIfnoV0 (#764) already has the fix that I proposed (no longer replacing `-` with `_`).

If you agree that works, it probably makes sense to close this PR because, IMO, my PR solves the problem more completely.

jaredgrubb:
I think that sounds ok -- as long as that PR is likely to merge before the next release? 
Otherwise, this PR is a very nice quick-fix for something that's broken today and your PR could rebase on top of it until it's ready.

rgoldberg:
@jaredgrubb I hope my PR will be merged before the next release. It & my other PRs seem ready to go. If there are issues wit migrating to ToolInfoV0, I can always extract my solution for this from that PR and submit it by itself.

Hopefully I'll hear back from the project members soon.

DO you think it makes sense to mark this PR as a draft so they don't merge it then mine, causing unnecessary merge issues, etc.?

Or, could you mention in the initial comment for this PR that a more complete solution already exists in #764 & can be extracted?

It just doesn't make sense to me to go for a quicker solution that has naming clash issues when the full solution is already available.

natecook1000:
@jaredgrubb Thanks for looking at this! I'm working on reviewing #764, and will make sure it lands before the next release, so that change will resolve the underlying issue.

jaredgrubb:
Sounds good! I look forward to the fix!

rgoldberg:
@jaredgrubb Thanks for noticing this, and for giving me a heads up. Sorry for having preempted your fix.

If you notice any other completion problems, please let me know.

Many completion bugs are still extant, with many due to string quoting, escaping, and/or delimiting issues. I just haven't had the chance to find & fix them all yet.

PR #782, opened 2025-05-19
Support async custom completion closures

### LINKED ISSUE #555: Permit asynchronous code to be given to the custom completion type

rgoldberg:
Support async custom completion closures.

Resolve #555

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -380,7 +380,7 @@ extension [ParsableCommand.Type] {
 
         """
 
-    case .custom:
+    case .custom, .customAsync:
       // Generate a call back into the command to retrieve a completions list
       return """
         \(addCompletionsFunctionName) -W\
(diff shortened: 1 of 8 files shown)


rgoldberg:
@natecook1000 This PR might be simpler to review than #764, so you might want to get this out of the way first.

Thanks for reviewing / merging everything.

natecook1000 (commented):
Looks good! Can you add a couple tests that verify that the async version is usable?

rgoldberg:
@natecook1000 Added a multi-shell custom completion test for async custom completion handlers.

natecook1000:
@swift-ci Please test

rgoldberg:
@natecook1000 I pushed an attempt to fix the Linux build issues. I didn't get any build or test issues on macOS 15.5, but I don't know how to run the Linux tests myself, or how to make the macOS tests mimic the Linux tests.

I assume that the Linux tests have stricter concurrency checking enabled than the macOS & Windows tests do. Maybe they should be changed to be as strict as the Linux tests. Should I open an issue for that? Does one already exist?

natecook1000:
@swift-ci Please test

PR #783, opened 2025-05-22
build: conditionally add Foundation dependency edge

compnerd:
When building without `-DFoundation_DIR=...`, we should not wire up a dependency on Foundation and instead rely on autolinking and the driver to collude to resolve the linkage. Failure to do so will instill a rogue `-lFoundation.lib` when linking on Windows.


compnerd:
@swift-ci please test

natecook1000 (approved):

PR #789, opened 2025-06-27
Handle Swift 6.2 sendability changes

natecook1000:
Adds `SendableMetatype` conformance to the `ExpressibleByArgument` and `ParsableArguments` protocols, enabline existential use in Swift 6.2.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
+++ b/Sources/ArgumentParser/Parsable Types/ExpressibleByArgument.swift
@@ -10,7 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 /// A type that can be expressed as a command-line argument.
-public protocol ExpressibleByArgument {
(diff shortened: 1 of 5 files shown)


natecook1000:
FWIW the [API breakage check](https://github.com/apple/swift-argument-parser/actions/runs/15934430842/job/44950935574?pr=789) only passes because it's using a 6.0 compiler and not a 6.2 compiler that would see the `SendableMetatype` requirement. That said, I don't believe this will be source breaking in practice because:
- it will only affect those using the Swift 6.2 compiler or greater
- it will only pose a problem for types with isolated conformances to `ParsableArguments` or `ExpressibleByArgument`, which there shouldn't really be any yet

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
With the updated approach the API breakage check fails now, but the notes above still apply.

PR #803, opened 2025-09-08
Adopt upcoming Swift language feature `MemberImportVisibility`

tshortli:
Enable the language feature flag for [SE-0444](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md). This instructs the compiler to be stricter about requiring import declarations in order to access member declarations.


rauhul:
@swift-ci test

rauhul:
@swift-ci test

rgoldberg:
@tshortli Quick questions: the [SE-0444 doc](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md) indicates that `MemberImportVisibility` was implemented in Swift 6.1; will enabling it affect a project when it is being compiled by an older version of the Swift compiler (6.0-)? I assume not.

If 6.0- compilers won't be affected, would enabling this flag for a project that can be built by Swift 6.0- compilers allow for certain code to compile on 6.1+, but not compile on 6.0- due to clashing implicitly imported symbols?

If that's the case, why even enable it now for Swift Argument Parser (SAP)? Would it make sense to enable it only after SAP requires 6.1+ for compilation?

Thanks for any insight you can provide. Sorry if I'm mistaken, but I've only enabled an experimental (though not upcoming) feature when all Swift compilers that should be able to compile my code support the feature.

tshortli:
> Quick questions: the [SE-0444 doc](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md) indicates that MemberImportVisibility was implemented in Swift 6.1; will enabling it affect a project when it is being compiled by an older version of the Swift compiler (6.0-)? I assume not.

If the compiler does not recognize the feature identifier specified with `-enable-upcoming-feature` and `-enable-experimental-feature` then it ignores flag entirely. Building ArgumentParser with `-enable-upcoming-feature MemberImportVisibility` when building with an older toolchain won't have any effect.

> If 6.0- compilers won't be affected, would enabling this flag for a project that can be built by Swift 6.0- compilers allow for certain code to compile on 6.1+, but not compile on 6.0- due to clashing implicitly imported symbols?

The source code will still be compatible with the older toolchains, because `MemberImportVisibility` simply requires that additional explicit imports be added to source code where previously those imports were only transitive. There shouldn't be any new ambiguities introduced by new imports just for older toolchains since those ambiguities would also be present with newer toolchains and therefore prevent the imports from being added in the first place.

> If that's the case, why even enable it now for Swift Argument Parser (SAP)? Would it make sense to enable it only after SAP requires 6.1+ for compilation?

No, I don't think it makes sense. Practically speaking that requirement would be very far away, and in the meantime we won't get the benefit in build environments that use newer compilers.

Issue #778, opened 2025-05-14
Short name option '-h' (conflicting with built-in help) is allowed, with strange results.


psummerland:
### Checklist
- [ x] If possible, I've reproduced the issue using the `main` branch of this package
- [ x] I've searched for [existing GitHub issues](https://github.com/apple/swift-argument-parser/issues)

### Steps to Reproduce

Add the `horse` option to the Repeat example provided with swift-argument-parser:

```swift
 @main
struct Repeat: ParsableCommand {
    @Option(name: .shortAndLong, help: "The number of times to repeat 'phrase'.")
    var count: Int? = nil

    @Option(name: .shortAndLong, help: "The name of the talking horse.")
    var horse: String

    @Flag(name: .shortAndLong,  help: "Include a counter with each repetition.")
    var includeCounter = false

    @Argument(help: "The phrase to repeat.")
    var phrase: String

    mutating func run() throws {

        let repeatCount = count ?? 2

        for i in 1...repeatCount {
            if includeCounter {
                print("\(i): \(horse) says: \"\(phrase)\"")
            } else {
                print(phrase)
            }
        }
    }
}
```
### Actual behavior

Help shows conflicting descriptions for "-h":

```
USAGE: repeat [--count <count>] [--include-counter] <phrase> --horse <horse>

ARGUMENTS:
  <phrase>                The phrase to repeat.

OPTIONS:
  -c, --count <count>     The number of times to repeat 'phrase'.
  -i, --include-counter   Include a counter with each repetition.
  -h, --horse <horse>     The name of the speaking horse.
  -h, --help              Show help information.
```
Works sometimes:
```
release> ./repeat --horse Secretariat "I trained hard and won." -i
1: Secretariat says: "I trained hard and won."
2: Secretariat says: "I trained hard and won."
```
What does "-h" mean?
```
release> ./repeat -h
Error: Missing value for '-h <horse>'
Help:  -h <horse>  The name of the speaking horse.
Usage: repeat [--count <count>] --horse <horse> [--include-counter] <phrase>
  See 'repeat --help' for more information.
```
Seems to mean the same as "--horse", so add the value:
```
release> ./repeat -h Secretariat
USAGE: repeat [--count <count>] --horse <horse> [--include-counter] <phrase>

ARGUMENTS:
  <phrase>                The phrase to repeat.

OPTIONS:
  -c, --count <count>     The number of times to repeat 'phrase'.
  -h, --horse <horse>     The name of the speaking horse.
[...]

PR #826, opened 2025-10-11
#778: Suppress builtin `-h,--help` from usage text when command already contains an option using `-h`

TiagoMaiaL:
This PR fixes part of the behavior described in #778, fixing the generated error and usage texts to not list the builtin `-h, --help`:

> I would expect -h to only be added if no existing flag or option already has that name.

I still need to address the other behaviors there, and to fix command invocation, applying the custom command and not the builtin help one.

#### Before

```
Error: Missing expected argument '<phrase>'

USAGE: repeat [--count <count>] [--include-counter] <phrase> --horse <horse>

ARGUMENTS:
  <phrase>                The phrase to repeat.

OPTIONS:
  --count <count>         How many times to repeat 'phrase'.
  -h, --horse <horse>     The name of the talking horse.
```

#### After

```
Error: Missing expected argument '<phrase>'

USAGE: repeat [--count <count>] [--include-counter] <phrase> --horse <horse>

ARGUMENTS:
  <phrase>                The phrase to repeat.

OPTIONS:
  --count <count>         How many times to repeat 'phrase'.
  -h, --horse <horse>     The name of the talking horse.
```

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Usage/HelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/HelpGenerator.swift
@@ -519,7 +519,19 @@ extension BidirectionalCollection where Element == ParsableCommand.Type {
       })
     else { return ArgumentSet() }
     self.versionArgumentDefinition().map { arguments.append($0) }
-    self.helpArgumentDefinition().map { arguments.append($0) }
+    
+    let shouldAppendBuiltinHelp = !arguments
+      .flatMap(\.names)
+      .contains { name in
+        if case .short(let char, _) = name {
+          return char == "h"
+        }
+        return false
+      }
+    
+    if shouldAppendBuiltinHelp {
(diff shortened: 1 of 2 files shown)


natecook1000:
Thanks for looking at this, @TiagoMaiaL! I think the core issue here is that there can be a conflict between the help names (which default to `-h` and `--help`, but can be set to other things in a command's configuration) and option/flag names defined inside a command. This fix explicitly looks for `-h`, but doesn't handle any other cases, and will suppress the help documentation from appearing in the help screen even if a tool has customized its help names.

A better approach could be to add a new validator that looks for this conflict, or to include the help names in the existing `UniqueNamesValidator`.

TiagoMaiaL:
I'll follow your guidance and open a new MR soon, thank you for reviewing @natecook1000

PR #838, opened 2025-12-11
Fix handling of visibility with nested OptionGroups.

bob-wilson:
The visibility of OptionGroups is ignored if they are nested inside another OptionGroup. Instead of trying to filter groups when creating an ArgumentSet, use the OptionGroup's visibility to modify the arguments within the group.

### Checklist

### DIFF
--- a/Tests/ArgumentParserUnitTests/HelpGenerationTests+GroupName.swift
+++ b/Tests/ArgumentParserUnitTests/HelpGenerationTests+GroupName.swift
@@ -233,6 +233,43 @@ extension HelpGenerationTests {
         """)
   }
 
+  fileprivate struct NestedGroups: ParsableArguments {
+    @OptionGroup(visibility: .hidden)
+    var flagsAndOptions: FlagsAndOptions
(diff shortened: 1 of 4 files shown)


Tests/ArgumentParserUnitTests/HelpGenerationTests+GroupName.swift
```diff
@@ -187,18 +187,23 @@ extension HelpGenerationTests {
...
     @OptionGroup(title: "Options Group", visibility: .hidden)
     var options: Options
 
-    @OptionGroup(visibility: .hidden)
-    var flagsAndOptions: FlagsAndOptions
-
-    @OptionGroup(visibility: .private)
-    var argsAndFlags: ArgsAndFlags
+    @OptionGroup()
+    var nested: NestedGroups
```
natecook1000: I think for completeness it would be best to keep this existing type and add a new `NestedHiddenGroups`/`testNestedHiddenGroups()` type/test function pair.

PR #853, opened 2026-01-15
GHA: Non-analysis workflows requires analysis to pass

bkhouri:
In order to prevent CI resources, run the "analysis" workflows first. If, and only if, they all pass, run other workflows that build and test the project.

This ensure we don't build when an analysis workflow will require the PR to be updated.

### Checklist


natecook1000:
It looks like the quick checks take about a minute, so this shouldn't prolong testing by much to gate the longer checks. Thanks, @bkhouri!

PR #854, opened 2026-01-15
Update version help command

donnywdavis:
I'm sure that this could be considered a personal preference, but wanted to submit to see what others think.

When checking for the version number of a cli I always assume that `-v` is available as a short name and try that before `--version`. However, using the built in version property on `CommandConfiguration` only enables the long format of `--version`. If we want to enable `-v` then we have to implement our own version flag and ignore the property in the configuration. 

This is just a small change to make `-v` available as a short name for the version flag.

```
OPTIONS:
  -v, --version           Show the version.
```

---

If we wanted to make this more flexible then we could extend `CommandConfiguration` to also accept a `NameSpecification` for the version. I just went with the simplest approach for now.

I also understand that an argument could be made that this would conflict with a flag called `verbose`. Someone may want to use `-v` as the short name for that. By making `-v` available  in `CommandConfiguration` it removes the option when having both of these flags in a single command. I feel like having the version number is more common than having a flag called verbose, so I personally wouldn't consider it a blocker as the verbose flag could just use the long name or a different custom short name. Just my opinion. Curious what y'all think.

### Checklist


rauhul:
I think this overlaps too much with using `-v` for `--verbose` @natecook1000 ?

natecook1000:
That's my thinking as well – I think the workaround to implement a version flag on its own is an acceptable one.

donnywdavis:
Sorry for the late reply on this. Thank y'all for having a look at this.

PR #855, opened 2026-01-16
Support async custom completion functions for `AsyncParsableCommand` via async/await

### LINKED ISSUE #852: Hangs for async custom completions
The implementation of async custom completions using `DispatchSemaphore` from #782 worked for me in the past, but async custom completions are hanging for me now, presumably due to deadlocks.

I don't know what caused the difference in behavior. I don't know if the hangs:

1. always existed under certain circumstances (& my current code is experiencing them while my older code didn't)
2. are caused by subsequent modifications to the async custom completion implementation
3. are caused by a newer Swift compiler, Swift runtime, and/or macOS
4. etc.

I'll try to whittle down a concise hang & non-hang examples in the future.

I have a local branch that fixes this that I can submit as a PR; it replaces `DispatchSemaphore` with `async/await` all the way down, so it only works for `AsyncParsableCommand`.

I can reinstate the `DispatchSemaphore` implementation in place for `ParsableCommand` if it still works for some async custom completions, or remove it if it no longer works at all (my current local branch does the latter, since it was simpler to investigate & implement that way, but it shouldn't be hard to reinstate the `DispatchSemaphore` implementation).

I'll investigated more (& potentially get feedback here), provide examples, clean up my local branch, and submit it as a PR.

**ArgumentParser version:**

main

**Swift version:**

swift-driver version: 1.127.14.1 Apple Swift version 6.2.3 (swiftlang-6.2.3.3.21 clang-1700.6.3.2)
Target: arm64-apple-macosx26.0

### Checklist

### Steps to Reproduce

Will provide a concise example in the future.

### Expected behavior

Async custom completion returns instead of hanging.

### Actual behavior

Async custom completion hangs instead of returning.

rgoldberg:
- Support async custom completion functions for AsyncParsableCommand via async/await.
  - Resolve #852.
  - No longer use `DispatchSemaphore`, so more modern standards compliant & will work for platforms that do not have `DispatchSemaphore`.
  - The existing `DispatchSemaphore` implementation hangs when any async custom completions run if the root command is a sync `ParsableCommand`; async custom completions do not hang because of `DispatchSemaphore` if the root command is an `AsyncParsableCommand`, regardless if the command containing the property for the completing argument is a sync `ParsableCommand` or an  `AsyncParsableCommand`. 
  - Async custom completions are thus not supported for root `ParsableCommand`.
- Make some related code more concise & consistent.
- Improve `ParsableArguments#_errorLabel` DocC (unrelated, but just noticed it should be improved).

A few potential issues with / questions about what you want from this PR:

- Async custom completions are supported only for `AsyncParsableCommand` root commands.
  - That seems OK to me because:
    - The existing implementation doesn't support root sync `ParsableCommand`.
    - I doubt there's any penalty to switch a root command from `ParsableCommand` to `AsyncParsableCommand`, since Swift Concurrency must be available, since the code is trying to use an async function for completion.
- If the root command is a sync `ParsableCommand`, a `ParserError.invalidState` is thrown during runtime when completion candidates are requested from an async custom completion function.
  - Do you want this detected at build time or completion script generation time?
    - If so, are there any utilities to facilitate detection at build time or completion script generation time?
  - Do you want a different error?
- Is the minor cleanup for tangentially related code acceptable here?
- Is the unrelated DocC update acceptable here?

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
@@ -25,6 +25,34 @@ public protocol AsyncParsableCommand: ParsableCommand {
 
 @available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
 extension AsyncParsableCommand {
+  /// Parses a new instance of this type from command-line arguments.
+  ///
+  /// - Parameter arguments: An array of arguments to use for parsing. If
+  ///   `arguments` is `nil`, this uses the program's command-line arguments.
+  /// - Returns: A new instance of this type.
+  /// - Throws: If parsing failed or arguments contains a help request.
+  public static func parse(
+    _ arguments: [String]? = nil
+  ) async throws -> Self {
+    try parse(try await parseAsRoot(arguments))
+  }
+
+  /// Parses an instance of this type, or one of its subcommands, from
+  /// command-line arguments.
+  ///
+  /// - Parameter arguments: An array of arguments to use for parsing. If
+  ///   `arguments` is `nil`, this uses the program's command-line arguments.
+  /// - Returns: A new instance of this type, one of its subcommands, or a
+  ///   command type internal to the `ArgumentParser` library.
+  /// - Throws: If parsing fails.
+  public static func parseAsRoot(
+    _ arguments: [String]? = nil
+  ) async throws -> ParsableCommand {
+    var parser = CommandParser(self)
+    let arguments = arguments ?? Array(CommandLine._staticArguments.dropFirst())
+    return try await parser.parse(arguments: arguments)
+  }
+
   /// Executes this command, or one of its subcommands, with the given arguments.
   ///
   /// This method parses an instance of this type, one of its subcommands, or
@@ -36,7 +64,7 @@ extension AsyncParsableCommand {
   ///   `arguments` is `nil`, this uses the program's command-line arguments.
   public static func main(_ arguments: [String]?) async {
     do {
-      var command = try parseAsRoot(arguments)
+      var command = try await parseAsRoot(arguments)
       if var asyncCommand = command as? AsyncParsableCommand {
         try await asyncCommand.run()
       } else {
--- a/Sources/ArgumentParser/CMakeLists.txt
+++ b/Sources/ArgumentParser/CMakeLists.txt
@@ -49,7 +49,8 @@ add_library(ArgumentParser
   Utilities/StringExtensions.swift
   Utilities/SwiftExtensions.swift
   Utilities/Tree.swift
-  
+
+  Validators/AsyncCompletionsValidator.swift
   Validators/CodingKeyValidator.swift
   Validators/NonsenseFlagsValidator.swift
   Validators/ParsableArgumentsValidation.swift
--- a/Sources/ArgumentParser/Parsable Properties/CompletionKind.swift
+++ b/Sources/ArgumentParser/Parsable Properties/CompletionKind.swift
@@ -41,12 +41,7 @@ public struct CompletionKind {
     case directory
     case shellCommand(String)
     case custom(@Sendable ([String], Int, String) -> [String])
-    #if !canImport(Dispatch)
-    @available(*, unavailable, message: "DispatchSemaphore is unavailable")
     case customAsync(@Sendable ([String], Int, String) async -> [String])
-    #else
-    case customAsync(@Sendable ([String], Int, String) async -> [String])
-    #endif
     case customDeprecated(@Sendable ([String]) -> [String])
   }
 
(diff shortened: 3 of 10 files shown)


rgoldberg:
Are there any changes that you would like for this PR?

natecook1000:
> If the root command is a sync ParsableCommand, a ParserError.invalidState is thrown during runtime when completion candidates are requested from an async custom completion function.

This should be detected at build time, if possible. There are a series of validators in the `Validators` directory that you can add to – they run on every execution in debug mode and effectively act as runtime compilation errors.

rgoldberg:
Thanks. I'll look into Validators soon, then add the check.

rgoldberg:
Is there any way to run some or all of the `ParsableArgumentsValidator`s via build plugins, so they always run at build time (for debug, release, whatever…) instead of requiring subsequent tests on a debug build?

If possible, I'll log an issue.

rgoldberg:
@natecook1000 I've added `AsyncCompletionsValidator` & tests for it.

I've rebased on current main.

Is there any way in the longterm to change SAP to run some or all of the `ParsableArgumentsValidator`s via build plugins, so they always run at build time (for debug, release, whatever…) instead of requiring subsequent tests on a debug build?

If possible, I'll log an issue.

natecook1000:
> Is there any way in the longterm to change SAP to run some or all of the ParsableArgumentsValidators via build plugins, so they always run at build time (for debug, release, whatever…) instead of requiring subsequent tests on a debug build?

I haven't written any build plugins, but from scanning the documentation it doesn't seem like an exact fit. A post-build plugin could be perfect, if such a thing existed? But if I'm wrong and there's a path to this, that would be great!

rgoldberg:
I think the build failures were because I needed to rebase on current main. Just pushed my rebase, so hopefully all checks should succeed now.

I'll look into your review comments, then push updated code; I just wanted to push my rebase first (without having yet addressed your comments) so we could find any other issues that CI might detect.

I've written one build plugin. It generated virtual Swift source before the rest of the code was compiled. I don't know all the options for build plugins; I'll try to take a cursory look, then get back to you, but it might be a bit before I can dedicate time to properly research. I don't know everything that `AsyncCompletionsValidator`s require, etc.

Thanks for reviewing this.

natecook1000 (approved):

rgoldberg:
If I need to do anything to fix the 2 CI failures, please let me know, & I'll fix things & resubmit.

The failures look like Docker failures that seem likely outside the scope of my PR's changes, though.

natecook1000:
Those passed in previous runs – merging to unblock the next bits. Thanks, @rgoldberg! 👏🏻👏🏻👏🏻

rgoldberg:
Thanks for reviewing & merging.

I will try to submit some fish shell improvements soon. I'll try to polish those up in time for 1.8.0.

If Tool Info V1 is merged (see #817), I can update completions to use that instead of V0 to save you guys the work.

ktoso:
FYI this is a breaking change so 1.8.0 source compatibility;

Noticed since swift-java broke. It has from: `1.5` argument parser dependency, and this changes an existing method to prefer an asynchronous version where it did not before -> therefore causing "async method called without await" compile errors.

See here: https://github.com/swiftlang/swift-java/pull/767

marcprux:
> FYI this is a breaking change so 1.8.0 source compatibility;

This broke a couple projects out from under me as well, with errors like:

```
/home/runner/work/skipstone/skipstone/Tests/SkipBuildTests/AndroidEmulatorSetupTests.swift:430:23: error: expression is 'async' but is not marked with 'await'
         ProcessInfo.mockEnvironment = mockEnv.environment()
 
         let command = try AndroidEmulatorCreateCommand.parse([])
                       |   `- note: call is 'async'
                       `- error: expression is 'async' but is not marked with 'await'
```

Might source compatibility be restored by annotating the async variant with [@_disfavoredOverload](https://github.com/swiftlang/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_disfavoredoverload)?

natecook1000:
Apologies for this source break! Tracking the issue here: https://github.com/apple/swift-argument-parser/issues/907

As far as I can tell, `@_disfavoredOverload` doesn't fix things, since the type checker sees the new static methods as shadowing declarations instead of as competing overloads.

Sources/ArgumentParser/Parsable Types/AsyncParsableCommand.swift
```diff
@@ -25,6 +25,34 @@ public protocol AsyncParsableCommand: ParsableCommand {
...
+
+  /// Parses an instance of this type, or one of its subcommands, from
+  /// command-line arguments.
+  ///
+  /// - Parameter arguments: An array of arguments to use for parsing. If
+  ///   `arguments` is `nil`, this uses the program's command-line arguments.
+  /// - Returns: A new instance of this type, one of its subcommands, or a
+  ///   command type internal to the `ArgumentParser` library.
+  /// - Throws: If parsing fails.
+  public static func parseAsRoot(
```
weissi, line 48: This is breaking the public API.

@natecook1000 / @rgoldberg This is failing builds

```
FooBarTests.swift:39:17: error: expression is 'async' but is not marked with 'await'
37 |         for invalid in invalids {
38 |             await XCTAssertThrowsError(
39 |                 try FooBar.parse([
   |                 |   `- note: call is 'async' in an autoclosure argument
   |                 `- error: expression is 'async' but is not marked with 'await'
40 |                     invalid,
41 |                 ]),

```

Issue #849, opened 2026-01-12
☂️ Command Aliases Support


rauhul:

rgoldberg:
#847 should probably be added as a child.

PR #891, opened 2026-04-21
Display command aliases in generated man-pages

### LINKED ISSUE #851: Command aliases should appear in generated man-pages

qflen:
Resolves #851 (and partially addresses the umbrella #849).

Subcommand aliases already flow through `CommandInfoV0.aliases` and surface in `--help` and dump-help output, but the `generate-manual` tool never consumed them.

## Changes

- `Tools/generate-manual/Extensions/ArgumentParser+MDoc.swift`: new `manualPageSubcommandLabel` helper that returns `commandName` alone, or `"commandName, alias1, alias2"` when aliases exist. Mirrors the format used by `HelpGenerator`.
- `Tools/generate-manual/DSL/SinglePageDescription.swift`: use the helper when rendering each subcommand list item.
- `Tools/generate-manual/DSL/Name.swift`: emit one `.Nm` macro per alias in the NAME section. This is the canonical BSD mdoc convention (see `ls(1)`, which lists `ls, dir` in NAME).
- Regenerated the two `testMath*Manual` snapshots. The other ten snapshots are unchanged because the `guard` in `manualPageSubcommandLabel` short-circuits for commands without aliases, so output for alias-less commands is byte-identical.

## Example output

Single-page (Math example, subcommand list):

```
.It Em "multiply, mul"
Print the product of the values.
...
.It Em "average, avg"
Print the average of the values.
```

Multi-page (NAME section on each subcommand page):

```
.Sh NAME
.Nm "math multiply"
.Nm mul
.Nd "Print the product of the values."
```

## Scope decisions

- `SeeAlso.swift`: left alone. It cross-references other man-pages, and aliases do not generate their own pages.
- `Synopsis.swift`: left alone. The bare `.Nm` auto-resolves to the first `.Nm` in the document, which remains the primary name.
- `MultiPageDescription.swift`: does not iterate subcommands, so no change needed.

## Test plan

- [x] `swift test --filter GenerateManualTests` passes (12/12)
- [x] Verified the existing Math example (which already declares `mul` and `avg` aliases via PR #627) now renders them in both single-page and multi-page output
- [x] Confirmed non-aliased commands produce byte-identical output (other 10 snapshots unchanged)

### DIFF
--- a/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathMultiPageManual().mdoc
+++ b/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathMultiPageManual().mdoc
@@ -77,7 +77,8 @@ and
 .Dt MATH.MULTIPLY 9
 .Os
 .Sh NAME
-.Nm "math multiply"
+.Nm "math multiply" ,
+.Nm mul
 .Nd "Print the product of the values."
 .Sh SYNOPSIS
 .Nm
@@ -150,7 +151,8 @@ and
 .Dt MATH.STATS.AVERAGE 9
 .Os
 .Sh NAME
-.Nm "math stats average"
+.Nm "math stats average" ,
+.Nm avg
 .Nd "Print the average of the values."
 .Sh SYNOPSIS
 .Nm
--- a/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathSinglePageManual().mdoc
+++ b/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathSinglePageManual().mdoc
@@ -28,7 +28,7 @@ Show the version.
 .It Fl h , -help
 Show help information.
 .El
-.It Em multiply
+.It Em "multiply, mul"
 Print the product of the values.
 .Bl -tag -width 6n
 .It Fl x , -hex-output
@@ -47,7 +47,7 @@ Calculate descriptive statistics.
 Show the version.
 .It Fl h , -help
 Show help information.
-.It Em average
+.It Em "average, avg"
 Print the average of the values.
 .Bl -tag -width 6n
 .It Fl -kind Ar kind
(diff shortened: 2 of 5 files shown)


rauhul:
Thanks for the contribution, do you have another reference for this:
> Tools/generate-manual/DSL/Name.swift: emit one .Nm macro per alias in the NAME section. This is the canonical BSD mdoc convention (see ls(1), which lists ls, dir in NAME).

`dir` isn't a command on macOS and therefore isn't present in the manual.

qflen:
You're right, `ls(1)` was a bad example on my part. Better evidence, all from macOS:

`grep(1)`:

```
NAME
     grep, egrep, fgrep, rgrep, bzgrep, bzegrep, bzfgrep, zgrep, zegrep,
     zfgrep -- file pattern searcher
```

`test(1)`:

```
NAME
     test, [ -- condition evaluation utility
```

And the convention is documented in `mdoc(7)`:

> `Nm [name]`: The name of the manual page, or (particularly in section 1, 6, and 8 pages) of an additional command or feature documented in the manual page.

That is, in executable sections (1, 6, 8), repeated `Nm` macros are the standard way to list aliased names on one page. If you'd prefer a different format, or would rather keep aliases only in the single-page subcommand list, happy to change it.

Tests/ArgumentParserGenerateManualTests/Snapshots/testMathMultiPageManual().mdoc
```diff
@@ -78,6 +78,7 @@ and
 .Os
 .Sh NAME
 .Nm "math multiply"
+.Nm mul
```
rauhul, line 81: This doesn't render particularly well from what I can tell. When I run `man` I see:
```
NAME
     math multiply mul – Print the product of the values.
```
Can we force a comma after multiply?

qflen: Yeah, done.

PR #892, opened 2026-04-21
Improve diagnostic for @Flag Bool? without inversion:

### LINKED ISSUE #835: Unhelpful compiler error for `@Flag var _: Bool?` without a default value
The following should be supported but isn't:

```
    @Flag(help: "...")
    var writeEmptyFiles: Bool?
```

The error is 

```
Referencing initializer 'init(name:help:)' on 'Flag' requires the types 'Bool?' and 'Int' be equivalent
```

It turns out the `inversion` parameter is required, so the following works:

```
    @Flag(inversion: .prefixedEnableDisable, help: "")
    var writeEmptyFiles: Bool?
```

this seems somewhat accidental? Are we just missing an initializer with defaulted inversion?

**ArgumentParser version:** `0.1.0` or the `main` branch, for example.
**Swift version:** swift-driver version: 1.139 Apple Swift version 6.2 (swiftlang-6.3.0.101.15 clang-2100.0.101.15); arm64-apple-macosx26.0

### Checklist
- [x] Reproduces on main

### Steps to Reproduce
The snippets above are enough to repro

### Expected behavior
We should support just @Flag() on optional boolean fields

### Actual behavior
fails to compile

qflen:
Resolves #835.

## Problem

When a user writes the following, which looks reasonable:

```swift
@Flag(help: "Write empty files.")
var writeEmptyFiles: Bool?
```

the compiler produces a confusing diagnostic:

```
error: referencing initializer 'init(name:help:)' on 'Flag' requires the types 'Bool?' and 'Int' be equivalent
```

The actual cause is that every `Flag where Value == Bool?` init requires an `inversion:` parameter. Without one, overload resolution falls through to the `Flag where Value == Int` counter initializer at `Flag.swift:385`, which is the first `init(name:help:)` the compiler can see, and the `Bool?` vs `Int` mismatch surfaces from there.

## Fix

Add an `@available(*, unavailable)` init overload on `Flag where Value == Bool?` whose signature matches the offending call shape. Overload resolution now picks this one because its `where` clause is more specific, and the custom `message:` points the user at the real cause:

```
error: 'init(name:help:)' is unavailable: An optional 'Bool?' @Flag requires an 'inversion:' parameter, such as '.prefixedNo' or '.prefixedEnableDisable'.
```

This follows the same pattern used for the existing unavailable no-arg `Flag.init()` at `Flag.swift:94-101`, whose doc comment already explains this technique:

> Explicitly marking this initializer unavailable means that when `Value` is a type supported by `Flag` like `Bool` or `EnumerableFlag`, the appropriate overload will be selected instead.

## Test plan

- [x] `swift test` passes (495 tests, no failures)
- [x] Reproduced the original confusing error on `main`
- [x] Verified the new diagnostic message appears for `@Flag(help: "...") var x: Bool?`
- [x] Verified the legitimate usage `@Flag(inversion: .prefixedEnableDisable) var x: Bool?` still compiles and runs correctly

### DIFF
--- a/Sources/ArgumentParser/Parsable Properties/Flag.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Flag.swift
@@ -251,6 +251,24 @@ extension Flag where Value == Bool? {
       exclusivity: exclusivity,
       help: help)
   }
+
+  /// Surfaces a clearer diagnostic when a user declares a `Bool?` flag without
+  /// the required `inversion:` parameter.
+  ///
+  /// Without this overload, the compiler falls back to the
+  /// `Flag where Value == Int` initializer and reports a confusing
+  /// "types 'Bool?' and 'Int' be equivalent" error.
+  @available(
+    *, unavailable,
+    message:
+      "An optional 'Bool?' @Flag requires an 'inversion:' parameter, such as '.prefixedNo' or '.prefixedEnableDisable'."
+  )
+  public init(
+    name: NameSpecification = .long,
+    help: ArgumentHelp? = nil
+  ) {
+    fatalError("unavailable")
+  }
 }
 
 extension Flag where Value == Bool {
--- a/Sources/ArgumentParser/Parsable Properties/Flag.swift
+++ b/Sources/ArgumentParser/Parsable Properties/Flag.swift
@@ -264,7 +264,7 @@ extension Flag where Value == Bool? {
       "An optional 'Bool?' @Flag requires an 'inversion:' parameter, such as '.prefixedNo' or '.prefixedEnableDisable'."
   )
   public init(
-    name: NameSpecification = .long,
+    name _name: NameSpecification = .long,
     help: ArgumentHelp? = nil
   ) {
     fatalError("unavailable")


rauhul (approved):
I want @natecook1000's eyes on this as it affects the public api, albeit with an unusable symbol

Sources/ArgumentParser/Parsable Properties/Flag.swift
```diff
@@ -251,6 +251,24 @@ extension Flag where Value == Bool? {
...
+  /// Without this overload, the compiler falls back to the
+  /// `Flag where Value == Int` initializer and reports a confusing
+  /// "types 'Bool?' and 'Int' be equivalent" error.
+  @available(
+    *, unavailable,
+    message:
+      "An optional 'Bool?' @Flag requires an 'inversion:' parameter, such as '.prefixedNo' or '.prefixedEnableDisable'."
+  )
+  public init(
+    name: NameSpecification = .long,
```
natecook1000: Minor change to suppress this symbol in the docs…
```suggestion
    name _name: NameSpecification = .long,
```

natecook1000 (commented):
Thanks, @qflen! LGTM with one minor change.

PR #894, opened 2026-04-30
Add `extendedDiscussion` to `CommandConfiguration`

### LINKED ISSUE #159: Allow configuration discussion to be moved to end

realrajaryan:
Resolves #159.

Adds an `extendedDiscussion` property to `CommandConfiguration` that renders after the argument, option, and subcommand sections in help output.

This gives command authors a place to put supplementary information i.e. usage examples, notes etc.

### Checklist

### DIFF
--- a/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testMathDoccReference().md
+++ b/Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testMathDoccReference().md
@@ -18,11 +18,6 @@ math [--version] [--help]
 *Show help information.*
 
 
-Examples:
-  math add 10 15 20
-  math multiply --hex-output 16 32
-  math stats average --kind median 5 8 12
-
 ## math.add
 
 Print the sum of the values.
@@ -231,5 +226,10 @@ math help [<subcommands>...] [--version]
 
 
 
+
+Examples:
+  math add 10 15 20
+  math multiply --hex-output 16 32
+  math stats average --kind median 5 8 12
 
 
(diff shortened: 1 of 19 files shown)


rauhul:
Can you add an extended discussion to one of the example tools so we can see how the generated content changes?

realrajaryan:
> Can you add an extended discussion to one of the example tools so we can see how the generated content changes?

added it to the math example

Tests/ArgumentParserGenerateDoccReferenceTests/Snapshots/testMathDoccReference().md
```diff
@@ -18,6 +18,11 @@ math [--version] [--help]
 *Show help information.*
 
 
+Examples:
```
rauhul: does it make sense for this to be below the options for the markdown generator? I think before or after could make sense, wdyt?

realrajaryan: oh ye, I'll move it after the subcommands to match the CLI help layout

PR #895, opened 2026-05-06
[windows] add PDB files to install target

charles-zablit:
This is part of https://github.com/swiftlang/swift/issues/88869.

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -94,6 +94,11 @@ function(_install_target module)
   endif()
 
   install(TARGETS ${module})
+  if(MSVC AND NOT type STREQUAL STATIC_LIBRARY)
+    install(FILES $<TARGET_FILE_DIR:${module}>/$<TARGET_FILE_BASE_NAME:${module}>.pdb
+      DESTINATION ${CMAKE_INSTALL_BINDIR}
+      OPTIONAL)
+  endif()
   if(type STREQUAL EXECUTABLE)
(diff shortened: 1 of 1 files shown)


compnerd (changes requested):
```suggestion
  install(FILES $<TARGET_PDB_FILE:${module}> 
    DESTINATION ${CMAKE_INSTALL_BINDIR}
    OPTIONAL)
```

charles-zablit:
> ```
>   install(FILES $<TARGET_PDB_FILE:${module}> 
>     DESTINATION ${CMAKE_INSTALL_BINDIR}
>     OPTIONAL)
> ```

Is there a specific reason for doing it this way instead of the other way around?

compnerd:
> > ```
> >   install(FILES $<TARGET_PDB_FILE:${module}> 
> >     DESTINATION ${CMAKE_INSTALL_BINDIR}
> >     OPTIONAL)
> > ```
> 
> Is there a specific reason for doing it this way instead of the other way around?

Yes, this makes the code easier to read, this allows CMake to track the dependency properly, and it doesn't require you to know where the PDB is emitted in practice. This is the canonical way that CMake recommends.

charles-zablit:
I tried this and it fails at configure time:
 
```
CMake Error:
  Error evaluating generator expression:

    $<TARGET_PDB_FILE:ArgumentParser>

  TARGET_PDB_FILE is not supported by the target linker.
```

It looks like `TARGET_PDB_FILE` is not available for Swift.

Issue #896, opened 2026-05-08
Failing test: Windows (nightly-main - windows-2022)


natecook1000:
The `Windows (nightly-main)` CI job has started failing with a "file with the same name already exists" error (see below). I'm going to disable the job for now to continue merging PRs.

```
[84/100] ArgumentParserEndToEndTests-product 392 / 463
[84/100] roll-product 394 / 468
error: Error Domain=NSCocoaErrorDomain Code=516 "A file with the same name already exists."UserInfo={NSFilePath=C:\source\.build\out\Intermediates.noindex\swift-argument-parser.build\Debug-windows\repeat-23CB7832C35147-testable-t.build\Objects-normal\x86_64\Modules\repeat.o, NSUserStringVariant=["Copy"], NSURL=file:///C:/source/.build/out/Intermediates.noindex/swift-argument-parser.build/Debug-windows/repeat-23CB7832C35147-testable-t.build/Objects-normal/x86_64/Modules/repeat.o, NSUnderlyingError=Win32Error(code: 80), NSDestinationFilePath=C:\source\.build\out\Products\Debug-windows\ExecutableModules\repeat-23CB7832C35147-testable.objlib\repeat.o, NSSourceFilePathErrorKey=C:\source\.build\out\Intermediates.noindex\swift-argument-parser.build\Debug-windows\repeat-23CB7832C35147-testable-t.build\Objects-normal\x86_64\Modules\repeat.o}
error: AssembleObjectLibrary C:\\source\\.build\\out\\Products\\Debug-windows\\ExecutableModules\\repeat-23CB7832C35147-testable.objlib normal failed with a nonzero exit code. Command line:     cd C:\source
    builtin-ObjectLibraryAssembler --linker-response-file-format windowsShellQuotedNewlineSeparated C:\source\.build\out\Intermediates.noindex\swift-argument-parser.build\Debug-windows\repeat-23CB7832C35147-testable-t.build\Objects-normal\x86_64\Repeat.o C:\source\.build\out\Intermediates.noindex\swift-argument-parser.build\Debug-windows\repeat-23CB7832C35147-testable-t.build\Objects-normal\x86_64\Modules\repeat.o --output C:\source\.build\out\Products\Debug-windows\ExecutableModules\repeat-23CB7832C35147-testable.objlib
[...]

PR #897, opened 2026-05-08
Disable the nightly-main Windows test job

natecook1000:
Failing for reasons unknown; tracked in #896.


dannytariqu10-hash (commented):
TCS

PR #899, opened 2026-05-18
Revert "Disable the nightly-main Windows test job"

owenv:
Reverts apple/swift-argument-parser#897


owenv:
windows snapshots are lagging behind macOS/Linux, so this is too early

PR #902, opened 2026-05-23
Fix: Use `visibility.base` consistently in `generateHelpNames` for `.longWithSingleDash`

### LINKED ISSUE #901: Bug: Inconsistent visibility string interpolation in `generateHelpNames` for `.longWithSingleDash` vs `.long` names
### Description

In `Sources/ArgumentParser/Usage/HelpGenerator.swift`, the `generateHelpNames(visibility:)` method on `NameSpecification` handles the `.long` and `.longWithSingleDash` cases inconsistently when generating non-default visibility help flag names.

**Affected file:** `Sources/ArgumentParser/Usage/HelpGenerator.swift`

### The Inconsistency

```swift
fileprivate func generateHelpNames(visibility: ArgumentVisibility) -> [Name] {
    self
      .makeNames(InputKey(name: "help", parent: nil))
      .compactMap { name in
        guard visibility.base != .default else { return name }
        switch name {
        case .long(let helpName):
          return .long("\(helpName)-\(visibility.base)")      // ✅ uses visibility.base
        case .longWithSingleDash(let helpName):
          return .longWithSingleDash("\(helpName)-\(visibility)") // ❌ uses visibility (full), not visibility.base
        case .short:
          return nil
        }
      }
      .sorted(by: >)
}
```

- The `.long` case correctly appends `visibility.base` (e.g., `"hidden"`)
- The `.longWithSingleDash` case appends the full `visibility` description instead of `visibility.base`

This means if a command uses single-dash long names (e.g., `-help`), the generated hidden/private help flag name would be different from what the double-dash variant generates, causing **inconsistent CLI behavior** where `-help-hidden` does not behave the same as `--help-hidden`.

### Steps to Reproduce

1. Create a command that uses `.longWithSingleDash` in its `helpNames` configuration
2. Call the command with a non-default visibility help flag
3. Observe the appended suffix differs from the `.long` variant

### Expected Behavior

Both `.long` and `.longWithSingleDash` cases should use `visibility.base` for consistent flag name generation:

```swift
case .longWithSingleDash(let helpName):
  return .longWithSingleDash("\(helpName)-\(visibility.base)") // ✅ fix: use .base
```

### Environment

- **swift-argument-parser version:** main branch (commit `af51a49`)
- **Swift version:** 5.9+
- **Platform:** All platforms

### Impact

Low-to-medium. Affects only commands that configure `helpNames` with `.longWithSingleDash` at non-default visibility levels, but the inconsistency is a latent bug that could surface in toolchains and developer tools built on top of this library.

---

*Reported after manual code review of `HelpGenerator.swift`.*

AkshatRaj00:
## Summary

This PR fixes an inconsistency in `generateHelpNames(visibility:)` in `HelpGenerator.swift` where the `.longWithSingleDash` case used the full `visibility` description instead of `visibility.base`, while the `.long` case correctly used `visibility.base`.

**Fixes #901**

## The Bug

```swift
// BEFORE (inconsistent)
case .long(let helpName):
  return .long("\(helpName)-\(visibility.base)")           // ✅ uses .base
case .longWithSingleDash(let helpName):
  return .longWithSingleDash("\(helpName)-\(visibility)")  // ❌ uses full visibility
```

## The Fix

```swift
// AFTER (consistent)
case .long(let helpName):
  return .long("\(helpName)-\(visibility.base)")
case .longWithSingleDash(let helpName):
  return .longWithSingleDash("\(helpName)-\(visibility.base)")  // ✅ now matches .long
```

## Impact

- **One-line change** with no API breakage
- Affects only commands that configure `helpNames` with `.longWithSingleDash` at non-default (`hidden`/`private`) visibility levels
- Ensures `-help-hidden` and `--help-hidden` use the same suffix generation logic

## Testing

Existing tests for help generation should continue to pass. The affected code path is only triggered when:
1. `helpNames` includes `.longWithSingleDash`
2. `visibility.base != .default` (i.e., `.hidden` or `.private`)

No new tests are broken by this change.

### DIFF
--- a/Sources/ArgumentParser/Usage/HelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/HelpGenerator.swift
@@ -444,7 +444,7 @@ extension NameSpecification {
         case .long(let helpName):
           return .long("\(helpName)-\(visibility.base)")
         case .longWithSingleDash(let helpName):
-          return .longWithSingleDash("\(helpName)-\(visibility)")
+          return .longWithSingleDash("\(helpName)-\(visibility.base)")  // Fix: use .base for consistency with .long case
         case .short:
           // Cannot create a non-default help flag from a short name.
           return nil


Sources/ArgumentParser/Usage/HelpGenerator.swift
```diff
@@ -450,7 +444,7 @@ extension NameSpecification {
         case .long(let helpName):
           return .long("\(helpName)-\(visibility.base)")
         case .longWithSingleDash(let helpName):
-          return .longWithSingleDash("\(helpName)-\(visibility)")
+          return .longWithSingleDash("\(helpName)-\(visibility.base)")  // Fix: use .base for consistency with .long case
```
natecook1000, line 447: Let's remove the comment here – it makes sense in the context of this PR, but not when reading the code by itself.
```suggestion
          return .longWithSingleDash("\(helpName)-\(visibility.base)")
```

PR #908, opened 2026-05-26
Revert source break in 1.8.0 `parse` methods

### LINKED ISSUE #907: 1.8.0 breaks source for `AsyncParsableCommand`s that manually invoke `parse...` method
As [noted here](https://github.com/apple/swift-argument-parser/pull/855#issuecomment-4539704450), the change in #855 was not actually source compatible, since overload resolution doesn't distinguish methods by whether they are `async` or not. Code in an `AsyncParsableCommand` extension that previously called `ParsableCommand.parse` or `parseAsRoot` now calls the new `async` versions, causing a compilation failure:

```swift
         let command = try MyCommand.parseAsRoot()
//                     |   `- note: call is 'async'
//                     `- error: expression is 'async' but is not marked with 'await'
```

The fix for clients is to add the `await` keyword; working on a source compatible fix for the issue.

natecook1000:
This removes the `async` overloads to `parse` and `parseAsRoot` that caused a source break in 1.8.0, and introduces new names that disambiguate the methods with `async` behavior. Instead of calling the existing methods with the `await` keyword, users can call `await asyncParse` or `await asyncParseAsRoot`.

This change also adds a validation step to the synchronous parsing path to check if an async completions function is called.

Users that have already updated their code to work around the source break will revert to the old methods with this change, which will yield a "No 'async' operations occur within 'await' expression" warning. They can remove the `await` keyword or update to call the async versions directly to fix the warning.

Resolves #907

### DIFF
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -303,6 +303,23 @@ extension CommandParser {
   mutating func parse(
     arguments: [String]
   ) throws(CommandError) -> ParsableCommand {
+    #if DEBUG
+    if self.rootCommand is AsyncParsableCommand.Type,
+      let error =
+        AsyncCompletionsValidator
+        .validate(self.rootCommand, parent: nil, forcedSyncParse: true)
+    {
+      configurationFailure(
+        """
+        This command uses asynchronous custom completion functions,
+        but is invoked using a synchronous call to `parse` or 
+        `parseAsRoot`. To fix this issue, call the asynchronous
+        versions of these functions: `asyncParse` or `asyncParseAsRoot`,
+        respectively.
+        """)
+    }
+    #endif
+
     do {
       if let (argument, arguments) =
         try parseCustomCompletion(from: arguments)
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -304,7 +304,8 @@ extension CommandParser {
     arguments: [String]
   ) throws(CommandError) -> ParsableCommand {
     #if DEBUG
-    if self.rootCommand is AsyncParsableCommand.Type,
+    if #available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *),
+      self.rootCommand is AsyncParsableCommand.Type,
       let error =
         AsyncCompletionsValidator
         .validate(self.rootCommand, parent: nil, forcedSyncParse: true)
--- a/Sources/ArgumentParser/Parsing/CommandParser.swift
+++ b/Sources/ArgumentParser/Parsing/CommandParser.swift
@@ -306,17 +306,19 @@ extension CommandParser {
     #if DEBUG
     if #available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *),
       self.rootCommand is AsyncParsableCommand.Type,
-      let error =
-        AsyncCompletionsValidator
-        .validate(self.rootCommand, parent: nil, forcedSyncParse: true)
+      self.rootCommand.asyncCompletions(
+        parent: nil,
+        propertyPath: String(describing: self.rootCommand)
+      ).isEmpty == false
     {
       configurationFailure(
         """
         This command uses asynchronous custom completion functions,
-        but is invoked using a synchronous call to `parse` or 
-        `parseAsRoot`. To fix this issue, call the asynchronous
-        versions of these functions: `asyncParse` or `asyncParseAsRoot`,
-        respectively.
+        but is invoked using a synchronous call to `parse` or
+        `parseAsRoot`. 
+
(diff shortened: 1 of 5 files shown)


Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -303,6 +303,24 @@ extension CommandParser {
...
+        AsyncCompletionsValidator
+        .validate(self.rootCommand, parent: nil, forcedSyncParse: true)
+    {
+      configurationFailure(
+        """
+        This command uses asynchronous custom completion functions,
+        but is invoked using a synchronous call to `parse` or 
+        `parseAsRoot`. To fix this issue, call the asynchronous
+        versions of these functions: `asyncParse` or `asyncParseAsRoot`,
+        respectively.
```
rgoldberg:
```suggestion
        but is invoked using a synchronous call to `parse` or
        `parseAsRoot`.

        To fix this issue, call the asynchronous versions of these
        functions: `asyncParse` or `asyncParseAsRoot`, respectively.
```

I split the text into separate diagnostic & prescription paragraphs for (IMO) better readability.

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -303,6 +303,24 @@ extension CommandParser {
   mutating func parse(
     arguments: [String]
   ) throws(CommandError) -> ParsableCommand {
+    #if DEBUG
+    if #available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *),
+      self.rootCommand is AsyncParsableCommand.Type,
+      let error =
+        AsyncCompletionsValidator
+        .validate(self.rootCommand, parent: nil, forcedSyncParse: true)
```
rgoldberg:
```suggestion
      !self.rootCommand.invalidAsyncCompletions(
        parent: nil,
        propertyPath: .init(describing: self.rootCommand)
      )
      .isEmpty
```

The `validate` overload is unnecessary.

Sources/ArgumentParser/Validators/AsyncCompletionsValidator.swift
```diff
@@ -35,11 +35,23 @@ struct AsyncCompletionsValidator: ParsableArgumentsValidator {
...
+  {
+    validate(type, parent: parent, forcedSyncParse: false)
+  }
+
+  static func validate(
+    _ type: ParsableArguments.Type,
+    parent: InputKey?,
+    forcedSyncParse: Bool
+  )
+    -> ParsableArgumentsValidatorError?
```
rgoldberg:
```suggestion
```

The `validate` overload is unnecessary.

Sources/ArgumentParser/Validators/AsyncCompletionsValidator.swift
```diff
@@ -35,11 +35,23 @@ struct AsyncCompletionsValidator: ParsableArgumentsValidator {
...
   {
     guard
       type is ParsableCommand.Type,
-      !(type is AsyncParsableCommand.Type)
-    else { return nil }
+      !(type is AsyncParsableCommand.Type) || forcedSyncParse
+    else {
+      return nil
+    }
```
rgoldberg:
```suggestion
      !(type is AsyncParsableCommand.Type)
    else { return nil }
```

The `validate` overload is unnecessary.

This & the other suggestion together just revert this file back to pre-PR version.

rgoldberg (changes requested):
Using `#if DEBUG` to call the validator in the sync `parse(_:)` is a great idea.

I just suggest:

1. Splitting the error text into 2 separate paragraphs to avoid "wall of text".
2. Removing the `validate` overload, since it's unnecessary.

Sources/ArgumentParser/Validators/AsyncCompletionsValidator.swift
natecook1000: I don't _think_ it's unnecessary. This validator now gets called two ways:

1. The validator process calls it. In that case, we want to verify that a root type that only conforms to `ParsableCommand` doesn't have any `async` custom completion handlers declared, so we exit early if `type` conforms to `AsyncParsableCommand`.
2. The synchronous `parse` method calls it. Here we are checking for an `AsyncParsableCommand` (which would have passed validation) that is being parsed synchronously. So we want to perform the "validation" step (aka find out if there are any `async` custom completion handlers) even though the type (correctly) conforms to `AsyncParsableCommand`.

So we need a way to say "do this no matter what" in the validator.

rgoldberg: See the suggestion in my [earlier comment](https://github.com/apple/swift-argument-parser/pull/908#discussion_r3306892662).

Calling `invalidAsyncCompletions(…)` directly achieves "do this no matter what", because it avoids the `guard` in `validate(…)`.

natecook1000: Ha, right on! That's what I ended up doing in my change, just didn't parse that you'd already shown it upthread. 👍🏻

Issue #41, opened 2020-03-02
Add support for params files


thomasvl:
Some tools/situations end up needing so many arguments that they can run in to the max command line length. This is usually handled via _argument files_ or _parameter files_ -- if an argument is prefixed with a marker (usually `@`), then the contents of the referenced file are read at that location in the command line as if the content were on the command line.  This also usually works recursively in that the read file could include another one of these directives to read yet another file.

There can be two types of "modes" that are supported for reading these files:
- _multiline_ : Every line of the files is a single argument.  This is the easy one, as it means spaces, etc. are part of the arg, there is no need to understand escaping, just split on newlines and handle things.
- _shell quoted_ : The files are parse like a Bourne shell would, this requires handing quotes and escapes.  `swift` and `clang` support files in this form.

Reference:
- Python ArgParse support: https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars
- Bazel docs with notes about generating files for these two modes: https://docs.bazel.build/versions/2.0.0/skylark/lib/Args.html#set_param_file_format

mattt:
This is something we've been thinking about for [`swift-doc`](https://github.com/SwiftDocOrg/swift-doc/issues/74), and it'd be great to have consensus around how this should work.

@thomasvl Thanks for sharing those references for parameter files. Coming from more of a Ruby JavaScript background, I'm less familiar with this approach; in those languages, the more common pattern seems to be configuration files encoded as JSON / YAML / TOML.

I'm curious to hear whether we see these two approaches (parameter file and configuration file) as alternatives or compliments to one another. Is it possible or indeed desirable to support both at once?

thomasvl:
My $0.02 is that params files are more meant as a way around OS arg limits and generally dealing with lots of args (especially file lists), so they reuse the main argument parsing logic, just using a (hopefully simple) tokenizer out of the file to supplement what comes on `argv`.

To me, configuration files tend to be a full replacement for the the command line flags. Instead of just a list of flags, it is a more structured form of all the args, usually allowing slightly better grammar since you scan do sub objects, etc. instead of making really long arg names.  So while the command line might have `--foobar=1` and `--foobaz=2` when you go to a config file, it could have something like `foo: { bar: 1, baz: 2 }`; or `--mumble 1 --mumble 2` it can be: `mumble: [1, 2]`.

I can't say I've seen something automatically bridge args and config files; but it could be possible.

natecook1000:
[...]

PR #909, opened 2026-05-27
Add native response file support

### LINKED ISSUE #846: Add native response file support
As a CLI developer, I want Swift Argument Parser to natively support response file so I can make use of them in my CLI.

https://bh0.phas.ubc.ca/C2/intel/docs/c_ug/linux209a.htm

bkhouri:
Add native response file support in Swift Argument Parser for all `ParsableCommand` and `AsyncParsableCommand` commands. When at least one `@file` reference is present in the input, parser error messages now include a multi-line `at <file>:<line>` / `included from <file>:<line>` block that pinpoints the offending argument and its full include chain.

Fixes: #846
Relates to #41 

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
+++ b/Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
@@ -177,14 +177,14 @@ tokens):
 
 #### Common mistakes
 
-- **Unclosed quotes** are implicitly closed at end-of-line (matching the
-  gcc/LLVM tokenizer), so the token that follows the opening quote
-  absorbs the rest of the line verbatim. Watch out for this: a stray `"`
-  or `'` in the middle of a line will silently gobble everything up to
-  the next newline into one argument.
+- **Unclosed quotes** are implicitly closed at end-of-file: whatever
+  content follows the opening quote — including any intervening
+  newlines — becomes part of a single token. Watch out for this: a
+  stray `"` or `'` will silently gobble everything up to the end of the
+  file (or the matching quote) into one argument.
 - **Mismatched quotes** (opening with `"` and closing with `'`, or vice
   versa) behave the same way — the second quote is treated as literal
-  content, and the token terminates at end-of-line.
+  content inside the still-open segment.
 - **Escaping in the wrong style** — writing `'a\nb'` produces the literal
   four characters `a\nb`, not `a<newline>b`. Use double quotes if you want
   the newline.
--- a/Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
+++ b/Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
@@ -1,0 +1,359 @@
+# Response Files
+
+Use response files to store command-line arguments in files for reuse and to handle commands with many arguments.
+
+## Overview
+
+Response files allow you to store command-line arguments in text files and reference them using the `@filename` syntax. This feature is particularly useful for:
+
+- Commands with many arguments that would be unwieldy on the command line
+- Reusable argument sets for different configurations
+- Build systems and automation scripts
+- Avoiding command-line length limitations
+
+Swift Argument Parser response file support is opt-in per command. To
+enable it, override the `responseFilePrefix` static property on your
+root `ParsableCommand` (or `AsyncParsableCommand`) with the character
+that should introduce a response-file reference. The default value is
+`nil`, meaning response files are disabled and any argument that would
+otherwise look like `@file` is passed through as a literal value.
+
+## Basic Usage
+
+Enable response files by declaring a prefix on the root command:
+
+```swift
+struct MyTool: ParsableCommand {
+    static var responseFilePrefix: Character? { "@" }
+    // ...
+}
+```
+
+Then reference a response file on the command line by prefixing the
+filename with the character you chose (`@` in this example):
+
+```bash
+mycommand @args.txt --verbose
+```
+
+This expands the contents of `args.txt` and includes them in the command-line arguments.
+
+### Customizing the Response File Prefix
+
+The `@` character is the most widely used convention (for example,
+`clang @args.txt`), but any single character works. If `@` conflicts
+with your command's own argument syntax, pick a different character
+instead. The root command's prefix applies to the entire invocation,
+including subcommand parsing.
+
+```swift
+struct MyTool: ParsableCommand {
+    static var responseFilePrefix: Character? { "+" }
+    // ...
+}
+```
+
+With this configuration, `mytool +args.txt` expands `args.txt` as a
+response file, while `mytool @foo` is passed through as a literal
+argument. The same doubling escape convention applies to the customized
+prefix — `++value` inside a response file produces the literal `+value`,
+mirroring the, implied response file prefix, `@@` → `@` behavior
+described in [Literal At Signs](#literal-at-signs).
+
+### Response File Format
+
+Response files support multiple formats:
+
+#### One Argument Per Line
+```
+--input
+input.txt
+--output
+output.txt
+--verbose
+```
+
+#### Space-Separated Arguments
+```
+--input input.txt --output output.txt --verbose
+```
+
+#### Mixed Format
+```
+--input input.txt
+--output output.txt
+--verbose --count 42
+```
+
+## Advanced Features
+
+### Quoted Arguments
+
+Response files support quoted arguments for values containing spaces, special
+characters, or characters that would otherwise be interpreted as part of the
+response-file syntax (like `#` starting a comment).
+
+```
+--input "file with spaces.txt"
+--message 'Hello, world!'
+--path "/Users/username/My Documents/file.txt"
+```
+
+#### Which quote style to use
+
+Both double quotes (`"..."`) and single quotes (`'...'`) are supported, and
+they behave differently:
+
+- **Double quotes (`"..."`)** — the *interpreting* quote. Escape sequences
+  inside the quotes are processed (see below), and the quote character can
+  itself be escaped with `\"`.
+- **Single quotes (`'...'`)** — the *literal* quote. Everything between the
+  quotes is taken verbatim; there is no escape processing. Use single quotes
+  when the value contains backslashes or double quotes that you want to
+  preserve exactly as written.
+- **Nested quotes** — a double-quoted string may contain unescaped single
+  quotes, and a single-quoted string may contain unescaped double quotes.
+  Only the outer quote style needs to be escaped or switched.
+
+Examples:
+
+```
+# Double-quoted: backslash escapes are processed.
+--message "line one\nline two"        # decoded as "line one<LF>line two"
+--path    "C:\\Users\\Bob\\file.txt"  # decoded as "C:\Users\Bob\file.txt"
+--quote   "she said \"hi\""           # decoded as 'she said "hi"'
+
+# Single-quoted: nothing is processed. Use this when you want backslashes
+# to reach the tool verbatim.
+--regex   'foo\d+\.bar'               # tool receives 'foo\d+\.bar' unchanged
+--json    '{"key": "value"}'          # tool receives '{"key": "value"}'
+
+# Mixing quote styles avoids escaping.
+--sql     "SELECT * FROM 'users'"
+--label   'the "special" one'
+```
+
+#### Recognized escape sequences (double quotes only)
+
+Inside double-quoted arguments, the following backslash escapes are
+recognized:
+
+| Escape | Decoded |
+|--------|---------|
+| `\n`   | newline |
+| `\t`   | tab     |
+| `\r`   | carriage return |
+| `\\`   | a single `\` |
+| `\"`   | a single `"` |
+
+Any other `\x` sequence (for an unrecognized `x`) is preserved verbatim as
+`\x`, so accidental escape characters don't silently disappear. Backslashes
+inside single-quoted arguments are always literal.
+
+#### Whitespace handling
+
+Whitespace inside a quoted value (of either style) is preserved and treated
+as part of the value — it does not split the argument. Outside of quotes,
+runs of spaces and tabs separate arguments as usual.
+
+```
+# Both --title and --tags receive a single value each.
+--title "Grand Total"
+--tags  '  keep   the   spaces  '
+```
+
+#### `@file` inside a quoted argument
+
+The response-file prefix is only interpreted when it starts an *unquoted*
+token. A quoted string that happens to begin with `@` is passed through as a
+literal value (this differs from the `@@`-escape mechanism described in
+[Literal At Signs](#literal-at-signs), which only applies to unquoted
+tokens):
+
+```
+--username "@admin"    # value is "@admin"; no file is opened
+--pattern  '@daily'    # value is "@daily"; single quotes protect it too
+```
+
+#### Common mistakes
+
+- **Unclosed quotes** are implicitly closed at end-of-line (matching the
+  gcc/LLVM tokenizer), so the token that follows the opening quote
+  absorbs the rest of the line verbatim. Watch out for this: a stray `"`
+  or `'` in the middle of a line will silently gobble everything up to
+  the next newline into one argument.
+- **Mismatched quotes** (opening with `"` and closing with `'`, or vice
+  versa) behave the same way — the second quote is treated as literal
+  content, and the token terminates at end-of-line.
+- **Escaping in the wrong style** — writing `'a\nb'` produces the literal
+  four characters `a\nb`, not `a<newline>b`. Use double quotes if you want
+  the newline.
+
+### Comments
+
+Use `#` to add comments to response files:
+
+```
+# Configuration for production build
+--release
+--optimize-for-size
+--output production.app  # Output directory
+```
+
+### Equals Format
+
+Arguments can use the equals format:
+
+```
+--input=input.txt
+--output=output.txt
+--count=42
+```
+
+### Nested Response Files
+
+Response files can reference other response files:
+
+```
+# main.txt
+@common-options.txt
+--specific-flag
+@environment-config.txt
+```
+
+```
+# common-options.txt
+--verbose
+--log-level debug
+```
+
+```
+# environment-config.txt
+--api-endpoint https://api.example.com
+--timeout 30
+```
+
+## Error Handling
+
+Swift Argument Parser provides clear error messages for response file issues:
+
+### File Not Found
+```
+Error: Response file not found: args.txt
+```
+
+### Malformed Content
+```
+Error: Malformed content in response file 'args.txt': Line 3: Unclosed quote in argument
+```
+
+### Recursive Inclusion
+```
+Error: Recursive response file inclusion detected: /path/to/recursive.txt
+```
+
+### Maximum Nesting Depth
+```
+Error: Maximum nesting depth (32) exceeded for response files
+```
+
+## Examples
+
+### Build Configuration
+
+Create different response files for different build configurations:
+
+```bash
+# debug.txt
+--configuration debug
+--enable-assertions
+--debug-info-format dwarf
+--output debug-build
+
+# release.txt
+--configuration release
+--optimize-for-speed
+--strip-debug-symbols
+--output release-build
+```
+
+Use them with your build command:
+
+```bash
+myBuild @debug.txt
+myBuild build @release.txt
+```
+
+### Testing with Multiple Inputs
+
+```bash
+# test-inputs.txt
+--input test1.swift
+--input test2.swift
+--input test3.swift
+--output-dir test-results
+--format junit
+--parallel
+```
+
+```bash
+myTest @test-inputs.txt
+```
+
+### Database Migration
+
+```bash
+# migration.txt
+--database-url postgresql://localhost/myapp
+--migration-dir migrations/
+--environment production
+--backup-before-migration
+--verbose
+```
+
+```bash
+migrate @migration.txt
+```
+
+## Mixing Response Files with Command-Line Arguments
+
+You can combine response files with regular command-line arguments:
+
+```bash
+mycommand @base-config.txt --override-setting value @additional-flags.txt
+```
+
+Arguments are processed in order, so later arguments can override earlier ones:
+
+```bash
+# If base-config.txt contains "--count 10"
+mycommand @base-config.txt --count 20  # Final count will be 20
+```
+
+## Literal At Signs
+
+To use a literal `@` character in an argument, escape it with `@@`:
+
+```bash
+mycommand --username @@admin  # Passes "@admin" as the username
+```
+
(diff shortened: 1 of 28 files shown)


Tests/ArgumentParserEndToEndTests/ResponseFileEndToEndTests.swift
```diff
@@ -0,0 +1,686 @@
...
+        --name
+        @@notafile
+        """)
+
+    AssertParse(SimpleCommand.self, ["@\(responseFile)"]) { command in
+      XCTAssertEqual(command.name, "@notafile")
+    }
+  }
+
+  func testResponseFileWithTerminator() throws {
```
natecook1000: Let's have a test with mixed arguments, with a response file with a terminator first – does it still parse the inline arguments after the response file or treat them as post-terminator args?

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
```diff
@@ -0,0 +1,226 @@
...
+--input input.txt
+--output output.txt
+--verbose --count 42
+```
+
+## Advanced Features
+
+### Quoted Arguments
+
+Response files support quoted arguments for values containing spaces:
```
cmcgee1024: question: quoting brings up the question of how escaping works for things like spaces in unquoted text, and escaping the quotes themselves. If you intend on using shell conventions, then what about single quotes, or escaping scapes with `\`?

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+--input=input.txt
+--output=output.txt
+--count=42
+```
+
+### Nested Response Files
+
+Response files can reference other response files:
```
cmcgee1024, line 216: suggestion: Maybe don't add nesting unless there's a compelling workflow to do so. By having response files it alleviates the problem of command lines that are too long for the OS, improves versioning, and clarity. I'm not sure if nesting adds much beyond that, and brings more levels of indirection for the user to figure out where an argument came from.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
```diff
@@ -0,0 +1,226 @@
...
+Arguments are processed in order, so later arguments can override earlier ones:
+
+```bash
+# If base-config.txt contains "--count 10"
+mycommand @base-config.txt --count 20  # Final count will be 20
+```
+
+## Literal At Signs
+
+To use a literal `@` character in an argument, escape it with `@@`:
```
cmcgee1024, line 336: praise: it is good to call this kind of escaping out like this.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
```diff
@@ -0,0 +1,226 @@
...
+
+```bash
+mycommand --username @@admin  # Passes "@admin" as the username
+```
+
+In response files:
+
+```
+--message @@important: This is critical
```
cmcgee1024, line 345: thought: if nesting isn't required then this kind of escaping becomes unnecessary in the response files themselves.

rauhul:
IMO this sort of feature will need to come with some sort of source location feature too

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
allevato: LLVM's response file parsing supports nesting. I know that swift-driver _today_ uses its own parsing instead of swift-argument-parser, but if there's any desire in the future to move existing LLVM-based tools from C++ onto Swift using swift-argument-parser, we'd need nested response file support to keep parity.

IMO, we should support it unless there's a compelling reason *not* to, to avoid unnecessary limitations on what the library can do. It looks like it falls out naturally from the implementation.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
thomasvl: https://github.com/apple/swift-argument-parser/issues/41 has some links to how this is supported by some other languages/packages - some of them let one control the support the files will support.

Since someone could be using this support to migrate a tool to Swift, they might need the quoted form while some other migration might need the shell escaping support.

Likewise for what the marker on the command line is to say a file is a params/response file. So it if the leading `@` maybe should be configurable.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
bkhouri: The change currently supports nested response file.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
bkhouri: we are supporting nested response files.  See [this conversation](https://github.com/apple/swift-argument-parser/pull/909#discussion_r3318935465).

bkhouri:
> IMO this sort of feature will need to come with some sort of source location feature too

Thanks for this feedback.  A source location feature has been added.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
bkhouri: The documents has been updated to provide additional information on the quotes arguments.

bkhouri:
marking as `Draft` to prevent accidentally merging before it's ready.

Sources/ArgumentParser/Documentation.docc/Articles/ExperimentalFeatures.md
```diff
@@ -13,6 +13,7 @@ If you have any feedback on experimental features, please [open a GitHub issue][
 | Name | Description | related PRs | Version |
 | ------------- | ------------- | ------------- | ------------- |
 | `--experimental-dump-help`  | Dumps command/argument/help information as JSON | [#310][] [#335][] | 0.5.0 or newer |
+| `--experimental-dump-arguments-source-location`  | Dumps the parsed argument tree with each value's source location (file:line for response-file args, `argv[N]` for command-line args). Accepts `=text` (default) or `=json`. | — | 1.9.0 or newer |
```
bkhouri: chore: need to update the version where this feature is available. 

@natecook1000: could you please confirm what version I should set this to? Is there a "Next" equivalent that is used in this project?

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
```diff
@@ -0,0 +1,317 @@
...
+Arguments are processed in order, so later arguments can override earlier ones:
+
+```bash
+# If base-config.txt contains "--count 10"
+mycommand @base-config.txt --count 20  # Final count will be 20
+```
+
+## Literal At Signs
+
+To use a literal `@` character in an argument, escape it with `@@`:
```
bkhouri, line 336: Chore: Add a reference back to the quoted strings as another way to add a literal `@`.

Sources/ArgumentParser/Documentation.docc/Articles/ExperimentalFeatures.md
bkhouri: chore (blocking): Add the related PR.

Add `[#909](https://github.com/apple/swift-argument-parser/pull/909)` as a related PR.

bkhouri: I updated the version to `unreleased`

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
bkhouri: done.

Sources/ArgumentParser/Parsing/SplitArguments.swift
```diff
@@ -684,14 +731,48 @@ func parseIndividualArg(_ arg: String, at position: Int) throws(ParserError)
...
+  ///     - arguments: The input from the command line.
+  ///     - responseFilePrefix: A leading character representing a response
+  ///       file argument
   ///
   /// - Throws: If parsing fails.
-  init(arguments: [String]) throws {
-    self.init(originalInput: arguments)
+  init(
+    arguments: [String],
+    responseFilePrefix: Character = "@"
```
rauhul: please no default set params (unless public api)

bkhouri:
this is a rather large PR. I will look to split it so the "source location" is a parent change to this.

bkhouri:
https://github.com/apple/swift-argument-parser/pull/931 augments the source location information.

Sources/ArgumentParser/Documentation.docc/Articles/ResponseFiles.md
bkhouri: Thanks.

Sources/ArgumentParser/Parsing/SplitArguments.swift
bkhouri: ACK.. I'll update this to make the response file prefix configurable, with `"@"` being the default.

allevato:
Apologies if I missed it, but can you add a test for quoted empty string behavior in response files?

For example, these files:

```
-foo
''
-bar
```
and
```
-foo
""
-bar
```
should both expand to three arguments: `["-foo", "", "-bar"]`.

This was on my mind because a few months ago I fixed a bug in LLVM's response file parser that was incorrectly dropping empty arguments (instead of preserving them as `gcc` would): https://github.com/llvm/llvm-project/pull/187566

In fact, some of the other test cases I added might also serve as good tests of quotation handling for this implementation, too.

bkhouri:
> Apologies if I missed it, but can you add a test for quoted empty string behavior in response files?

Thanks for the feedback @allevato .  I added more test coverage around empty quoted string s, and some with a new line character (`\n`) in a quoted string.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -22,6 +22,30 @@ public protocol ParsableCommand: ParsableArguments {
...
+  ///
+  ///     struct MyTool: ParsableCommand {
+  ///       static var responseFilePrefix: Character? { "+" }
+  ///       // ...
+  ///     }
+  ///
+  /// With the above configuration, `mytool +args.txt` expands `args.txt`
+  /// as a response file, while `mytool @foo` is treated as a literal
+  /// argument.
+  static var responseFilePrefix: Character? { get }
```
natecook1000: Does this need to be in the `ParsableCommand` protocol or could it be moved to `CommandConfiguration`? That seems like its more natural home.

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -12,6 +12,11 @@
 struct CommandError: Error {
   var commandStack: [ParsableCommand.Type]
   var parserError: ParserError
```
natecook1000, line 21: Could you document these as well?

Sources/ArgumentParser/Parsing/CommandParser.swift
```diff
@@ -87,7 +96,7 @@ extension CommandParser {
   ) -> Tree<ParsableCommand.Type>? {
     guard let (origin, element) = split.peekNext(),
       element.isValue,
```
natecook1000:
```suggestion
```
The changed line below makes this one unnecessary…

Sources/ArgumentParser/Parsing/InputOrigin.swift
```diff
@@ -113,15 +122,131 @@ extension InputOrigin {
...
+  }
+}
+
 extension InputOrigin.Element {
   static func < (lhs: Self, rhs: Self) -> Bool {
+    // Ordering (least → greatest):
+    //   .argumentIndex(a) < .argumentIndex(b) when a < b
+    //   .argumentIndex(_) < .responseFile(_,_) < .defaultValue
+    // Two `.responseFile` values are ordered by lexicographically
+    // comparing their flattened chain step arrays.
```
natecook1000, line 194: IIUC, the arguments in a response file are expanded "in place", such that they take the position of the `@file` command-line argument that includes them. Are there any issues with that in-place expansion leading to an actual ordering of arguments that's different from this ordering, where all the arguments from expanded response files are positioned after _all_ command-line arguments?

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+// This source file is part of the Swift Argument Parser open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+internal import Foundation
```
natecook1000, line 15: Let's use this standard check:

```suggestion
#if canImport(FoundationEssentials)
internal import FoundationEssentials
#else
internal import Foundation
#endif
```

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+    var hasResponseFile = false
+
+    for (argvIndex, argument) in arguments.enumerated() {
+      if isResponseFileArgument(argument) {
+        hasResponseFile = true
+        guard let fileName = extractResponseFileName(argument) else {
+          result.append(
+            .init(value: argument, chain: [.argv(index: argvIndex)]))
+          continue
+        }
```
natecook1000: Seems like this is double checking – let's move the `extract...` call up to be the initial condition of the `if`, and eliminate this unreachable guard condition.

```suggestion
      if let fileName = extractResponseFileName(argument) {
        hasResponseFile = true
```

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    return (result, hasResponseFile)
+  }
+
+  /// Determines if an argument is a response file reference.
+  /// - Parameter argument: The argument to check
+  /// - Returns: True if the argument starts with the response file prefix
+  func isResponseFileArgument(_ argument: String) -> Bool {
+    guard argument.count > prefix.utf8.count else { return false }
```
natecook1000: This check is implicitly covered by the next one.

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+    let fileName = String(argument.dropFirst(prefix.utf8.count))
+    return !fileName.isEmpty
+  }
+
+  /// Extracts the filename from a response file argument.
+  /// - Parameter argument: The response file argument (e.g., "@file.txt")
+  /// - Returns: The filename portion, or nil if not a valid response file argument
+  func extractResponseFileName(_ argument: String) -> String? {
+    guard isResponseFileArgument(argument) else { return nil }
+    return String(argument.dropFirst(prefix.utf8.count))
```
natecook1000: This will be incorrect if `prefix` is non-ASCII.
```suggestion
    return String(argument.dropFirst(prefix.count))
```

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+      throw ResponseFileError.fileNotFound(fileURL)
+    }
+
+    // Read file content
+    let content: String
+    do {
+      content = try String(contentsOf: fileURL, encoding: .utf8)
+    } catch {
+      throw ResponseFileError.readError(fileURL, error)
+    }
```
natecook1000:
```suggestion
    let content: String
    do {
      content = try String(contentsOf: fileURL, encoding: .utf8)
    } catch CocoaError.fileReadNoSuchFile {
      throw ResponseFileError.fileNotFound(fileURL)
    } catch {
      throw ResponseFileError.readError(fileURL, error)
    }
```

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+  /// Resolves a file name to an absolute `URL`, handling relative paths.
+  ///
+  /// - Parameters:
+  ///   - fileName: The file name to resolve
+  ///   - parentURL: Optional parent file `URL` for relative resolution
+  /// - Returns: The resolved absolute URL
+  fileprivate func resolveFileURL(
+    _ fileName: String,
+    relativeTo parentURL: URL? = nil
+  ) -> URL {
```
natecook1000, line 313: I think this method is duplicating an existing API: https://developer.apple.com/documentation/foundation/url/init(filepath:directoryhint:relativeto:)

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
```diff
@@ -0,0 +1,578 @@
...
+    return URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+      .appendingPathComponent(fileName)
+  }
+
+  /// Returns `true` if `path` is an absolute filesystem path.
+  ///
+  /// On POSIX, that means it starts with `/`. On Windows, that includes
+  /// drive-rooted paths (`C:\...` / `C:/...`), UNC paths (`\\server\...`),
+  /// and drive-relative-absolute paths (starting with `\` or `/`).
+  fileprivate static func isAbsolutePath(_ path: String) -> Bool {
```
natecook1000: Let's try hard not to rewrite our own version of this logic – is this necessary if `resolveFileURL` is simplified?

Sources/ArgumentParserTestHelpers/TemporaryFileFixture.swift
```diff
@@ -0,0 +1,96 @@
...
+// This source file is part of the Swift Argument Parser open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
natecook1000, line 14:
```suggestion
#if canImport(FoundationEssentials)
internal import FoundationEssentials
#else
internal import Foundation
#endif
```

natecook1000 (commented):
Thanks, @bkhouri! Initial feedback here – I haven't reviewed the tests yet, will do that next.

Sources/ArgumentParserTestHelpers/TemporaryFileFixture.swift
bkhouri: ack

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
bkhouri: It does not.  The `CommandConfiguration` is indeed a more natural home.

Sources/ArgumentParser/Parsing/CommandParser.swift
bkhouri: Will do.

Sources/ArgumentParser/Parsing/CommandParser.swift
bkhouri: Right, not sure how I missed that 🤦🏽

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: ack

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: That's cleaner. thanks.

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: `prefix` is of type `Character`. I don't want to hardcode `1` here in the [unlikely] event, that the response files have a multi-character prefix.  So trying to future proof this by taking the `.count`.

I'll change this to 
```suggestion
    return String(argument.dropFirst(String(prefix).count))
```

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: ack/

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: ack.

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: ack.

Sources/ArgumentParser/Parsing/InputOrigin.swift
bkhouri: I don't believe there are ordering issues.  Every expanded argument (CLI or response-file origin) is assigned a sequential post-expansion `InputIndex` in `SplitArguments.swift`, and all origins use `.argumentIndex(SplitArguments.Index)`. Natural `<` ordering on `.argumentIndex` therefore reflects the true in-place expansion order — a response-file arg at post-expansion index 5 sorts between CLI arg at index 4 and CLI arg at index 6. Chain provenance for user-facing error messages lives separately in `FormattingContext.responseFileChains` keyed by the same post-expansion InputIndex.

There are also tests that help verify the ordering:
 - in `ResponseFileExpanderTests.swift:112-238` that validate the ordering within the response files processing is as expected.
 - in `ResponseFileEndToEndTests.swift`, `responseFileWithMixedArgsLastWinsCLI:122` parses a sample CLI `[@file, --count 100]`, where the response file sets `--count 2`. The test asserts `count == 100`, which comes after the response file on the command line.
 - in `ResponseFileEndToEndTests.swift`, `responseFileWithMixedArgsLastWinsResponseFile:141` is the same as the previous, but reversed the order on the command line.


I added additional automated test that will validate the CLI argument expansion is as expected.

Sources/ArgumentParser/Parsing/ResponseFileExpander.swift
bkhouri: I can't recall why I didn't use the existing API.

Tests/ArgumentParserEndToEndTests/ResponseFileEndToEndTests.swift
bkhouri: added

PR #913, opened 2026-05-31
ArgumentParser: make invalid cast explicit

compnerd:
The error codes on Windows are unsigned (and do make use of the sign bit). We explicitly treat the values as signed, so make the conversion explicit. This is required with future enhancements to enable canonical importing of constants on Windows.


rauhul (approved):

natecook1000:
@compnerd Is this issue something that is catchable in testing?

compnerd:
@natecook1000 sadly not - its an API shape thing: `Int32` vs `UInt32`.

PR #932, opened 2026-07-18
Add ArgumentParserFoundation package trait

mba6comma1:
This PR adds a package trait, `ArgumentParserFoundation` (enabled by default), that conditionalizes ArgumentParser's dependencies on `Foundation.NSError`, `Foundation.LocalizedError`, and `Foundation.JSONEncoder`, such that the library can be used without having to link Foundation. External API availability is unchanged (so it's source-compatible either way); the only major caveat I'm aware of when `ArgumentParserFoundation` is disabled is that `--experimental-dump-help` is rendered all but useless (since it's reliant on `JSONEncoder`).

I'm not sure where the best place to add documentation for this trait would be, or how to go about testing it. (I did manually verify that trivial programs compiled with the trait disabled do not link Foundation in the resulting binary, but I don't know what the best way of automating such a check would be.)

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
+++ b/Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
@@ -23,7 +23,11 @@ internal struct DumpHelpGenerator {
   }
 
   func rendered() -> String {
+    #if ArgumentParserFoundation
     JSONEncoder.encode(self.toolInfo)
+    #else
+    "(cannot dump help to JSON without Foundation)"
+    #endif
   }
 }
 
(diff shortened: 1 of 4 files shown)


rauhul:
can you explain the need to avoid FoundationEssentials, is it binary size?

Sources/ArgumentParser/Usage/DumpHelpGenerator.swift
```diff
@@ -23,7 +23,11 @@ internal struct DumpHelpGenerator {
   }
 
   func rendered() -> String {
+    #if ArgumentParserFoundation
     JSONEncoder.encode(self.toolInfo)
+    #else
+    "(cannot dump help to JSON without Foundation)"
+    #endif
```
natecook1000, line 30: If we're compiling a tool without access to JSONEncoder, I think we shouldn't parse `--experimental-dump-help` as a valid flag…

PR #935, opened 2026-07-28
Document the option terminator

### LINKED ISSUE #39: Document `--` to support all remaining command line args as Arguments
Looks like some of the tests already confirm it, but the documentation doesn't seem to call out that the library does support the standard of taking an `--` argument to then map all following command line args to be Arguments (and never recognize them as Flags/Options).

vivekjm:
## Summary

- add a new guide for people using command-line tools built with ArgumentParser
- use the `--` option terminator as the guide's first topic, with examples for dash-prefixed and option-like positional values
- link the guide from the documentation's Essentials section and keep the declaration article focused on defining command interfaces

Fixes #39

## Testing

- `git diff --check` (passed)
- `swift test --skip CountLinesExampleTests.testCountLines` (passed; the skipped fixture is path-sensitive and fails because this checkout path contains spaces)
- `swift package plugin generate-documentation --target ArgumentParser` was attempted previously, but this toolchain reported that the plugin subcommand was unavailable

### DIFF
--- a/Sources/ArgumentParser/Documentation.docc/ArgumentParser.md
+++ b/Sources/ArgumentParser/Documentation.docc/ArgumentParser.md
@@ -49,6 +49,7 @@ and then either calls your `run()` method or exits with a useful message.
 
 ### Essentials
 
+- <doc:UsingArgumentParserTools>
 - <doc:GettingStarted>
 - ``ParsableCommand``
 - ``AsyncParsableCommand``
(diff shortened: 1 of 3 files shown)


natecook1000:
Thanks for looking at this, @vivekjm! This isn't the correct spot to document the `--` behavior, since that article is about how to declare arguments in your command. We do need an article about _using_ tools built with ArgumentParser. Would you like to take a crack at creating that, with this as the initial section?

vivekjm:
Yes, thanks for the direction. I moved the detailed explanation into a new **Using ArgumentParser Tools** article, linked it from the documentation’s Essentials section, and restored the declaration article to its brief original note. The new guide starts with the `--` behavior and includes both dash-prefixed and option-like positional-value examples.

I also reran the suite with the known path-sensitive CountLines fixture skipped: `swift test --skip CountLinesExampleTests.testCountLines` passes.

PR #937, opened 2026-07-31
Fix Bash completion option filtering

### LINKED ISSUE #936: Generated bash completions: not_found fallback checks non_repeating_flags twice instead of non_repeating_options
### Summary

In the bash completion script generated by `--generate-completion-script bash`,
`__fanctl_offer_flags_options`'s (name varies by command) `not_found` fallback loop is a
copy-paste of the loop above it and iterates `non_repeating_flags` a second time instead
of `non_repeating_options`. As a result, an already-typed **option** (e.g. `--interval`,
`--count` — anything that takes a value, as opposed to a boolean flag) can still be
offered again by completion, since it is never actually removed from
`non_repeating_options`.

### Location

`Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift`, in the generated-
script template inside the function documented as removing already-used flags/options
(around the `# Remove ${word} from ${non_repeating_flags} or ${non_repeating_options} so
it isn't offered again` comment). Confirmed present as of the `1.8.2` tag
(`6a52f3251125d74daf04fcbd5e6f08a75d074382`):

```bash
# Remove ${word} from ${non_repeating_flags} or ${non_repeating_options} so it isn't offered again
local not_found=true
local -i index
for index in "${!non_repeating_flags[@]}"; do
    if [[ "${non_repeating_flags[${index}]}" = "${word}" ]]; then
        unset "non_repeating_flags[${index}]"
        non_repeating_flags=("${non_repeating_flags[@]}")
        not_found=false
        break
    fi
done
if "${not_found}"; then
    for index in "${!non_repeating_flags[@]}"; do          # <-- should be non_repeating_options
        if [[ "${non_repeating_flags[${index}]}" = "${word}" ]]; then    # <-- should be non_repeating_options
            unset "non_repeating_flags[${index}]"                        # <-- should be non_repeating_options
            non_repeating_flags=("${non_repeating_flags[@]}")            # <-- should be non_repeating_options
            break
        fi
    done
fi
```

### Expected behaviour

The fallback branch (entered when `word` was not found among `non_repeating_flags`)
should search and remove from `non_repeating_options` instead, matching the doc comment
directly above it.

### Actual behaviour

The fallback branch searches `non_repeating_flags` again — a no-op the second time,
since anything found there would already have been removed and `not_found` set `false`
in the first loop. `non_repeating_options` is never pruned, so an option already present
on the command line (e.g. `myTool --interval 5 --<TAB>`) can still be offered by
completion.

### How this was found

Noticed reviewing generated completions committed for a downstream project's own CLI
(a swift-argument-parser consumer). Confirmed by diffing the generated output against
this exact source location at the resolved dependency version.

### Suggested fix

```bash
if "${not_found}"; then
    for index in "${!non_repeating_options[@]}"; do
        if [[ "${non_repeating_options[${index}]}" = "${word}" ]]; then
            unset "non_repeating_options[${index}]"
            non_repeating_options=("${non_repeating_options[@]}")
            break
        fi
    done
fi
```

ZHUOLIN0928:
Fixes #936.

Summary:
- Search `non_repeating_options` in the fallback branch of the Bash completion parser.
- Remove the matched option from the option array instead of checking the flags array a second time.
- Update the generated Math Bash completion snapshot.

Validation:
- `swift build --product math`
- Generated Bash completion matches the updated snapshot.
- `bash -n` passes for the generated script.
- `git diff --check` passes.

The full XCTest suite could not run in this environment because the installed Command Line Tools do not provide the XCTest module.

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -107,10 +107,10 @@ extension CommandInfoV0 {
                         fi
                     done
                     if "${not_found}"; then
-                        for index in "${!non_repeating_flags[@]}"; do
-                            if [[ "${non_repeating_flags[${index}]}" = "${word}" ]]; then
-                                unset "non_repeating_flags[${index}]"
-                                non_repeating_flags=("${non_repeating_flags[@]}")
+                        for index in "${!non_repeating_options[@]}"; do
+                            if [[ "${non_repeating_options[${index}]}" = "${word}" ]]; then
+                                unset "non_repeating_options[${index}]"
+                                non_repeating_options=("${non_repeating_options[@]}")
                                 break
                             fi
                         done
--- a/Tests/ArgumentParserExampleTests/Snapshots/testMathBashCompletionScript().bash
+++ b/Tests/ArgumentParserExampleTests/Snapshots/testMathBashCompletionScript().bash
@@ -85,10 +85,10 @@ __math_offer_flags_options() {
                     fi
                 done
                 if "${not_found}"; then
-                    for index in "${!non_repeating_flags[@]}"; do
-                        if [[ "${non_repeating_flags[${index}]}" = "${word}" ]]; then
-                            unset "non_repeating_flags[${index}]"
-                            non_repeating_flags=("${non_repeating_flags[@]}")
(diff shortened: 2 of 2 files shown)


natecook1000:
Merged at the command line in 94ef71b845342ab6f27f4a15349de1a8d2275277

Thanks, @ZHUOLIN0928! 👏🏻

PR #939, opened 2026-08-02
generate-manual: don't repeat inherited arguments in nested sections

### LINKED ISSUE #799: Redundant help and version options in generated man page
When generating a man page with the `generate-manual` plug-in, the top-level --help and --version options get repeated for every subcommand.

This might make sense for the --help display and the completion output, but in the man page, it is very redundant and distracting.

Example, from [my utiluti tool](https://github.com/scriptingosx/utiluti):

```
UTILUTI(1)                  General Commands Manual                 UTILUTI(1)

NAME
     utiluti – Read and set default URL scheme and file type handlers.

SYNOPSIS
     utiluti subcommand [--version] [--help]

DESCRIPTION
     --version
             Show the version.

     -h, --help
             Show help information.

     url     Manipulate default URL scheme handlers

             --version
                     Show the version.

             -h, --help
                     Show help information.

             get     Get the path to the default application.

                     scheme  the url scheme, e.g. 'http' or 'mailto'

                     --bundle-id
                             list bundle identifiers instead of paths

                     --version
                             Show the version.

                     -h, --help
                             Show help information.

             list    List all applications that can handle this URL scheme.

                     scheme  the url scheme, e.g. 'http' or 'mailto'

                     --bundle-id
                             list bundle identifiers instead of paths

                     --version
                             Show the version.

                     -h, --help
                             Show help information.

             set     Set the default app for this URL scheme.

                     scheme  the url scheme, e.g. 'http' or 'mailto'

                     bundleID
                             bundle identifier for the app

                     --version
                             Show the version.
```

**ArgumentParser version:** `1.6.1`
**Swift version:**

```
swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

### Checklist

### Steps to Reproduce

- Create a ParsableCommand with subcommands and a version.
- use generate-manual plugin
- view the generated man page with

```
man -M .build/plugins/GenerateManual/outputs/<commandName>/: <commandName>
```

### Expected behavior
The `--version` and `-h, --help` options should be shown once, at the top level and _not_ be repeated for every subcommand.

At least there should be a way to mark them as global, so they are not repeated.

### Actual behavior
The `--version` and `-h, --help` options are repeated for every sub level.

maxches99:
Fixes #799.

### Motivation

`SinglePageDescription` nests each subcommand's description inside of its super command's list, and `--experimental-dump-help` reports the synthesized `--version` and `--help` flags in the `arguments` of *every* command. The two flags were therefore repeated at every level of the command tree, which buries the arguments that are actually specific to a subcommand.

This is visible in this repository's own snapshots — before this change, `testMathSinglePageManual().mdoc` listed the pair six times:

```
     stats   Calculate descriptive statistics.

             --version
                     Show the version.

             -h, --help
                     Show help information.

             average
                     Print the average of the values.

                     --kind kind
                             The kind of average to provide.

                     values...
                             A group of floating-point values to operate on.

                     --version
                             Show the version.

                     -h, --help
                             Show help information.
```

### Modifications

`SinglePageDescription` now carries the set of arguments documented by enclosing commands and skips an argument when an ancestor's section already describes an identical one, so it is documented once by the outermost command which declares it.

The rule is expressed generally rather than special-casing the two flag names: `ArgumentInfoV0` is `Hashable`, so entries are compared for full equality. Besides the synthesized flags this also collapses an option shared between a command and its subcommands through a common `@OptionGroup`. Sibling subcommands do not suppress each other — only ancestors suppress descendants — so an option that two subcommands happen to declare separately is still documented under each of them.

Multi-page manuals are deliberately unaffected: `MultiPageDescription` does not recurse, each command gets a self-contained page there, and its arguments still need to be listed in full.

The same rendering, after the change:

```
     stats   Calculate descriptive statistics.

             average
                     Print the average of the values.

                     --kind kind
                             The kind of average to provide.

                     values...
                             A group of floating-point values to operate on.
```

### Result

Global options are documented once, at the top of the page, as the issue asks for.

### Testing

- Re-recorded `testMathSinglePageManual().mdoc`; the other eleven manual snapshots are unchanged, confirming multi-page output and commands without subcommands are untouched.
- Full `swift test` passes.
- Rendered the generated `math` page through `man` to confirm the result reads correctly.
- Separately built a small tool with a subcommand that has no arguments of its own and an `@OptionGroup` shared between a command and its subcommands, to confirm an emptied section does not emit a stray `.Bl`/`.El` pair and that the shared option is documented once.

### DIFF
--- a/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathSinglePageManual().mdoc
+++ b/Tests/ArgumentParserGenerateManualTests/Snapshots/testMathSinglePageManual().mdoc
@@ -23,10 +23,6 @@ Print the sum of the values.
 Use hexadecimal notation for the result.
 .It Ar values...
 A group of integers to operate on.
-.It Fl -version
-Show the version.
-.It Fl h , -help
-Show help information.
 .El
 .It Em "multiply, mul"
 Print the product of the values.
@@ -35,18 +31,10 @@ Print the product of the values.
 Use hexadecimal notation for the result.
 .It Ar values...
 A group of integers to operate on.
-.It Fl -version
-Show the version.
-.It Fl h , -help
-Show help information.
 .El
 .It Em stats
 Calculate descriptive statistics.
 .Bl -tag -width 6n
-.It Fl -version
-Show the version.
-.It Fl h , -help
-Show help information.
 .It Em "average, avg"
 Print the average of the values.
 .Bl -tag -width 6n
@@ -55,20 +43,12 @@ The kind of average to provide.
 .Pp
 .It Ar values...
 A group of floating-point values to operate on.
-.It Fl -version
-Show the version.
-.It Fl h , -help
-Show help information.
 .El
 .It Em stdev
 Print the standard deviation of the values.
 .Bl -tag -width 6n
 .It Ar values...
 A group of floating-point values to operate on.
-.It Fl -version
-Show the version.
-.It Fl h , -help
-Show help information.
 .El
 .It Em quantiles
 Print the quantiles of the values (TBD).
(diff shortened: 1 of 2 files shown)


natecook1000 (approved):
I agree with this fix – thanks, @maxches99!

Issue #710, opened 2025-02-10
Switch to swift testing


rauhul:
Self explanatory, swift testing offer more features and uses parallel execution by default which is a nice win

TiagoMaiaL:
Hey @rauhul, if possible I'd like to help with this migration. Do you have any specific order in which files should be ported? Or maybe do you know a simple test suite I could begin with?

natecook1000:
I think I'd like to hold off on this switch for a bit, since swift-testing has a Swift 6 requirement, and we're still supporting back to 5.7.

PR #949, opened 2026-08-22
Migrate ArgumentParserExampleTests to Swift Testing

bkhouri:
Migrate the four `ArgumentParserExampleTests` suites from XCTest to
Swift Testing:

- `RollDiceExampleTests`
- `CountLinesExampleTests`
- `RepeatExampleTests`
- `MathExampleTests`

Add Swift Testing counterparts for the process-executing test helpers
in `ArgumentParserTestHelpers`, which were previously only available as
methods on `extension XCTest`:

- `requireExecuteCommand` (String and [String] overloads) — uses
  `#require` for unrecoverable preconditions (missing executable,
  process failing to launch) and `#expect` for the exit-code and
  stdout/stderr comparisons.
- `expectSnapshot` — reads/writes snapshots next to the caller's file
  using `#function` as the default snapshot name, preserving the
  existing `Snapshots/testMath*CompletionScript().{bash,zsh,fish}`
  files.

Preserved the exact `testMath{Bash,Zsh,Fish}CompletionScript` method
names so `#function`-based snapshot lookup continues to resolve the
existing baselines without renaming files.

Relates to #710

### Checklist

---

### DIFF
--- a/Sources/ArgumentParserTestHelpers/TestHelpers+SwiftTesting.swift
+++ b/Sources/ArgumentParserTestHelpers/TestHelpers+SwiftTesting.swift
@@ -10,7 +10,17 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import Foundation
 import Testing
+
+private final class _BundleMarker {}
+
+private var _debugURL: URL {
+  let bundleURL = Bundle(for: _BundleMarker.self).bundleURL
+  return bundleURL.lastPathComponent.hasSuffix("xctest")
+    ? bundleURL.deletingLastPathComponent()
+    : bundleURL
+}
 
 public func expectResultFailure<T, U: Error>(
   _ expression: @autoclosure () -> Result<T, U>,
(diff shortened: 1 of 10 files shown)


Sources/ArgumentParserTestHelpers/TestHelpers+SwiftTesting.swift
```diff
@@ -66,3 +76,138 @@ public func expectEqualStrings(
...
+  environment: [String: String] = [:],
+  sourceLocation: SourceLocation = #_sourceLocation
+) throws -> String {
+  #if os(Windows)
+  return ""
+  #elseif !canImport(Darwin) || os(macOS)
+  let arguments = Array(command.dropFirst())
+  let commandName = String(command.first!)
+  let commandURL = _debugURL.appendingPathComponent(commandName)
+  // guard (try? commandURL.checkResourceIsReachable()) ?? false else {
```
bkhouri: chore (blocking): remove commented code

Sources/ArgumentParserTestHelpers/TestHelpers+SwiftTesting.swift
```diff
@@ -66,3 +76,138 @@ public func expectEqualStrings(
...
+  if !environment.isEmpty {
+    if let existingEnvironment = process.environment {
+      process.environment =
+        existingEnvironment.merging(environment) { (_, new) in new }
+    } else {
+      process.environment = environment
+    }
+  }
+
+  // guard (try? process.run()) != nil else {
```
bkhouri: chore (blocking): remove commented code

Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift
```diff
@@ -11,26 +11,28 @@
...
-final class CountLinesExampleTests: XCTestCase {
-  override func setUp() {
+@Suite struct CountLinesExampleTests {
+  init() {
     Platform.Environment[.columns] = nil
   }
 
-  func testCountLines() throws {
+  @Test func countLines() throws {
     guard #available(macOS 12, *) else { return }
```
bkhouri, line 26: question: Out of curiosity, why do test test only run on macOS?  Should the test, and related binary, be updated to run on all platforms os is the intents to verify Swift Argument Parser with the `@available(...)` API?

Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift
```diff
@@ -11,26 +11,28 @@
...
+    try requireExecuteCommand(
       command: "count-lines \(testFile.path)", expected: "20\n")
-    try AssertExecuteCommand(
+    try requireExecuteCommand(
       command: "count-lines \(testFile.path) --prefix al", expected: "4\n")
   }
 
-  func testCountLinesHelp() throws {
+  @Test func countLinesHelp() throws {
     guard #available(macOS 12, *) else { return }
```
bkhouri, line 36: question: Out of curiosity, why do test test only run on macOS? Should the test, and related binary, be updated to run on all platforms os is the intents to verify Swift Argument Parser with the `@available(...)` API?

Tests/ArgumentParserExampleTests/MathExampleTests.swift
```diff
@@ -212,40 +213,46 @@ final class MathExampleTests: XCTestCase {
 // swift-format-ignore: AlwaysUseLowerCamelCase
 // https://github.com/apple/swift-argument-parser/issues/710
 extension MathExampleTests {
-  func testMathBashCompletionScript() throws {
-    let script = try AssertExecuteCommand(
+  @Test(
+    .requiresProcessExecution
```
bkhouri, line 219: question: This trait disabled the test on some platforms, for example Windows.  Is this expected?

Tests/ArgumentParserExampleTests/MathExampleTests.swift
```diff
@@ -212,40 +213,46 @@ final class MathExampleTests: XCTestCase {
...
+    let script = try requireExecuteCommand(
       command: "math --generate-completion-script bash")
-    try assertSnapshot(actual: script, extension: "bash")
+    try expectSnapshot(actual: script, extension: "bash")
   }
 
-  func testMathZshCompletionScript() throws {
-    let script = try AssertExecuteCommand(
+  @Test(
+    .requiresProcessExecution
```
bkhouri, line 227: question: This trait disabled the test on some platforms, for example Windows.  Is this expected?

Tests/ArgumentParserExampleTests/CountLinesExampleTests.swift
natecook1000: For all these Qs, I think the gating factor was what I could get working at the time of setting up these executable tests. So the question is more about whether we can get these executable tests to run on Linux/Windows, there isn't a principled reason not to.

natecook1000 (approved):

PR #951, opened 2026-08-22
Migrate more end-to-end suites to Swift Testing

bkhouri:
Migrate six `ArgumentParserEndToEndTests` suites from XCTest to Swift
Testing:

- `TransformEndToEndTests`
- `OptionGroupEndToEndTests`
- `ValidationEndToEndTests`
- `OptionalEndToEndTests`
- `PositionalEndToEndTests`
- `SubcommandEndToEndTests`

`OptionGroupEndToEndTests` and `SubcommandEndToEndTests` previously
relied on `XCTestExpectation` via the
`TestableParsableArguments`/`TestableParsableCommand` protocols. Add
Swift Testing counterparts to `TestHelpers+SwiftTesting.swift`:

- `TestExpectation`: a reference-typed flag that carries a `fulfilled`
  bool. Since ArgumentParser's parse/validate/run pipeline is
  synchronous, we don't need a timeout-based expectation like
  `XCTestExpectation`.
- `TestableSwiftTestingParsableArguments` /
  `TestableSwiftTestingParsableCommand`: mirror the XCTest protocols;
  their default `validate()` / `run()` fulfill `didValidateExpectation`
  / `didRunExpectation`, which tests then assert with
  `#expect(command.didValidateExpectation.fulfilled)`.

Centralizing the type means future changes (e.g. moving to
`confirmation`, adding a counter, adding Sendable) touch only the
helper.

`ValidationEndToEndTests`'s `FooCommand.run()` contained an inline
`XCTAssertEqual`; convert it to `#expect(foo == bar)` so the assertion
routes through Swift Testing when run under `@Test`.
`SubcommandEndToEndTests`'s `Math.run()` gets the same treatment.

`PositionalEndToEndTests` keeps
`disabled_parsing_BadlyFormedPositional` as an untagged method
(matching the previous "not discovered" behavior of the `disabled_`
prefix under XCTest); its `XCTFail` becomes `Issue.record`.

Relates to #710

### Checklist

---

### DIFF
--- a/Tests/ArgumentParserEndToEndTests/SubcommandEndToEndTests.swift
+++ b/Tests/ArgumentParserEndToEndTests/SubcommandEndToEndTests.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Argument Parser open source project
 //
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2020-2026 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -14,8 +14,7 @@ import ArgumentParserTestHelpers
 import Testing
 import XCTest
 
-final class SubcommandEndToEndTests: XCTestCase {
-}
+@Suite struct SubcommandEndToEndTests {}
 
 // MARK: Single value String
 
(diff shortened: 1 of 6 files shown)


Tests/ArgumentParserEndToEndTests/SubcommandEndToEndTests.swift
```diff
@@ -231,7 +230,7 @@ extension BaseCommand.SubCommand {
 
 // swift-format-ignore: AlwaysUseLowerCamelCase
 // https://github.com/apple/swift-argument-parser/issues/710
-extension SubcommandEndToEndTests {
+final class SubcommandEndToEndTestsXCTest: XCTestCase {
```
bkhouri, line 233: I'm sure how to migrate this test to Swift Testing.

@natecook1000 @rauhul: Any ideas here?

natecook1000: I'm not sure what exactly this is trying to test! I think we could just manually update a value in the `validate()` method and check that it worked. What do you think?

bkhouri: The test is using `XCTestExpection` in `SubSubCommand`, which is indirectly a subcommand of `BaseCommand`. See [right line 215](https://github.com/apple/swift-argument-parser/pull/951/changes#diff-206cf154bd5111534aa05811a8548d517a74a5e320d723e0f5ac79089ea6936eR215).

I'm not familiar with  `XCTestExpectation`.

natecook1000 (approved):
@bkhouri All look good! Can you file a separate issue to migrate or remove that one `SubSubCommand` test that we're unsure of?

bkhouri:
> @bkhouri All look good! Can you file a separate issue to migrate or remove that one `SubSubCommand` test that we're unsure of?

I created #962.

PR #952, opened 2026-08-22
Migrate large end-to-end suites to Swift Testing

bkhouri:
Migrate seven `ArgumentParserEndToEndTests` suites from XCTest to Swift
Testing:

- `SourceCompatEndToEndTests`
- `UnparsedValuesEndToEndTests`
- `DefaultSubcommandEndToEndTests`
- `FlagsEndToEndTests`
- `NestedCommandEndToEndTests`
- `RepeatingEndToEndTests`
- `RepeatingEndToEndTests+ParsingStrategy` (extension of the above)

`FlagsEndToEndTests` previously had a small `FlagsEndToEndTestsSWT`
Swift Testing suite alongside the XCTest class; fold its
`parsingCaseIterable_Help` test back into the main suite so all
flag-parsing tests live in one place.

`NestedCommandEndToEndTests` had a private `AssertParseFooCommand`
helper that curried `AssertParseCommand` with `Foo.self` as the root;
port it to `expectParseFooCommand` calling `expectParseCommand` for
the same purpose.

`DefaultSubcommandEndToEndTests` previously used
`AssertParseCommandErrorMessage` (XCTest-only, not ported to Swift
Testing yet). Inline the parse-and-check-message logic in that one
test with `#require` on `Test.current` semantics.

`RepeatingEndToEndTests+ParsingStrategy` is an extension of
`RepeatingEndToEndTests` and must be migrated in the same PR so
`extension RepeatingEndToEndTests` refers to the new `@Suite struct`
rather than the old `XCTestCase` class.

Relates to #710

### Checklist

---

### DIFF
--- a/Tests/ArgumentParserEndToEndTests/DefaultSubcommandEndToEndTests.swift
+++ b/Tests/ArgumentParserEndToEndTests/DefaultSubcommandEndToEndTests.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Argument Parser open source project
 //
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2020-2026 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -12,12 +12,10 @@
 import ArgumentParserTestHelpers
 import ArgumentParserToolInfo
 import Testing
-import XCTest
 
 @testable import ArgumentParser
 
-final class DefaultSubcommandEndToEndTests: XCTestCase {
-}
+@Suite struct DefaultSubcommandEndToEndTests {}
 
 // MARK: -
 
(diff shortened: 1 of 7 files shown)


Tests/ArgumentParserEndToEndTests/DefaultSubcommandEndToEndTests.swift
```diff
@@ -124,183 +124,181 @@ extension DefaultSubcommandEndToEndTests {
...
+    // AssertParseCommandErrorMessage is XCTest-only; approximate with a parse
+    // that must throw and inspect the message.
+    do {
+      _ = try MyCommand.parseAsRoot(["--verbose", "bad-parent"])
+      Issue.record("Parsing should have failed.")
+    } catch {
+      let message = MyCommand.message(for: error)
+      #expect(
+        message == "Command 'Other' is not a parent of the current command.")
+    }
```
natecook1000: Can we use `#expect(throws:...)` for these?

bkhouri: probably not until the minimum version is 6.1, but I'll try a variation of the API.

natecook1000 (approved):

PR #955, opened 2026-08-22
Migrate remaining unit tests to Swift Testing

bkhouri:
Migrate three `ArgumentParserUnitTests` suites from XCTest to Swift
Testing:

- `HelpGenerationTests` — the base file was already using `@Suite`,
  `@Test`, and `#expect`; drop the stale `import XCTest`. The
  `+*.swift` extension files were migrated in earlier PRs.
- `ParsableArgumentsValidationTests` — the largest unit-test file.
  One assertion required an update: comparing `error.type` (an
  existential `any ParsableArguments.Type`) against a concrete type
  fails to type-check under `#expect`'s stricter operand resolution,
  so wrap both sides in `ObjectIdentifier`.
- `CompletionScriptTests` — marked `@Suite(.serialized)` because the
  tests share the `CompletionShell.requesting` global while
  generating completion scripts. Under XCTest the sequential runner
  hid this; Swift Testing's default parallelism caused cross-shell
  contamination in the generated scripts. Private
  `assertCustomCompletion(s)` ported to
  `expectCustomCompletion(s)`.

Relates to #710

### Checklist

---


Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
```diff
@@ -82,47 +81,72 @@ final class ParsableArgumentsValidationTests: XCTestCase {
...
+    //   as? CodingKeyValidator.MissingKeysError
+    // {
+    //   #expect(
+    //     error.missingCodingKeys == [
+    //       InputKey(name: "phrase", parent: parent),
+    //       InputKey(name: "includeCounter", parent: parent),
+    //     ])
+    // } else {
+    //   Issue.record("Expected MissingKeysError")
+    // }
```
natecook1000: For these sections, do we need to preserve the original XCTest version as comments?

natecook1000 (approved):

Tests/ArgumentParserUnitTests/ParsableArgumentsValidationTests.swift
bkhouri: I kept the original code in there in case it wasn't supported on some platforms/Swift versions.  I'll remove them.

PR #959, opened 2026-08-26
Fix parsing of empty long option values

### LINKED ISSUE #958: --opt= silently collapses to a bare flag and consumes the following token (short -o= behaves differently)
## Description

`ParsedArgument.init(_:)` (`Sources/ArgumentParser/Parsing/SplitArguments.swift`, ~lines 21–31) collapses an empty value after `=` into a bare flag:

```swift
let indexOfEqualSign = str.firstIndex(of: "=") ?? str.endIndex
let (baseName, value) = (str[..<indexOfEqualSign], str[indexOfEqualSign...].dropFirst())
let name = Name(baseName)
self = value.isEmpty ? .name(name) : .nameWithValue(name, String(value))
```

Double-dash tokens reach this initializer whole, so `--out=` becomes `.name(.long("out"))`. A value-taking `@Option` then consumes the **next** token via `popNextElementIfValue(after:)`, meaning:

```
prog --out= file.txt
```

parses as `out = "file.txt"` — the user's positional/file argument is silently swallowed as the option value, with no diagnostic. Every downstream positional shifts.

The single-dash path (`SplitArguments.swift` ~743–750) behaves differently and correctly: `-o=` produces `.nameWithValue(.short("o"), "")`, preserving the explicit empty string and leaving `file.txt` positional. Identical user intent (`--opt=` is GNU-conventional for an empty-string value), divergent silent outcomes.

Static trace through `parseIndividualArg` → `Name` dash-stripping → option value-pop against today's main; no Swift toolchain locally, so not executed.

## Steps to Reproduce

```swift
struct O: ParsableArguments {
    @Option(name: .long) var out: String
    @Argument var file: String
}
// input: ["--out=", "file.txt"]
// current: out="file.txt", missing positional → or silently misassigned
```

## Expected Behavior

`--out=` should yield `nameWithValue(.long("out"), "")` exactly like `-o=` does for short flags: the user explicitly typed a value slot, so the empty string is intentional and the following token must remain positional.

## Actual Behavior

Empty value collapses to a bare flag; the next token is consumed as the value.

## Impact

Silent misassignment of user data to the wrong parameter — worst case an input file path becomes an option value. No error surfaces; parsing succeeds with shifted semantics.

## Suggested Direction

Treat presence of `=` as authoritative regardless of value emptiness for long options:

```swift
self = str.contains("=") || !value.isEmpty
    ? .nameWithValue(name, String(value))
    : .name(name)
```

(keeping short-option behavior unchanged), plus regression tests for `["--out=", "file.txt"]` vs `["-o=", "file.txt"]`.

actuallyhrishikesh:
Fixes #958.

This change treats the presence of `=` in a long option as an explicit attached value, including empty values.

For example:

--output=

is now parsed as an option with an empty string value rather than being treated as a bare flag.

Tests:
- Added tokenizer coverage for empty long option values.
- Added an end-to-end regression test for `--output=`.
- Ran the full test suite and formatting checks.

Compatibility note: `--flag=` is now treated as a flag with an explicit empty attached value, consistent with the existing single-dash `-f=` behavior and the issue's explicit `=` rule.

### DIFF
--- a/Sources/ArgumentParser/Parsing/SplitArguments.swift
+++ b/Sources/ArgumentParser/Parsing/SplitArguments.swift
@@ -25,7 +25,7 @@ enum ParsedArgument: Equatable, CustomStringConvertible {
     )
     let name = Name(baseName)
     self =
-      value.isEmpty
+      indexOfEqualSign == str.endIndex
       ? .name(name)
       : .nameWithValue(name, String(value))
   }
--- a/Tests/ArgumentParserEndToEndTests/SimpleEndToEndTests.swift
+++ b/Tests/ArgumentParserEndToEndTests/SimpleEndToEndTests.swift
@@ -49,6 +49,24 @@ extension SimpleEndToEndTests {
   }
 }
 
+// MARK: Empty option value
+
+private struct Qux: ParsableArguments {
+  @Option() var output: String
+  @Argument() var file: String
+}
+
+// swift-format-ignore: AlwaysUseLowerCamelCase
+// https://github.com/apple/swift-argument-parser/issues/710
+extension SimpleEndToEndTests {
+  func testParsing_EmptyOptionValue() throws {
+    AssertParse(Qux.self, ["--output=", "file.txt"]) { qux in
+      XCTAssertEqual(qux.output, "")
+      XCTAssertEqual(qux.file, "file.txt")
+    }
+  }
+}
+
 // MARK: Single value Int
 
 private struct Foo: ParsableArguments {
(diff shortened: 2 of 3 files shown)


rauhul:
how does this compare to #961

actuallyhrishikesh:
Thanks for pointing this out. I hadn’t seen #961 when I opened #959. They appear to address the same underlying issue: preserving the presence of = for long options even when the attached value is empty, so that the following positional argument isn’t consumed.

My #959 focuses on treating --output= as an explicit empty attached value, with tokenizer and end-to-end regression coverage. #961 addresses the same issue by explicitly tracking whether = was present in ParsedArgument.init(_:).

Given the overlap, I’m happy to defer to whichever implementation you think fits the codebase better.

natecook1000:
@actuallyhrishikesh Thanks for this PR – really appreciate the contribution! We're going to go with the solution in #961 instead, as I think that interpretation of the command-line arguments will be more useful for CLI tool users.

PR #961, opened 2026-08-29
Fix --opt= silently consuming the next positional token

### LINKED ISSUE #958: --opt= silently collapses to a bare flag and consumes the following token (short -o= behaves differently)

shubhransh-gupta:
### Motivation

Fixes #958.

`ParsedArgument.init(_:)` treated a long option with an empty value after `=` (e.g. `--out=`) identically to a bare long option (e.g. `--out`). The initializer would collapse `--out=` to `.name(.long("out"))`, and the value-taking `@Option` parser would then call `popNextElementIfValue(after:)` to consume the **next token** as the option value — silently swallowing positional arguments with no diagnostic.

Short options did not have this bug: `-o=` already produced `.nameWithValue(.short("o"), "")`.

### Modifications

**`Sources/ArgumentParser/Parsing/SplitArguments.swift`**

Track whether `=` was present in the raw input independently of whether the value substring is empty:

```swift
// Before
self = value.isEmpty ? .name(name) : .nameWithValue(name, String(value))

// After
let hasEqualSign = indexOfEqualSign != str.endIndex
self = hasEqualSign || !value.isEmpty
    ? .nameWithValue(name, String(value))
    : .name(name)
```

When `=` is present, `--out=` now produces `.nameWithValue(.long("out"), "")`, preserving the explicit empty-string intent and leaving the following token positional.

**Tests**

- `Tests/ArgumentParserUnitTests/SplitArgumentTests.swift` — four unit tests:
  - `longOptionWithEmptyValue`: `--out=` → `.nameWithValue(.long("out"), "")`
  - `shortOptionWithEmptyValue`: `-o=` → `.nameWithValue(.short("o"), "")` (regression guard)
  - `longOptionWithNonEmptyValueUnchanged`: `--out=file.txt` unchanged
  - `longOptionWithoutEqualsSign`: `--out` still produces bare `.name`

- `Tests/ArgumentParserEndToEndTests/EqualsEndToEndTests.swift` — four end-to-end tests confirming that `["--out=", "file.txt"]` parses as `out="", file="file.txt"` rather than `out="file.txt"` with a missing positional.

### Result

All 264 tests pass (260 pre-existing + 4 new).


shubhransh-gupta:
Hi @natecook1000 @bkhouri — Just updated the branch to ensure all swift-format lint style checks and test suites pass cleanly with 0 failures across the 264 unit tests. When you have a moment, could you please take a look? Thank you!

Tests/ArgumentParserEndToEndTests/EqualsEndToEndTests.swift
```diff
@@ -83,3 +83,52 @@ extension EqualsEndToEndTests {
...
+  @Option(name: [.short, .long]) var name: String?
+  @Argument var file: String
+}
+
+// https://github.com/apple/swift-argument-parser/issues/958
+extension EqualsEndToEndTests {
+  /// `--out=` must accept an explicit empty-string value.
+  ///
+  /// The following positional `file.txt` must remain a positional argument.
+  func testLongOptionEmptyValueDoesNotConsumePositional() throws {
```
rauhul: Please update to swift-testing

rauhul:
I agree this is a good fix

shubhransh-gupta:
Hi @rauhul,

Updated the tests in `EqualsEndToEndTests.swift` to Swift Testing (`@Test`, `expectParse`, `#expect`), rebased on the latest `upstream/main`, and verified with `swift-format` and the full test suite (570 tests passing with 0 failures).

Thank you!

natecook1000 (approved):
Thanks for this fix, @shubhransh-gupta! 👏🏻

PR #963, opened 2026-09-01
Provide an interface for tools to implement `SIGINFO` handling.

grynspan:
### Description

This PR adds a new protocol, `InfoProvidingCommand`, to which root command types can conform to indicate they respond to `SIGINFO` or platform equivalents:

- On Darwin and the BSDs, the shell raises `SIGINFO` when you press Ctrl+T;
- On Linux, `SIGUSR1` is used by (weak) convention for this purpose and can be raised with `kill` or `pkill`.
- On Windows, Ctrl+Break serves the same purpose and triggers a Win32-specific callback.

### Detailed Design

The new API is an additional protocol that refines `ParsableCommand`:

```swift
/// A parsable command that can provide information back to the user while it
/// runs.
///
/// The user may request information from a running process via a
/// platform-specific mechanism:
///
/// - On Apple platforms, FreeBSD, and OpenBSD, by sending `SIGINFO` to the
///   process or by pressing Ctrl+T in the Terminal application.
/// - On Linux, by sending the `SIGUSR1` signal to the process.
/// - On Windows, by pressing Ctrl+Break in the Terminal application.
///
/// If your root command conforms to this protocol, Swift Argument Parser
/// automatically sets up a signal handler to listen for `SIGINFO` (or the
/// platform-specific equivalent).
///
/// On platforms that do not support providing information, conformance to this
/// protocol has no effect.
@available(anyAppleOS 26, *)
public protocol InfoProvidingParsableCommand: Sendable, ParsableCommand {
  /// Provide information about the state of the process and about the
  /// currently-running command.
  ///
  /// The information you provide is program-specific. Programs will often
  /// provide a status update such as the percentage or count of work completed,
  /// information about the currently-running work item, etc.
  ///
  /// - Important: Swift Argument Parser calls this function asynchronously
  ///   while your program is running. If your command type is actor-isolated,
  ///   ensure your ``AsyncParsableCommand/run()`` implementation periodically
  ///   yields control by suspending, sleeping, or calling [`Task.yield()`](https://developer.apple.com/documentation/swift/task/yield()).
  ///
  ///   If ``AsyncParsableCommand/run()`` never yields, the Swift runtime may
  ///   not be able to schedule calls to this function and it will appear to the
  ///   user as if it is not implemented.
  nonisolated(nonsending) func provideInfo() async
}
```

Types that conform to this protocol can then implement `provideInfo()` to print whatever information is meaningful at that moment. An example program is included to show how you might use it.

### Documentation Plan

I plan to scrabble something together after the design has been critiqued.

### Test Plan

Added a new unit test based on a Swift Testing exit test. Signals are generally hard to test, and I am open to adding more tests if folks have ideas.

The example program can also be run at desk to validate behavior in an interactive way.

### Source Impact

There should be no impact on existing code.

### Checklist

### DIFF
--- a/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
@@ -26,7 +26,7 @@
 ///
 /// On platforms that do not support providing information, conformance to this
 /// protocol has no effect.
-@available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *)
+@available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
 public protocol InfoProvidingParsableCommand: Sendable, ParsableCommand {
   #if compiler(>=6.2)
   /// Provide information about the state of the process and about the
@@ -52,16 +52,18 @@ public protocol InfoProvidingParsableCommand: Sendable, ParsableCommand {
 
 // MARK: -
 
-@available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *)
+@available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, watchOS 6, *)
 extension SIGINFOHandler {
   convenience init<T>(for command: T) where T: InfoProvidingParsableCommand {
     self.init {
+      guard #available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *) else {
+        _ = Task {
+          await command.provideInfo()
+        }
+        return
+      }
       #if compiler(>=6.3)
       _ = Task.immediate {
-        await command.provideInfo()
-      }
-      #else
-      _ = Task {
         await command.provideInfo()
       }
       #endif
--- a/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
@@ -56,18 +56,18 @@ public protocol InfoProvidingParsableCommand: Sendable, ParsableCommand {
 extension SIGINFOHandler {
   convenience init<T>(for command: T) where T: InfoProvidingParsableCommand {
     self.init {
-      guard #available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *)
-      else {
-        _ = Task.detached {
+      #if compiler(>=6.3)
+      if #available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *) {
+        _ = Task.immediate {
           await command.provideInfo()
         }
         return
       }
-      #if compiler(>=6.3)
-      _ = Task.immediateDetached {
+      #endif
+
+      _ = Task {
         await command.provideInfo()
       }
-      #endif
     }
   }
 }
--- a/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
+++ b/Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
@@ -58,13 +58,13 @@ extension SIGINFOHandler {
     self.init {
       guard #available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *)
       else {
-        _ = Task {
+        _ = Task.detached {
           await command.provideInfo()
         }
         return
       }
       #if compiler(>=6.3)
-      _ = Task.immediate {
+      _ = Task.immediateDetached {
         await command.provideInfo()
       }
       #endif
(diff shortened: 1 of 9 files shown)


Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
```diff
@@ -0,0 +1,70 @@
...
+  ///   while your program is running. If your command type is actor-isolated,
+  ///   ensure your ``AsyncParsableCommand/run()`` implementation periodically
+  ///   yields control by suspending, sleeping, or calling [`Task.yield()`](https://developer.apple.com/documentation/swift/task/yield()).
+  ///
+  ///   If ``AsyncParsableCommand/run()`` never yields, the Swift runtime may
+  ///   not be able to schedule calls to this function and it will appear to the
+  ///   user as if it is not implemented.
+  nonisolated(nonsending) func provideInfo() async
+  #else
+  func provideInfo() async
```
compnerd: This feels a bit repetitive ... `InfoProvidingCommand.provideInfo`. Would `InfoProvidingCommand.information` or `InfoProvidingCommand.provide` might be nicer.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
```diff
@@ -158,6 +158,18 @@ extension ParsableCommand {
...
+      var siginfoHandler: SIGINFOHandler?
+      if #available(macOS 26, iOS 26, watchOS 26, tvOS 26, visionOS 26, *),
+        let command = command as? any InfoProvidingParsableCommand
+      {
+        siginfoHandler = SIGINFOHandler(for: command)
+      }
+      siginfoHandler?.register()
+      defer {
+        siginfoHandler?.unregister()
+      }
```
compnerd, line 171: Given that this pattern occurs at least twice, would it be justified to do:

```swift
withInfoHandler(...) {
  try command.run()
}
```

Sources/ArgumentParser/Utilities/SIGINFOHandler.swift
```diff
@@ -0,0 +1,123 @@
...
+#elseif canImport(Android)
+import Android
+#endif
+
+/// A class whose instances represent `SIGINFO` handlers configured in this
+/// process.
+///
+/// - Note: This class is responsible for handling `SIGINFO`, `SIGUSR1` (Linux),
+///   and Ctrl+Break (Windows). Naming is hard.
+final class SIGINFOHandler: Sendable {
```
compnerd, line 35: I do wish that we had a better name than `SIGINFOHandler` (perhaps `InfoProviderHandler`?).

Sources/ArgumentParser/Utilities/SIGINFOHandler.swift
```diff
@@ -0,0 +1,123 @@
...
+  /// The set of `SIGINFO` handlers configured in this process.
+  ///
+  /// Generally, this array will contain no more than `1` element, but it can be
+  /// larger if multiple commands are made to run in a single process.
+  private static let allSIGINFOHandlers = Mutex<[SIGINFOHandler]>([])
+
+  #if canImport(Dispatch)
```
compnerd, line 42: Do we have places where we do not support dispatch but support s-a-p?

grynspan: Wasm does not have Dispatch.

Sources/ArgumentParser/Utilities/SIGINFOHandler.swift
grynspan: It's not API so we can change it. We're thinking about using it for Ctrl+C too to support cancellation. But for now, it's an implementation detail so I'm not sweating it that much.

Sources/ArgumentParser/Parsable Types/InfoProvidingCommand.swift
grynspan: `provide()` would be too vague outside the context of the protocol. @rauhul suggested just `info()`.

Sources/ArgumentParser/Parsable Types/ParsableCommand.swift
grynspan: Maybe. I'm not too too bothered by it but I can add a helper if other folks want it.

Sources/ArgumentParser/Utilities/SIGINFOHandler.swift
grynspan: `ControlKeyObserver` maybe?

compnerd: > `ControlKeyObserver` maybe?

I really like this if you are okay with the length.

PR #965, opened 2026-09-02
Support option completion for default subcommands

### LINKED ISSUE #488: Add completion for default subcommand to supercommand
As it currently stands, although a command can designate its default subcommand and argument parsing works correctly, there is no completion for arguments, options, or flags of the subcommand unless the subcommand name is specified — which eliminates much of the convenience of having a default subcommand. I'd like it if the completion for the default subcommand was added to the list of all subcommands as the completion for the super command.

Here's an example[^1]:
```swift
@main
struct Greeter: ParsableCommand {
    static let configuration = CommandConfiguration(subcommands: [Informal.self, Formal.self], defaultSubcommand: Informal.self)

    struct Informal: ParsableCommand {
        @Option var firstName: String

        func run() throws {
            print("Hi, \(firstName)")
        }
    }

    struct Formal: ParsableCommand {
        @Option var fullName: String

        func run() throws {
            print("Good day, \(fullName)")
        }
    }
}
```
Here's what happens if you run it:
```
$ greeter informal --f<Tab> # successfully completes to the below line
$ greeter informal --firstName

$ greeter formal --f<Tab> # successfully completes to the below line
$ greeter formal --fullName

# here's the problem
$ greeter <Tab> # lists the following completions
--help    -h        formal    help      informal
$ greeter --f<Tab> # no completions
```
Instead, I would like `greeter <Tab>` to list all subcommands (as it currently does) _and_ all the completions for its default command, so:
```
$ greeter <Tab>
--first-name  --help        -h            formal        help          informal
```

For backwards compatibility and user configurability, it might make sense to make this an optional parameter to the `@Argument`/`@Option`/`@Flag` initializer (or perhaps `CompletionKind`?)

[^1]: I made up this code sample because the project I was working on when I found this error is more complex, but this should get the point across

pentaoa:
Adds a default subcommand's named arguments to its parent command's generated completions, so options and flags remain completable when the default subcommand name is omitted.

The completion metadata retains the owning subcommand for custom completion callbacks and removes duplicate option names inherited through nested defaults. Positional arguments are intentionally not merged because they would compete with explicit subcommand names.

Fixes #488.

### Checklist
- [x] I've added tests that validate Bash, Zsh, and Fish output
- [x] I've read the Contribution Guidelines
- [x] I've updated the existing math completion, dump-help and manual snapshots

### Testing
- Generated Bash, Zsh, and Fish scripts match the updated math completion snapshots.
- `Tests/ArgumentParserExampleTests/CompletionTests` contains command-line/Tab/output fixtures using the existing `Math → Add` hierarchy. All 9 real-shell cases pass with bash, zsh and fish; the original scripts fail the 6 omitted-default cases. The runner uses Python's standard library and a PTY; its README includes the reproduction command.
- `math` and `generate-manual` build successfully. The three dump-help/manual snapshot changes correspond to the inferred `defaultSubcommand: add` metadata and the now-optional subcommand in the synopsis.
- `CompletionExampleTests` registers all shell/fixture pairs with Swift Testing and reports driver failures as test failures. All nine cases passed through the actual Swift Testing entry point in an isolated build with the package-generated resource accessor. The full package suite remains blocked locally by missing XCTest in this Command Line Tools-only installation; the repository CI matrix must validate it.
- Fixed the Python driver license header, import order and line lengths reported by CI. The POSIX cleanup API is covered by the repository's existing unacceptable-language ignore mechanism.

AI assistance: Codex was used to investigate the issue, implement the change, and draft tests and this PR description. I reviewed the diff and verified the behavior described above.

### DIFF
--- a/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
+++ b/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift
@@ -209,22 +209,23 @@ extension CommandInfoV0 {
       declareTopLevelArray = ""
     }
 
-    let positionalArguments = positionalArguments
+    let arguments = completionArguments
+    let positionalArguments = arguments.filter {
+      $0.argument.kind == .positional
+    }
 
-    let arguments = arguments ?? []
-
-    let flags = arguments.filter { $0.kind == .flag }
-    let options = arguments.filter { $0.kind == .option }
+    let flags = arguments.filter { $0.argument.kind == .flag }
+    let options = arguments.filter { $0.argument.kind == .option }
     if !flags.flatMap(\.completionWords).isEmpty
       || !options.flatMap(\.completionWords).isEmpty
     {
       result += """
-            \(declareTopLevelArray)repeating_flags=(\(flags.filter(\.isRepeating).flatMap(\.completionWords).joined(separator: " ")))
-            \(declareTopLevelArray)non_repeating_flags=(\(flags.filter { !$0.isRepeating }.flatMap(\.completionWords).joined(separator: " ")))
-            \(declareTopLevelArray)repeating_options=(\(options.filter(\.isRepeating).flatMap(\.completionWords).joined(separator: " ")))
-            \(declareTopLevelArray)non_repeating_options=(\(options.filter { !$0.isRepeating }.flatMap(\.completionWords).joined(separator: " ")))
+            \(declareTopLevelArray)repeating_flags=(\(flags.filter(\.argument.isRepeating).flatMap(\.completionWords).joined(separator: " ")))
+            \(declareTopLevelArray)non_repeating_flags=(\(flags.filter { !$0.argument.isRepeating }.flatMap(\.completionWords).joined(separator: " ")))
+            \(declareTopLevelArray)repeating_options=(\(options.filter(\.argument.isRepeating).flatMap(\.completionWords).joined(separator: " ")))
+            \(declareTopLevelArray)non_repeating_options=(\(options.filter { !$0.argument.isRepeating }.flatMap(\.completionWords).joined(separator: " ")))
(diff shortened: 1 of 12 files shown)


rauhul:
Tbh for changes like this I would really prefer we have a fixture for each of bash/zsh/fish with:
- A ParsableCommand (or hierarchy)
- A example command line with a tab location e.g. `math ad<DEBUG>d 1`
- The expected completion output

That would allow us to verify the end user experience is what we expect.

pentaoa:
Added interactive fixtures using the existing `Math → Add` hierarchy in `Tests/CompletionTests`. Each case specifies the command line, Tab location and resulting line; the runner generates the scripts and exercises real bash/zsh/fish line editors in a PTY. It covers the omitted default command, a cursor before a suffix, and nested value completion.

All 9 cases pass locally; the original scripts fail the 6 omitted-default cases. The README includes the command to reproduce this without Python dependencies. I also updated the three dump-help/manual snapshots missed when switching to the inferred default command name, which caused the CI failures.

The full Swift Testing suite remains unavailable on my Command Line Tools-only host, so I haven't claimed a local full-suite pass. The example and manual generator build successfully.

rauhul:
this is very cool, though for the harness to be a good citizen in this repo it would be preferable for the tests to run via swift-testing

pentaoa:
Added `CompletionExampleTests` to the existing example test target. `swift test --filter CompletionExampleTests` now runs each installed shell/fixture pair separately and reports the PTY driver output on failure; the driver and fixtures are package resources.

All nine cases passed through Swift Testing in an isolated build using the package-generated resource accessor. The full package suite is still blocked locally by missing XCTest in my Command Line Tools installation. I also fixed the license header and Python lint failures from CI.

PR #967, opened 2026-09-11
Don't install a built-in help name that the command already declares

hxperl:
Declaring an argument whose name matches a built-in help name is accepted, and the two definitions then coexist: the command ends up with two live arguments using the same name, and which one wins depends on which code path you hit. This is #778.

### Reproduction

Add a `-h, --horse` option to the `repeat` example (this is the reporter's case, on `main` at b53c2f0):

```swift
@Option(name: .shortAndLong, help: "The name of the talking horse.")
var horse: String
```

```
$ .build/debug/repeat --help
USAGE: repeat [--count <count>] --horse <horse> [--include-counter] <phrase>

OPTIONS:
  -c, --count <count>     How many times to repeat 'phrase'.
  -h, --horse <horse>     The name of the talking horse.
  -h, --help              Show help information.      <-- -h listed twice

$ .build/debug/repeat -h
Error: Missing value for '-h <horse>'                 <-- here -h is --horse
  ...
$ .build/debug/repeat -h Secretariat
USAGE: repeat ...                                     <-- here -h is --help
$ echo $?
0                                                     <-- and it exits 0
```

That last one is the part that bites: the command line is simply incomplete (`<phrase>` is missing), but instead of the usual validation error and exit code 64, the parser decides a help flag was present, prints the help screen and exits successfully.

The same shape shows up in generated completion scripts. `DefaultAsFlagCompletionTests`' fixture command declares `@Flag var help: Bool`, and the recorded bash snapshot on `main` contains:

```
local -a non_repeating_flags=(--help -h --help)
```

### Cause

`getHelpNames(visibility:)` in `HelpGenerator.swift` computes the built-in help names purely from `CommandConfiguration.helpNames` (or the `[.short, .long]` default) and never looks at what the command itself declares. Everything downstream reads from it — `CommandParser.checkForBuiltInFlags` for detection, `argumentsForHelp` for the help screen, `allArguments` for `--experimental-dump-help` and the completion generators — so all of them get the colliding name.

### Fix

`getHelpNames(visibility:)` now drops any help name that is already declared by the last command in the stack. Filtering at that one function is what keeps parsing and rendering in agreement, and it generalises: it covers a configured `helpNames` (including `customShort`/`customLong`) and the `--help-hidden` visibility variant, not just the default `-h`/`--help` pair. If every help name collides, `helpArgumentDefinition()` already returns `nil` for an empty name list, which is the same state a command reaches today by configuring `helpNames: []`.

This is the "I would expect `--help`/`-h` to only be added if no existing flag or option already has that name" half of what @rgoldberg described and @natecook1000 agreed could be treated as a bug fix. I deliberately left the third item — the auto-generated `help` **subcommand** shadowing a user-declared subcommand named `help` — out of this PR, because it lives in subcommand registration rather than in help-name resolution and wants its own change. Happy to follow up with it.

**Judgement call worth disagreeing with:** I made the declared name win and the built-in help name yield. The alternative is to diagnose the collision loudly — the reporter's own expectation was "stop immediately with a fatal error" — which is arguably more helpful but is a source-breaking change for anyone who currently ships a command with a `-h` option. Silently yielding matches the existing `helpNames: []` escape hatch and keeps `--help` working in the common `-h`-only collision. If you would rather have a precondition failure or a build-time diagnostic, say so and I will redo it.

Also note the added `ArgumentSet` construction on the help-name path: `checkForBuiltInFlags` is called several times per parse, so this reflects the command's properties a handful of extra times per run. It was not measurable in the test suite, but if you would prefer, the declared names can be threaded in from the call sites instead of recomputed.

### Snapshot changes

Three `DefaultAsFlagCompletionTests` snapshots change, all of them the duplicate disappearing:

- bash: `non_repeating_flags=(--help -h --help)` → `(--help -h)`
- zsh: `'(-h --help)'{-h,--help}'[Show help information.]'` → `'-h[Show help information.]'`
- fish: `-s 'h' -l 'help'` → `-s 'h'`

### Tests

`Tests/ArgumentParserEndToEndTests/HelpNameCollisionEndToEndTests.swift`, six tests covering: `-h <value>` reaching the declared option; an incomplete command line producing the validation error rather than a successful help screen; `--help` reaching a declared `help` flag while `-h` still opens help; a custom `helpNames` collision; and the two rendered help screens. Four of the six fail on `main` without the source change.

### Verification

macOS 14 / arm64, Swift 6.2, `swift test`: 568 tests, 12 failures — all twelve in `GenerateManualTests`, all of them a date-formatting difference in my environment (`.Dd 5 12, 1996` vs `.Dd May 12, 1996`) that reproduces identically on unmodified `main`. Aside from those, the full suite passes. `swift format lint --strict` is clean on both changed Swift files. I have not tested on Linux or Windows; CI will need to confirm those.

I have not updated `CHANGELOG.md` yet — happy to add a Fixes entry referencing this PR number if you want it in this change.

### Checklist


rauhul:
Seems right, ill let @natecook1000 review this though

